PR #1, opened 2024-02-08
Add proposal for ADR-03

blinry:
Does this seem reasonable to you?

### DIFF
--- a/docs/decisions/03-process-structure.md
+++ b/docs/decisions/03-process-structure.md
@@ -1,0 +1,101 @@
+---
+status: accepted
+date: 2024-02-08
+---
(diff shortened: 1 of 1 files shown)


docs/decisions/03-process-structure.md
```diff
@@ -0,0 +1,101 @@
+---
+status: proposal
```
zormit:
```suggestion
status: accepted
```

zormit: happy to already merge it `accepted` :)

zormit (approved):
yes, seems reasonable, let's take it in :)

PR #5, opened 2024-03-14
Add ADR-04 about who writes files, and when

blinry:


docs/decisions/04-file-saving.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Decision Outcome
+
+Chosen option: "When editors open a file, they take ownership of it. When no editor has the file open, the daemon writes to it using its own logic"
+
+Notes:
+
+- We could provide another "interface" to the content, through ethersync tooling, like `ethersync run "sed -i s/foo/bar file"`.
+- Maybe we should implement a ping mechanism, so that the daemon can detect when editors have crashed.
+- When a file is opened in more than one editor at once, there will be conflicting writes anyway. This is true without a software like Ethersync. The only way to prevent that would be to stop editors from writing at all, which doesn't seem desirable.
```
zormit:
```suggestion
- When a file is opened in more than one editor at once, there could be conflicting writes anyway. This is true without a software like Ethersync. The only way to prevent that would be to stop editors from writing at all, which doesn't seem desirable.
```

docs/decisions/04-file-saving.md
```diff
@@ -0,0 +1,108 @@
...
+- To incorporate changes done by external tools, we watch the file system, compute diffs, and apply changes to the CRDT. This seems true both when the file is open or closed in editors.
```
zormit:
```suggestion
- To incorporate changes done by external tools, we watch the file system, compute diffs, and apply changes to the CRDT. This is independent of whether the file is open or closed in editors.
```

zormit (approved):

PR #9, opened 2024-04-12
Add pre-commit tooling configuration

zormit:
For my own sake (i have now multiple times missed to run the formatter before pushing my stuff), I have set up a pre-commit hook. I find this tool relatively convenient, but it's maybe too opinionated... (and adds a Python dependency to the dev tooling)

https://pre-commit.com/index.html

What do you think about it? Should we add it "for everyone"? If not, I can probably find a way to put my config elsewhere or git-ignore it...


blinry (approved):
I like it.

PR #10, opened 2024-05-03
Only open/close/send edits for the file that was initially opened

zormit:
and only when it's shown :)

### DIFF
--- a/vim-plugin/plugin/ethersync.lua
+++ b/vim-plugin/plugin/ethersync.lua
@@ -9,6 +9,8 @@ local client
 
 -- Toggle to simulate the editor going offline.
 local online = false
+-- currently we're only supporting editing *one* file. This string identifies, which one that is.
+local theFile
 
 -- Queues filled during simulated "offline" mode, and consumed when we go online again.
 local opQueueForDaemon = {}
@@ -105,7 +107,10 @@ local function connect(socket_path)
         end,
     })
     online = true
+end
 
+-- Send "open" message to daemon for this buffer.
+local function openCurrentBuffer()
     local filename = "file://" .. vim.fs.basename(vim.api.nvim_buf_get_name(0))
     client.notify("open", { uri = filename })
 end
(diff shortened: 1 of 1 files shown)


zormit:
Open question:
- [ ] how to detach from events? is it even necessary?
- [ ] I had a bug where I got a couple of "close" messages when i switch *to* the buffer, but I couldn't reproduce that.

blinry:
Neat, thanks for working on this! :)

I changed it to use buffer events that trigger when loading/unloading the buffer completely, to allow background edits, that seems to work well.

And I added a check to route incoming deltas to the correct buffer, even when it's not the current one.

Another open thing for this PR:

* [ ] The Vim behaviour tests and the fuzzer seem to be broken, I haven't checked why!

blinry:
I fixed the tests by sending the correct file URIs back to Vim!

And Vim now only connects when opening the first "Ethersync-enabled" file, this seems more stable to me than always doing it in the beginning. For example, when the user goes into an Ethersync-enabled directory, but opens a file outside of it.

Pretty happy with this PR, still needs a polishing pass – think some comments in the Vim plugin are incorrect now that I changed the buffer events.

blinry (approved):
Fixed the remaining comments, I think it's ready to be merged!

daemon/src/daemon.rs
```diff
@@ -726,7 +729,7 @@ async fn listen_socket(
                         Ok(rev_delta) = editor_message_rx.recv() => {
                             debug!("Received editor message to send to it.");
                             let message = EditorProtocolMessage::Edit {
-                                uri: "file://hamwanich".to_string(),
```
zormit, line 729: Bye bye `hamwanich`, it was nice to have you. Welcome `{}`!

vim-plugin/plugin/ethersync.lua
```diff
@@ -244,39 +259,23 @@ function Ethersync()
...
-    autocmd!
-    autocmd BufEnter * lua Ethersync()
-    autocmd BufUnload * lua EthersyncClose()
-augroup END
-]],
-    false
-)
-
-vim.api.nvim_create_user_command("Ethersync", Ethersync, {})
+vim.api.nvim_create_autocmd({ "BufRead", "BufNewFile" }, { callback = EthersyncOpenBuffer })
```
zormit, line 277: Sweet! Feels like switching from Assembly code to C proper because we can afford it :D

zormit: `BufNewFile` is an interesting one. That already goes in the direction of creating files?

zormit (commented):
Left some comments just for fun (nothing blocking or anything of bigger importance). I'll test it now live and if it works I'll merge.

PR #11, opened 2024-05-06
Add Nix setup to CI

blinry:
This runs the fuzzer and the test suite in a Nix environment which has an Ethersync-enabled Neovim installed.


blinry:
Seems to work, happy with this PR!

zormit (approved):

PR #12, opened 2024-05-08
Clean up Vim plugin code

blinry:
This mainly pulls out the delta input/output logic into a helper file, removes unused/untested code, and adds the possibility to configure the socket via the `ETHERSYNC_SOCKET` environment variable.

This breaks our debug() in Vim, because the changetracker file doesn't have access to the `client`. It always was a bit weird anyway – maybe we should write it to a file, like [Vim's LSP feature does](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/log.lua)?

### DIFF
--- a/daemon/src/actors.rs
+++ b/daemon/src/actors.rs
@@ -261,7 +261,7 @@ pub mod tests {
             let mut cmd = tokio::process::Command::new("nvim");
             cmd.arg("--headless").arg("--embed");
             let (nvim, _, _) = new_child_cmd(&mut cmd, handler).await.unwrap();
-            nvim.command("Ethersync")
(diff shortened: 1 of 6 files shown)


zormit (approved):
I have added a "function docstring" for trackChanges, please check it. This cleanup looks very nice, thanks for pulling it out, I like the idea with the callback.

One thing I've noted is that in the "unhappy path" it's not behaving a bit random: When I try to start the plugin without a running daemon, then I get a lot of on/off message noise that I have to acknowledge. Before it just said "Ethersync activated...", which was also untrue. But maybe fixing that is a new feature, so I'll approve.

PR #13, opened 2024-05-16
Add logging to nvim plugin

zormit:
This is WIP.

Currently I see two ways to do logging:
- roll it ourselves quick & dirty and just write stuff to a file with io
- use another plugin / package to do it

I have found https://github.com/smartpde/debuglog, which doesn't seem too bad. Maybe there are better ones, I haven't done the full research.

This branch currently shows both options in a proof of concept way.

To enable the logging via plugin, you'll need to install debuglog, e.g. like this for Lazy:

```lua
return {
  {
    "smartpde/debuglog",
    config = function()
      require("debuglog").setup({})
    end,
  },
}
```

### DIFF
--- a/vim-plugin/lua/dlog.lua
+++ b/vim-plugin/lua/dlog.lua
@@ -22,6 +22,11 @@ local function noop(_) end
 ---@return fun(msg: string, ...): any logger function
 function M.logger(logger_name)
     if has_debuglog then
+        debuglog.enable("*")
+        debuglog.set_config({
+            log_to_file = true,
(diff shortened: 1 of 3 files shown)


blinry:
Notes:

- blinry prefers the simpler version
- we could add timestamps to the log
- and maybe a toggle like `:EthersyncDebugEnable` / `Disable`

zormit:
We continued this work directly on `main` (as one does in this project :-P), so closing this one. see 8d87734.

PR #16, opened 2024-05-17
VS Code plugin

blinry:
This is the draft of a VS Code plugin, generated following [this guide](https://code.visualstudio.com/api/get-started/your-first-extension) To try it, open `src/extension.ts` in VS Code, and press F5. Then, run the "Hello world" command. This will activate the plugin.

There still are a couple of problems:

- [ ] When receiving edits too quickly, the `ignoreEdits` mechanism will not work quite right, I think the events are ran "too async"? We need to find a better way to fix that.
- [ ] The plugin doesn't check for the `.ethersync` directory currently.
- [ ] When opening more than one buffer, it will still try to send open and edit commands.

### DIFF
--- a/vscode-plugin/.eslintrc.json
+++ b/vscode-plugin/.eslintrc.json
@@ -1,0 +1,24 @@
+{
+    "root": true,
+    "parser": "@typescript-eslint/parser",
+    "parserOptions": {
+        "ecmaVersion": 6,
+        "sourceType": "module"
+    },
+    "plugins": ["@typescript-eslint"],
+    "rules": {
+        "@typescript-eslint/naming-convention": [
+            "warn",
+            {
+                "selector": "import",
+                "format": ["camelCase", "PascalCase"]
+            }
+        ],
+        "@typescript-eslint/semi": "warn",
+        "curly": "warn",
+        "eqeqeq": "warn",
+        "no-throw-literal": "warn",
+        "semi": "off"
+    },
+    "ignorePatterns": ["out", "dist", "**/*.d.ts"]
+}
--- a/vscode-plugin/.gitignore
+++ b/vscode-plugin/.gitignore
@@ -1,0 +1,5 @@
+out
+dist
+node_modules
+.vscode-test/
+*.vsix
--- a/vscode-plugin/.prettierrc
+++ b/vscode-plugin/.prettierrc
@@ -1,0 +1,4 @@
+tabWidth: 4
+semi: false
+bracketSpacing: false
+printWidth: 120
--- a/vscode-plugin/.vscode-test.mjs
+++ b/vscode-plugin/.vscode-test.mjs
@@ -1,0 +1,5 @@
+import {defineConfig} from "@vscode/test-cli"
+
+export default defineConfig({
+    files: "out/test/**/*.test.js",
+})
--- a/vscode-plugin/.vscode/extensions.json
+++ b/vscode-plugin/.vscode/extensions.json
@@ -1,0 +1,5 @@
+{
+    // See http://go.microsoft.com/fwlink/?LinkId=827846
+    // for the documentation about the extensions.json format
+    "recommendations": ["dbaeumer.vscode-eslint", "ms-vscode.extension-test-runner"]
+}
--- a/vscode-plugin/.vscode/launch.json
+++ b/vscode-plugin/.vscode/launch.json
@@ -1,0 +1,17 @@
+// A launch configuration that compiles the extension and then opens it inside a new window
+// Use IntelliSense to learn about possible attributes.
+// Hover to view descriptions of existing attributes.
+// For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Run Extension",
+            "type": "extensionHost",
+            "request": "launch",
+            "args": ["--extensionDevelopmentPath=${workspaceFolder}"],
+            "outFiles": ["${workspaceFolder}/out/**/*.js"],
+            "preLaunchTask": "${defaultBuildTask}"
+        }
+    ]
+}
--- a/vscode-plugin/.vscode/settings.json
+++ b/vscode-plugin/.vscode/settings.json
@@ -1,0 +1,11 @@
+// Place your settings in this file to overwrite default and user settings.
+{
+    "files.exclude": {
+        "out": false // set this to true to hide the "out" folder with the compiled JS files
+    },
+    "search.exclude": {
+        "out": true // set this to false to include "out" folder in search results
+    },
+    // Turn off tsc task auto detection since we have the necessary tasks as npm scripts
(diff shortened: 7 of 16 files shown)


blinry:
Continuing the discussion about what needs to be done here!

It seems that VS Code's [Position](https://code.visualstudio.com/api/references/vscode-api#Position) type, contrary to what I believed, counts the `character` property in UTF-16. (I haven't confirmed that.) But our editor protocol currently expects UTF-32/"Unicode codepoint" counts.

## Options

(Some already mentioned by you in other channels.)

1. Write code in the VS Code plugin to convert and send UTF-32.
	- Good, because we wouldn't need to change anything else right now. Could be the fastest route to short-term success?
	- Bad, because it seems like work *all* Javascript-based text editors would need to do. Goes against our "let's make it simple for plugins" principle.
2. Change the editor protocol to expect UTF-16 instead.
	- Good, because LSP [mandates](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#positionEncodingKind) utf-16 support in servers. So all editors with LSP support will also support it, I guess!
	- Neutral, because we'd need to change some of our unit tests in the daemon, the way that the daemon transforms between EditorDeltas and Deltas, and the calculations in our "most annoying" piece of code [in the Neovim plugin](https://github.com/ethersync/ethersync/blob/d064a44ef04b3c575a46f2d412a7ec9558c770d0/vim-plugin/lua/changetracker.lua#L59). But it seems we need to do work *somewhere* anyway, I'd prefer to do it once, and mostly in the daemon. In the Neovim plugin, we can swap in "utf-16" in some places, and in our annoying code, could we make use of functions in [sync.lua](https://github.com/neovim/neovim/blob/3bcd5624beec5d3fd90b9d7977fdc36a61ee1643/runtime/lua/vim/lsp/sync.lua)? In Rust, there's stuff like [encode_utf16](https://doc.rust-lang.org/std/primitive.str.html#method.encode_utf16).
	- Bad, because I hoped we could go with the "most neutral" option, UTF-32.
3. Allow sending a "encoding" property in each "delta" message. The daemon could then accept utf-32 and utf-16, and do the correct thing.
	- Bad, because we can't really use this in the other direction: We kinda need to know which encoding the editor would like.
4. Do a negotiation step in the editor protocol, like LSP does, to find out the supported encodings of the client. Then use that.
	- Good, because the LSP people apparently arrived that this solution. So it might be the "best one", in some sense.
	- Good, because it's not much work for editors.
	- Neutral, it complicates our editor protocol a bit.
	- Bad, because the daemon then has to support utf-32 and utf-16.

Research questions:

- Are there other people producing UTF-32 updates from VS Code? Didn't you find a GitHub issue regarding that, @zormit?

blinry:
If I had to pick one option now, I'd go for 2.

zormit:
> It seems that VS Code's [Position](https://code.visualstudio.com/api/references/vscode-api#Position) type, contrary to what I believed, counts the character property in UTF-16. (I haven't confirmed that.) But our editor protocol currently expects UTF-32/"Unicode codepoint" counts.

If you want to verify my claim, check out the https://github.com/ethersync/ethersync/tree/understand-vscode-positions branch.

---

Commenting on the options (thanks for listing them!):
1. "Write code in the VS Code plugin to convert and send UTF-32."
- Bad, because we would need to keep track of old content ourselves inside the editor. The [event](https://code.visualstudio.com/api/references/vscode-api#TextDocumentChangeEvent) only provides us with the new state of the document. We do track the old content already inside the daemon to do OT and it's a lot of duplicate data.
2. "Change the editor protocol to expect UTF-16 instead."
- I personally see your "Neutral" argument as Bad. I don't want to rip up existing working code. It can be a good baseline to develop other plugins against. It probably still has some bugs but it's been well tested while dogfeeding. So I'm more for "not touching the running system".
- If I "steelman" your position, then I would say it's a good opportunity for refactoring "the ugly code"
3. We seem to agree that this is not a proper solution :D
4. This would be my personal preference.

blinry:
I confirmed that the `[understand-vscode-positions](https://github.com/ethersync/ethersync/tree/understand-vscode-positions)` experiment inserts at the wrong places if there are carrots! Too bad. And weird that that's not well documented.

blinry:
There's discussion on "the best" unit for `character` in [this LSP issue](https://github.com/microsoft/language-server-protocol/issues/872). An Emacs LSP library seems to use UTF-32, as well? (I didn't fact-check this.)

I tried to find out what unit IntelliJ uses, but, fustratingly, it also [just says "character"](https://plugins.jetbrains.com/docs/intellij/coordinates-system.html).

blinry:
I tried out the current branch of this PR, and it seems to work pretty well – is my understanding correct that the plugin and the conversion from/to UTF-32 basically works, and that the missing piece is getting the content before the edit, to allow for correct conversion in all situations?

I wonder how much work it would be to track the previous content. (We need to do that in Vim, as well, in the changetracker module.) Probably an additional dictionary like the `revisions` dictionary, right? And putting the file content in there after each change? Maybe option 1 is the one that requires the least work from us, right now?

zormit:
Huh, I forgot that we do track the previous content in nvim. We indeed do.

https://github.com/ethersync/ethersync/blob/39afe5ada8ad62902c2c4e9203184ca116e61157/vim-plugin/lua/changetracker.lua#L14-L16

Somehow my programmer mind takes this as a very bad thing but for nvim it seems to work well so I agree that we should try option 1 and see how it performs. I'll tackle that today.

zormit:
Cool. This was really the easiest way by far! 🎉 It mostly seems to work. It's still sending some invalid stuff from time to time which makes the daemon rope-crash. Also it's a very rough code and relies on happy path quite heavily. It's probably still a lot of work to get this "smooth"

Here's an incomplete collection of todos:
- [ ] figure out why it's crashing from time to time. It has to do with changes where the peers type in parallel as far as I can tell from the logs. So something with the atomicity still doesn't work, maybe?
- [ ] unhappy path, daemon disconnect: if the daemon crashes, `connection.sendNotification` throws an "Error: Connection is closed.". We probably should stop syncing at this point.
- [ ] proper "open": Dealing with tabs from previously existing sessions. Also sometimes we get two opens, does that hurt?
- [ ] "open": check whether we're in the correct project directory. (by [testing the workspaceFolders](https://github.com/ethersync/ethersync/blob/99e121d62274382fb1c2936c65189eb7dc90af5c/vscode-plugin/src/extension.ts#L18), i guess)
- [ ] check if newline stuff is handled correctly, especially at end of file and empty file edge cases. Currently we store the previous content without line endings. Could that be a source of problems?
- [ ] (re)loading a file from disk will cause (unwanted) edits. (maybe [this open collab code](https://github.com/TypeFox/open-collaboration-tools/blob/a01592853cfa8ef871ac187497822cbcc5955860/packages/open-collaboration-vscode/src/collaboration-instance.ts#L411) is related/helpful?)
- [ ] unhappy path, JSON-RPC: Currently we're sending Notifications. Upgrade this to Requests to be more resilient/helpful in case it's getting rejected?

PR #17, opened 2024-05-17
Refactoring: Separating out responsibilities in/from daemon.rs

zormit:
This is a huge refactoring (which _hopefully_ makes everything down the line a bit easier to understand and cleaner).

Mostly this is about separating things and applying @Darksonn's [actor-handle pattern](https://ryhl.io/blog/actors-with-tokio/) over and over again. Also a huge help and inspiration (and source of a lot of neat tricks) was the sample implementation [here](https://github.com/Darksonn/telnet-chat).


daemon/src/lib.rs
```diff
@@ -1,5 +1,8 @@
 pub mod actors;
```
zormit, line 1: not entirely sure if we really need all the modules listed here, or if a more tree-like import works as well.

daemon/src/main.rs
```diff
@@ -5,7 +5,7 @@ use std::io;
 use std::path::PathBuf;
 use tokio::signal;
 
-mod client;
+mod lsp_client;
```
zormit, line 8: i'm not sure about the concrete name, but I felt like "client" is not specific enough :)

we have three kinds of clients in our codebase:
- the standalone "ethersync client" binary (now called `lsp_client`)
- the `editor` (representing an editor plugin connecting through LSP client and socket ), which is client to the daemon
- the `peer` (representing a fellow daemon to sync with) which is a client to the daemon

daemon/src/peer.rs
```diff
@@ -0,0 +1,288 @@
...
+    }
+}
+
+#[derive(Clone)]
+pub struct TCPActorHandle {
+    automerge_message_tx: mpsc::Sender<AutomergeSyncMessage>,
+    shutdown_token: CancellationToken,
+}
+
+/// The TCP statemachine works as follows:
```
zormit, line 251: - [ ] finish/rework this docstring (and I think it's not in the right place anymore)

PR #18, opened 2024-05-23
Enable ethersync to work with multiple files

zormit:
This PR introduces the first iteration of a multi-file feature. We're assuming that files are "static" (i.e. won't be renamed, added or removed after starting Ethersync) for simplicity.

### DIFF
--- a/daemon/Cargo.lock
+++ b/daemon/Cargo.lock
@@ -81,6 +81,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0952808a6c2afd1aa8947271f3a60f1a6763c7b912d210184c5149b5cf147247"
 
 [[package]]
+name = "assert_matches"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
+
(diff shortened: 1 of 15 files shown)


blinry:
Pretty happy with this PR! :)

During refactoring, a pattern emerged: Often, it seems useful to pass a `(file_path: String, delta: TextDelta)` tuple.

We might want to give that type a name? So far, the refactored functions use the name "file delta". Another option could be `NamedTextDelta`.

zormit:
> Pretty happy with this PR! :)

Same! :)

> During refactoring, a pattern emerged: Often, it seems useful to pass a `(file_path: String, delta: TextDelta)` tuple.
> 
> We might want to give that type a name? So far, the refactored functions use the name "file delta". Another option could be `NamedTextDelta`.

I quite like `NamedTextDelta`, but I'm also fine with `FileDelta`. My perspective on this is, that this is an abstraction level, where we do carry around a part of the file name as an identifier, so it's more like a key or an id. Automerge uses `ObjId` for these kinds of things, I think.

What I'm saying is, that at the point of such a Delta it shouldn't matter anymore whether it's a file or not in most places. And in places where it matters, we have to produce "the actual filename" anyways. Which could be different once we support renames.

But for now anything is okay, as long as we understand what we mean :) A deeper understanding of this concept will probably form over time.

Issue #19, opened 2024-06-20
Learn + document how other libraries handles ephemeral data like cursor positions


blinry:
We probably have two main references we can look at: automerge-repo and Y.js!

blinry:
For inspiration, automerge-repo's ephemeral messages look like this:

```
{
  type: "ephemeral",
  ; The peer who sent this message
  senderId: peer_id,
  ; The target of this message
  targetId: peer_id,
  ; The sequence number of this message within its session
  count: uint,
  ; The unique session identifying this stream of ephemeral messages
  sessionId: str,
  ; The document ID this ephemera relates to
  documentId: document_id,
  ; The data of this message (in practice this is arbitrary CBOR)
  data: bstr
}
```

[This comment](https://github.com/automerge/automerge-repo/blob/caf338c97d8c2c669870e3d3efc34e0eabf3ca60/packages/automerge-repo/src/network/messages.ts#L34-L44) explains the reasoning behind some of the fields:

Peers "gossip" the ephemeral messages they receive around to their connected peers. To avoid endless loops, they need to be able to detect which ones they've already seen. To do this, each sending peer generates a random session id at startup, and adds sequence numbers to their messages, which allows other peers to discard messages they've already seen.

I'm not sure yet what the purpose of `targetId` is.

blinry:
Y.js's awarenes handling is implemented here: https://github.com/yjs/y-protocols/blob/master/awareness.js
More info about the API can be found here: https://docs.yjs.dev/api/about-awareness

It's a last-write-wins CRDT, which stores the latest state data for each peer under their ID. If peers don't receive an update in 30 seconds, their state is deleted.

blinry:
It's noticeable that both approaches have a general, unstructured "data" field for exchanging this ephemeral information, instead of defining "cursor" or "name" fields. Maybe we should also keep it general like that in the daemon-to-daemon communication, to allow later additions more easily (like user colors, for example).

blinry:
Hedgedoc passes simple messages [like this](https://github.com/hedgedoc/hedgedoc/blob/develop/commons/src/message-transporters/message.ts#L34-L45) around, working with [these datastructures](https://github.com/hedgedoc/hedgedoc/blob/develop/commons/src/message-transporters/realtime-user.ts). Very simple.

PR #23, opened 2024-06-21
Add draft of ADR-07 about cursor positions

blinry:
This encapsulates findings and insights from #19.

No decision made yet. To be discussed :) Maybe there are also more than two options?

### DIFF
--- a/docs/decisions/07-cursor-positions.md
+++ b/docs/decisions/07-cursor-positions.md
@@ -1,0 +1,62 @@
+---
+status: draft
+date: 2024-06-21
+---
+# How do we transmit cursor positions?
+
+## Context and Problem Statement
+
+An important feature of collaborative editing is being able to see each other's cursors. How should we implement transmitting them?
+
+## Decision Drivers
+
+Our solution should ideally be:
+
+* Fast
+* Correct (cursors never appear at wrong/shifted locations)
+* Simple (little protocol overhead, little complexity)
+* Take up little disk space
+
+## Considered Options
+
+* {title of option 1}
+* {title of option 2}
+* {title of option 3}
+* … <!-- numbers of options can vary -->
+
+## Decision Outcome
+
+Chosen option: "{title of option 1}", because
+{justification. e.g., only option, which meets k.o. criterion decision driver | which resolves force {force} | … | comes out best (see below)}.
+
+## Pros and Cons of the Options
+
+### Store user state in a map in the main Automerge document
+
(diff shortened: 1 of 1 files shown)


docs/decisions/07-cursor-positions.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Pros and Cons of the Options
+
+### Store user state in a map in the main Automerge document
+
+The Automerge document for an Ethersync project could be split into a `files` key (which contains the file contents like before), and a `peers` key (which stores information about the peers' state, like cursor positions).
+
+* Good, because transmiting this information would be trivial, with no work required from us at all. The Automege sync protocol handles it for us.
+* Bad, because this "ephemeral" data would become part of the CRDT, that each peer needs to persist. The persisted data would be larger as a result, and contain data that's completely irrelevant after a short amount of time.
+* Neutral, because it's unclear how to mark user state as "stale". Each peer could delete states after they haven't been updated after (for example) 30 seconds, but could this lead to "fights" between peers that have a connection to the peer, and ones that don't. And if we put the "last seen timestamp" in the CRDT, we assume that all daemons have the same wall time, which isn't necessarily true…
```
zormit, line 38: In my mind the sync would _attempt_ to communicate a last seen timestamp to the peers / editors, and those have to work with whatever they have. I think the "stale" logic could be part of the plugin, where each plugin decides how/when make a cursor disappear.

docs/decisions/07-cursor-positions.md
```diff
@@ -0,0 +1,62 @@
...
+
+Ideally, user states that have been overwritten anyway could be truncated in the Automerge history, but I don't think that's possible yet.
+
+Because Automerge uses "last writer wins" conflict resolution, the latest peer state should eventually end up in all peers.
+
+### Gossip the user states around, and store them in last-write-wins registers
+
+This would be the appraoch that automerge-repo and Y.js use. Peers send their state information to their connected peers, along with a session ID/user ID and an increasing sequence number, allowing the peers to discard messages they've already seen (and thus, avoiding endless loops). All peers would forward received state messages to all of their other peers, to make sure that the information is available in the whole network.
+
+* Good, because no data accumulates in some CRDT that needs to be persisted forever.
```
zormit, line 48: So the CRDT is more basic in a sense? I'm slightly confused about this statement in combination with the suggestion to use a CRDT crate. We need _some_ conflict resolution, right?

To confirm what you're saying here: Last Write Wins is technically already a CRDT, but not one that _persists its state forever_?

blinry: That's correct! :)

PR #25, opened 2024-06-24
Basic Cursor Awareness in Vim synchronized through automerge

blinry:
This PR adds a first iteration on cursor awareness. The decision on how to approach it can be found in ADR-07 (#23). This feature currently shows the cursor when it's there, but doesn't properly hide it again if a peer disconnects.

### DIFF
--- a/daemon/src/daemon.rs
+++ b/daemon/src/daemon.rs
@@ -293,6 +293,7 @@ impl DocumentActor {
                         PatchEffect::CursorChange(cursor_state) => {
                             cursor_states.push(cursor_state);
                         }
+                        PatchEffect::NoEffect => {}
                     }
                 }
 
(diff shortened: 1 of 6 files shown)


daemon/src/daemon.rs
```diff
@@ -133,19 +147,43 @@ impl Document {
             // This might change in a future more peer to peer world.
             panic!("It seems {file_path} was already initialized.");
         }
+
+        // TODO: I don't love the assumption that the first document to initialize a text
```
zormit, line 153: Same!

Maybe a bit hacky solution to this could be: What if we take the persisted initialized maps, bake them into the code and load them at each daemon that doesn't have a persisted history yet? Then all would have the same hashes and foundation. Like a database schema that is initialized.

zormit:
- [ ] daemon crashes, if editor is not running and it tries to communicate about cursors

PR #49, opened 2024-07-10
refactoring the vim plugin

zormit:
The ideas in this PR
- be more consistent in our coding style. I have used the https://github.com/luarocks/lua-style-guide as an inspiration.
	- switch from camelCase to snake_case where it made sense (it did in almost all places, `newText` being one exception)
	- cut conditionals a bit [more short](https://github.com/luarocks/lua-style-guide#conditional-expressions)
- add more linting (to the CI action as well as to pre-commit)
	- introduce `luacheck`
	    - note that I decided not to check for maximum line length, as I found [this argument](https://github.com/luarocks/lua-style-guide#line-lengths) convincing and I'm often a bit annoyed by having to cut lines. It does make sense in some places (like git commits), but here I'd be happy to comply with a style guide. 
- also I've exchanged the `stylua` github action to use a pre-compiled one instead of compiling it ourselves every time.

### DIFF
--- a/vim-plugin/lua/changetracker.lua
+++ b/vim-plugin/lua/changetracker.lua
@@ -10,7 +10,7 @@ local ignore_edits = false
 -- Subscribes the callback to changes for a given buffer id and reports with a delta.
 --
 -- The delta can be expected to be in the format as specified in the daemon-editor protocol.
-function M.trackChanges(buffer, callback)
+function M.track_changes(buffer, callback)
     -- Used to remember the previous content of the buffer, so that we can
     -- calculate the difference between the previous and the current content.
     local prev_lines = vim.api.nvim_buf_get_lines(buffer, 0, -1, true)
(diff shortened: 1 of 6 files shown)


vim-plugin/lua/changetracker.lua
```diff
@@ -2,15 +2,15 @@ local sync = require("vim.lsp.sync")
 local utils = require("utils")
 local debug = require("logging").debug
 
-local M = {}
+local changetracker = {}
```
zormit: I have seen a couple of plugins which use the module name as their "module table" and I wanted to try it. Happy to revert these changes again.

The upside is that it reads nicely in the function definitions, the downside is that it makes the code a bit less mobile if we want to restructure some functions.

I guess it comes down to personal taste and just having a consistent style. Also I think it somewhat can prepare towards a more object oriented approach, not just having a "capital M" everywhere, but by thinking more explicitly "this is a changetracker function".

blinry: I think I prefer the `M` approach, similar to how I prefer using `Self` in Rust whereever possible. Mostly, I think this makes renaming things easier/results in a smaller diff. But yes, it's a taste thing :)

PR #59, opened 2024-07-18
Implement adding and deleting files

blinry:
This is a simple solution for now:

- Create files by opening them in an Ethersync-enabled editor.
- Delete files by deleting them on disk; a file watcher will pick it up.


blinry:
Should we also use the file watcher to listen for file creations?

PR #66, opened 2024-07-23
Make connections using libp2p

zormit:
- [x] if we have old-disconnected connections, the daemon will try to send to them. They're not cleaned up properly.
- [x] Read secret from STDIN, instead of providing it on the command line. (we chose to create .ethersync/config instead)


blinry:
(Fuzzer doesn't run yet, because we can't access the libp2p multiaddr for connecting one peer to the other from the outside yet.)

blinry:
Note that, together with the "safer-file-io" branch (#68), this needs an adaption in that code that saves/read the key pairs.

zormit:
even though still highly experimental, it seems to work in the broad strokes. thus we agreed to merge.

PR #67, opened 2024-07-23
persist automerge state to disk take it as source to detect changes 

zormit:


daemon/Cargo.toml
```diff
@@ -31,6 +31,7 @@ public-ip = "0.2.2"
 tokio-util = "0.7.11"
 ignore = "0.4.22"
 notify = "6.1.1"
+difference = "2.0.0"
```
zormit: It's probably not a big problem, given that text diffing algorithms are not the most cutting edge and security-sensitive technology, but just noting that this package seems to be quite old and unmaintained

https://github.com/johannhof/difference.rs/issues/45
https://rustsec.org/advisories/RUSTSEC-2020-0095.html

zormit (commented):
haven't looked at all the details yet, but the idea makes sense :) Just leaving one comment about `difference` for now, which we might want to find an alternative for.

Issue #62, opened 2024-07-18
Make sure that the daemon can only read and write inside its base directory


blinry:
- [x] Safety checks for reading files
- [x] Safety checks for writing files
- [ ] #99

Test with paths like "../foo", for example.

blinry:
Here's how we can we files outside the project directory on another peer's device! Send these messages to the socket:

```
{"jsonrpc":"2.0","method":"open","params":{"uri":"file:///home/blinry/tmp/playground/../secret"}}
{"jsonrpc":"2.0","method":"edit","params":{"uri":"file:///home/blinry/tmp/playground/../secret","delta":{"revision":0,"delta":[{"range":{"start":{"line":0,"character":0},"end":{"line":0,"character":0}},"replacement":"hacked!!!!"}]}}}
```

The peer's daemon will write to (or create) the files without complaining.

We think an overall approach might be to have two functions for writing and reading files, which are very carefully written not to leave the project directory.

And to sanitize everything that comes in from outside our computer. Even though, interestingly, that's literally only Automerge sync messages right now. How would we sanitize those? :thinking:

blinry:
I also noticed that when running the "attack" described in [this post](https://github.com/ethersync/ethersync/issues/62#issuecomment-2244746525), the file "../secret" is still created in the CRDT. However, when a peer wants to write it to disk, this now fails.

But preventing the creation of file entries with "invalid names" would be good to avoid. Also those reaching us from the outside.

PR #68, opened 2024-07-23
Route all file/directory I/O through safe(r) functions in the safety module

blinry:
Addresses #62.

All uses of `fs::` are now in the security module, which check that the operations happen within the specified base directory. I think.

It's possible that other file operations might happen, which are not called via an `fs::` method… how could we find out?


blinry:
I also noticed that when running the "attack" described in [this post](https://github.com/ethersync/ethersync/issues/62#issuecomment-2244746525), the file "../secret" is still created in the CRDT. However, when a peer wants to write it to disk, this now fails.

But preventing the creation of file entries with "invalid names" would be good to avoid. Also those reaching us from the outside.

blinry:
An example of a file interaction that still isn't routed through the security module always is `file_path.exists()`.

PR #81, opened 2024-07-31
Some lints and refactorings

EdJoPaTo:
Ran a bunch of clippy lints, mainly pedantic ones and improved things I saw while looking at the code.

I tried to split them into their own commits. Feel free to cherry-pick or deny some of them. I can rebase and remove commits if wished for.

A good idea might also be to [configure lints in the `Cargo.toml`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-lints-section).

### DIFF
--- a/daemon/clippy.toml
+++ b/daemon/clippy.toml
@@ -1,0 +1,1 @@
+avoid-breaking-exported-api = false
--- a/daemon/src/bin/fuzzer.rs
+++ b/daemon/src/bin/fuzzer.rs
@@ -1,5 +1,6 @@
 #![allow(dead_code)]
 use ethersync::actors::{Actor, Neovim};
+use ethersync::connect::PeerConnectionInfo;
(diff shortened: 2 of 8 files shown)


EdJoPaTo:
> Do you have specific recommendations for which lints to activate in `Cargo.toml`? Would you enforce specific Clippy lints?

I would start with pedantic and allow the ones individually where it's ok when they warn (for example number conversions like `usize as u16` come up relatively often)

Personally I use a relatively long list of lints: https://github.com/EdJoPaTo/LinuxScripts/blob/85238cfcc0553ceaf7200101d9ad84fde290240e/configs/bin/cargo-pedantic#L29-L73

Ratatui for example does not include all nursery lints by default and individually includes the ones that seem relevant. The allowed ones should probably be removed and fixed / allowed individually :innocent: 
https://github.com/ratatui-org/ratatui/blob/b344f95b7cfc3b90221c84cfef8afcf6944bafc1/Cargo.toml#L66-L106

PR #91, opened 2024-08-02
Flake improvements

Sohalt:
* Add nixpkgs overlay
* Separately package nvim plugin
* Provide ethersync on nvim PATH


Sohalt:
Not strange at all. I just missed it. Re-added it just now :)

PR #96, opened 2024-08-06
nix improvements v2

zormit:
These changes were introduced in #91, but it turned out they are not yet functional. We missed to run the CI, and in fact it doesn't run. This has been fixed in the meantime, so we can try again to get the nix improvements through the door :)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,50 +8,66 @@
   };
 
   outputs = {
+    self,
     nixpkgs,
     naersk,
     ...
   }: let
+    supportedSystems = [
+      "x86_64-linux"
+      "aarch64-linux"
+      "x86_64-darwin"
+      "aarch64-darwin"
+    ];
+    lib = nixpkgs.lib;
     forAllSystems = function:
-      nixpkgs.lib.genAttrs [
-        "x86_64-linux"
-        "aarch64-linux"
-        "x86_64-darwin"
-        "aarch64-darwin"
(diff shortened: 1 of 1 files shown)


zormit:
@Sohalt unfortunately we had to revert the improvements 😢 as the CI didn't like them. It's a bit of a weird behavior, you can see an example in 854e1c87ac, where it just times out after 6 hours (🤯 btw).

There seems to be a tricky problem with the way nvim is behaving in the new vs. old environment. We couldn't make sense of it yet... But what our fuzzer tries to do is to "remote control" nvim, which seems to not happen. The daemon waits forever for a connection.

If you work on this, feel free to open _another_ PR or just reference your branch here, it's up to you :)

Sohalt:
If you try to run CI locally you see that the issue was in line 42 of `daemon/src/actor.rs`, because it failed to start nvim.
The issue was that the nvim wrapper script didn't pass arguments along to nvim.

zormit:
yeah we saw that problem locally but didn't succeed in fixing it.
```
thread 'main' panicked at /Users/mn/code/ethersync/ethersync/daemon/src/actors.rs:42:14:
Opening file in nvim failed: DecodeError(InvalidMessage(NotAnArray(Integer(PosInt(27)))), "nvim_command")
```

Also it's quite curious why the CI does just block and not fail.

Anyways, thanks! As it's fixed in #98 i'm closing this one.

PR #98, opened 2024-08-06
Flake improvements

Sohalt:
* Add nixpkgs overlay
* Separately package nvim plugin
* Provide ethersync on nvim PATH

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -64,10 +64,16 @@
         ;
       default = ethersync;
     });
-
     devShells = forAllSystems (pkgs: {
       default = pkgs.mkShell {
-        packages = with pkgs; [cargo rustc neovim];
+        packages =
+          (with pkgs; [cargo rustc neovim])
+          ++ (
+            # macOS systems seem to require these extra packages for building Rust code.
+            if (lib.strings.hasInfix "darwin" pkgs.system)
+            then (with pkgs; [darwin.apple_sdk.frameworks.CoreServices libiconv])
+            else []
+          );
       };
     });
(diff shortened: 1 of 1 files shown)


zormit:
Thanks for this!

Quoting you from #96 

> The issue was that the nvim wrapper script didn't pass arguments along to nvim.

it seems that the magic incantation

```
 final.writeShellScriptBin "nvim" ''
          PATH=${lib.makeBinPath [ethersync]}:$PATH ${nvim-custom}/bin/nvim $@'';
```
injects the correct paths now. Just writing this down for posterity and that I can understand what we did there later :)

--- 

I have another question. Being on a Mac I came across other difficulties, I need to add some dependencies
```
diff --git a/flake.nix b/flake.nix
index 2939900..d0e2e90 100644
--- a/flake.nix
+++ b/flake.nix
@@ -67,7 +67,7 @@

     devShells = forAllSystems (pkgs: {
       default = pkgs.mkShell {
-        packages = with pkgs; [cargo rustc neovim];
+        packages = with pkgs; [cargo rustc neovim darwin.apple_sdk.frameworks.CoreServices libiconv];
       };
     });
   };
```

I think it doesn't really make sense to add them for all systems... But I don't speak the nix language 😅 Could you (or maybe @blinry?) help me adding them only for darwin systems?

blinry:
We solved it like this for now:

https://github.com/Sohalt/ethersync/blob/c778a040bbb7e6c7e9652d66932e184890f536af/flake.nix#L67-L77

Issue #29, opened 2024-06-27
In "user mode", most panics in the daemon shouldn't actually kill the daemon


blinry:
But in "dev mode", it's nice to immediately see errors, and "fail early".

alerque:
Another one is a faulty editor that has opened the JSON-RPC stream and *should* be reading from it, but fails to do so when a message comes across will crash not only the client but also the remote daemon it is connected to.

alerque:
Once we address the panic cases we know we need to avoid it might be worth using something like [human_panic](https://docs.rs/human-panic/latest/human_panic/) to solicit feedback on the cases where we thought it was okay to leave something and it wasn't. At the moment I think it might be too noisy because we have a bunch of known panic situations that should be addressed before we get into the edge cases.

PR #109, opened 2024-08-08
When receiving faulty JSON-RPC on the socket, send an error back to the caller

blinry:
This should make it easier to develop editor plugins, when you get feedback that your message couldn't get parsed.

Partial fix for #29.

### DIFF
--- a/daemon/src/types.rs
+++ b/daemon/src/types.rs
@@ -139,8 +139,7 @@ pub enum JSONRPCFromEditor {
 }
 impl JSONRPCFromEditor {
     pub fn from_jsonrpc(jsonrpc: &str) -> Result<Self, anyhow::Error> {
-        let error_message = format!("Failed to deserialize editor message: {jsonrpc}");
-        let message = serde_json::from_str(jsonrpc).expect(&error_message);
+        let message = serde_json::from_str(jsonrpc)?;
         Ok(message)
     }
 }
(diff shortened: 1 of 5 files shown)


daemon/src/types.rs
```diff
@@ -139,8 +139,8 @@ pub enum JSONRPCFromEditor {
 }
 impl JSONRPCFromEditor {
     pub fn from_jsonrpc(jsonrpc: &str) -> Result<Self, anyhow::Error> {
-        let error_message = format!("Failed to deserialize editor message: {jsonrpc}");
-        let message = serde_json::from_str(jsonrpc).expect(&error_message);
+        //let error_message = format!("Failed to deserialize editor message: {jsonrpc}");
+        let message = serde_json::from_str(jsonrpc)?; //.context(error_message)?;
```
zormit: Should we remove the comments here, aka, are you happy with the "question mark only"? Or was this something not yet working?

zormit (approved):
looks good to me, just left one comment on the comment :)

(yay, an `expect` less! =))

daemon/src/types.rs
blinry: Ah no, it was an oversight. I removed the comments.

blinry:
Ideas for how to debug the failing fuzzer further:

- [ ] Enable debugging in the fuzzer, and look hard at the output?
- [ ] Check if it's always the case that daemon == nvim and peer == nvim2 (is it a peering problem?)
- [ ] Change the fuzzer so that it runs relatively fast locally (decrease sleeps), but increase number of steps, and with that configuration, bisect and see where it starts failing

zormit:
I think this one can be merged as well now :)

PR #110, opened 2024-08-08
vim: Only start tracking buffer if "open" succeeds

### LINKED ISSUE #88: vim: When an "open" errors, unregister buffer and don't send "edit" messages
Maybe extend `send_request` with an error callback?

blinry:
Fixes #88.

### DIFF
--- a/vim-plugin/plugin/ethersync.lua
+++ b/vim-plugin/plugin/ethersync.lua
@@ -129,6 +129,8 @@ local function track_edits(filename, uri)
     end)
     cursor.track_cursor(0, function(ranges)
         local params = { uri = uri, ranges = ranges }
+        -- Even though it's not "needed" we're sending requests in this case
+        -- to ensure we're processing/seeing potential errors.
         send_request("cursor", params)
     end)
 end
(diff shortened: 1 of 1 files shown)


vim-plugin/plugin/ethersync.lua
```diff
@@ -106,38 +113,37 @@ local function on_buffer_open()
...
-        send_request("edit", params)
-    end)
-    cursor.track_cursor(0, function(ranges)
-        local params = { uri = uri, ranges = ranges }
-        send_notification("cursor", params)
+            send_request("edit", params)
+        end)
+        cursor.track_cursor(0, function(ranges)
+            local params = { uri = uri, ranges = ranges }
+            send_request("cursor", params)
```
zormit: was this turned to a `send_request` on purpose? if, how so?

blinry: For the time being, I thought it might be interesting to see if our "cursor" messages fail, as well (for example, because they're being sent for ignored files).

zormit: makes sense :) I added a comment to it.

zormit (approved):

Issue #77, opened 2024-07-30
Emacs plugin


peacememories:
Hi, this project seems super interesting, especially as a multi-editor collaborative platform that seems to be focused on being straightforward to implement for editors. Kind of like the language server of collaborative editing I guess :P

I would be very interested in having an Emacs plugin, since that's the main editor I use nowadays, and the only reason I switch over to VSCode is to collaborate with colleagues who use it.

If this is not already in the works by someone I think I'll start trying to write one. Keep your expectations low, though, it'd be my first Emacs plugin :sweat_smile:

cosmicexplorer:
oh wow, those are almost exactly the same message types i had in my prior attempt at doing this sort of thing! super straightforward! @blinry how is the daemon invoked or connected to? can we expect the user to handle that or should we consider packaging a version of the daemon with our extension? can there be multiple daemons on one machine at once? what is the desired interface between the end user and configuring the daemon? should it be hidden as an implementation detail or make it more pluggable? how stable is the daemon invocation (cli?) interface and how does that compare to the stability of the json messages from the daemon?

super exciting work!!

zormit:
@cosmicexplorer thanks for these really good questions!

> how is the daemon invoked or connected to?

The current assumption is that the daemon is manually started. It might get systemd support or something similar in the near future, but we're not they're yet. So to start it, do something around the lines of `ethersync daemon`, as explained in the README.

> can we expect the user to handle that or should we consider packaging a version of the daemon with our extension?

We would expect that an installation of the daemon already exists. This should not be the concern of the editor plugin developer, packaging it with the extension is not what we imagined, but maybe we'll introduce a pattern/process for the editor plugin to make sure the daemon is actually running and if not start it.

> can there be multiple daemons on one machine at once?

Currently, and we haven't polished this yet, the assumption is one daemon per project. It's somewhat tricky to start a second daemon: You need to use another socket. The `ethersync daemon` has a parameter for this, and the `ethersync client` takes it from the `ETHERSYNC_SOCKET` env var.

> what is the desired interface between the end user and configuring the daemon? should it be hidden as an implementation detail or make it more pluggable?

What are you imagining to configure here. I would say it shouldn't be 

> how stable is the daemon invocation (cli?) interface and how does that compare to the stability of the json messages from the daemon?

The daemon CLI is currently not very stable. We are changing it from iteration to iteration to refine it.
[...]

PR #112, opened 2024-08-10
Emacs plugin

cosmicexplorer:
Follow up from #77. This *would* be working and printing messages, but I have absolutely no idea how to get the daemon to write anything to the unix socket.

# Confusion on how to test
This section https://github.com/ethersync/ethersync/blob/main/docs/editor-plugin-dev-guide.md#seeing-what-an-existing-ethersync-plugin-sends which simply says to start up the neovim plugin is extremely opaque. I have tried using the `tools/dummy-jsonrpc.py` script:
```bash
> python ../tools/dummy-jsonrpc.py --message-type open "$(readlink -f playground/file.md)" | ./target/release/ethersync client --debug
```

And we get this output from the daemon itself:
```bash
21:02:04 DEBUG ThreadId(17) ethersync::daemon: Handling RevDelta from editor: RevisionedEditorTextDelta {
    revision: 0,
    delta: EditorTextDelta(
        [
            EditorTextOp {
                range: Range {
                    start: Position {
                        line: 0,
                        character: 0,
                    },
                    end: Position {
                        line: 0,
                        character: 0,
                    },
                },
                replacement: "hello, world",
            },
        ],
    ),
}
```

But for the life of me I have not been able to make the daemon write *to* the socket, such that reading from `/tmp/ethersync` with `socat` or `ethersync client` ever produces any output. I was able to get a local http peer attached, but that did not seem to produce any change.

I notice that when a file is changed after the daemon is killed, it notices this and prints out a warning when it is restarted. Is the daemon also supposed to be tracking changes made to files and sending them somewhere? It looks like there is code to watch for file change notifications--when is that activated? How does the daemon incorporate both editor changes as well as filesystem changes?

It seems like I don't understand the concept of [file ownership](https://github.com/ethersync/ethersync/blob/main/docs/editor-plugin-dev-guide.md#file-ownership). It's very much not clear to me how the daemon is supposed to be sending changes through it when writing to the stdin of `ethersync client` does not modify file contents, and modifying the file contents does not seem to get the daemon to write anything to the `/tmp/ethersync` socket.

I would really appreciate some more clarity on what "ownership" means, and I would particularly like to understand how to get the daemon to actually write anything to the socket, so I can actually test my editor integration. Or is the editor mostly supposed to be writing to the daemon? But then how does the editor receive inputs from others? Or is the editor supposed to be constantly refreshing the file from disk after the daemon modifies it? I assume not, but then how am I supposed to send input to the daemon so that it will send output to my editor?

I suppose I will be diving further into the rust code soon in order to figure all of this out. The editor plugin dev guide describes how to watch the daemon socket, but not how to get the daemon to actually write anything to the socket, except by telling the reader to execute the neovim plugin. I would appreciate an explanation of how to make the daemon write output to the socket so I can test my editor integration. Thanks.

### DIFF
--- a/emacs-plugin/ethersync.el
+++ b/emacs-plugin/ethersync.el
@@ -42,16 +42,20 @@
 
 (require 'cl-lib)
 (require 'dash)
+(require 'files)
 (require 'jsonrpc)
 (require 'pcase)
 (require 'rx)
-(require 'subr-rx)
+(require 'subr-x)
 
 
 ;; Customization Helpers
 (defun ethersync--always-safe-local (_)
   "Use as a :safe predicate in a `defcustom' form to accept any local override."
   t)
+
+(defconst ethersync--default-socket-name "ethersync"
+  "Name of socket file.")
 
 
 ;; Public error types
@@ -68,7 +72,8 @@
   :group 'files
   :group 'shadow)
 
-(defcustom ethersync-socket-location "/tmp/ethersync"
+(defcustom ethersync-socket-location
+  (expand-file-name ethersync--default-socket-name (temporary-file-directory))
   "File path to the unix domain socket used by ethersync."
   :type 'file
   :safe #'ethersync--always-safe-local
@@ -119,7 +124,7 @@ reading and writing JSONRPC messages with basic HTTP-style enveloping headers su
 \"Content-Length:\".")
 
 (cl-defmethod initialize-instance :after ((client ethersync-client)
-                                          &key process &allow-other-keys)
+                                          (&key process &allow-other-keys))
   (process-put process 'jsonrpc-connection client))
 
 (cl-defmethod ethersync--client-socket-path ((client ethersync-client))
@@ -163,7 +168,7 @@ reading and writing JSONRPC messages with basic HTTP-style enveloping headers su
   (unless (file-exists-p socket-location)
     (signal 'ethersync-client-connect-error
             `("[ethersync] socket path does not exist" ,socket-location)))
-  (unless (ethersync--file-is-socket path)
+  (unless (ethersync--file-is-socket socket-location)
     (signal 'ethersync-client-connect-error
             `("[ethersync] path is not a socket" ,socket-location))))
 
@@ -178,7 +183,7 @@ reading and writing JSONRPC messages with basic HTTP-style enveloping headers su
 ;; Logic
 (defun ethersync--create-client-process (socket-location)
   (ethersync--assert-socket-path socket-location)
-  (let* ((new-buffer-name (generate-new-buffer ethersync--process-buffer-name t))
+  (let* ((new-buffer-name (generate-new-buffer ethersync--process-buffer-name-base t))
          (socket-proc (make-network-process
                        :name ethersync--process-name
                        :buffer new-buffer-name
@@ -197,8 +202,37 @@ reading and writing JSONRPC messages with basic HTTP-style enveloping headers su
 
 (defun ethersync--process-sentinel (proc change))
 
+(defun ethersync--scan-for-matching-socket-path (socket-path)
+  "See if a client has already been initialized to SOCKET-PATH, and return it if so."
+  (cl-loop for client in ethersync--active-clients
+           for prev-path = (ethersync--client-socket-path client)
+           if (file-equal-p socket-path prev-path)
+           return client))
+
+(defun ethersync--completing-read-socket-path (socket-filename)
+  "Read in the file path for the ethersync socket."
+  (let ((base-dir (temporary-file-directory)))
+    (expand-file-name
+     (read-file-name "ethersync socket path: "
+                     base-dir socket-filename
+                     t socket-filename
+                     #'ethersync--file-is-socket)
+     base-dir)))
+
 
 ;; Autoloaded functions
+;;;###autoload
+(defun ethersync-connect (&optional socket-location)
+  "Connect to a socket at SOCKET-LOCATION, or return the existing client for that socket."
+  (interactive "P")
+  (let ((socket-location
+          (pcase-exhaustive socket-location
+            (`nil ethersync-socket-location)
+            ((and (pred stringp) s) s)
+            ('(4) (ethersync--completing-read-socket-path ethersync--default-socket-name)))))
+    (or (ethersync--scan-for-matching-socket-path socket-location)
+        (push (ethersync--create-client-process socket-location) ethersync--active-clients))))
+
 
 (provide 'ethersync)
 ;;; ethersync.el ends here
--- a/emacs-plugin/ethersync.el
+++ b/emacs-plugin/ethersync.el
@@ -233,6 +233,16 @@ reading and writing JSONRPC messages with basic HTTP-style enveloping headers su
     (or (ethersync--scan-for-matching-socket-path socket-location)
         (push (ethersync--create-client-process socket-location) ethersync--active-clients))))
 
+;;;###autoload
+(defun ethersync-clear-all-clients ()
+  "Iterate through every attached client and kill its process."
+  (interactive)
+  (cl-loop for client = (pop ethersync--active-clients)
+           while client
+           for sock-proc = (ethersync--client-process client)
+           for proc-buf = (process-buffer sock-proc)
+           do (kill-buffer proc-buf)))
+
 
 (provide 'ethersync)
 ;;; ethersync.el ends here
--- a/emacs-plugin/ethersync.el
+++ b/emacs-plugin/ethersync.el
@@ -115,17 +115,30 @@ An initial space is used to hide the buffer from the buffer list.")
 (defclass ethersync-client (jsonrpc-connection)
   ((-process
     :initarg :process :type process :accessor ethersync--client-process
-    :documentation "Process object wrapped by this client."))
+    :documentation "Process object wrapped by this client.")
+   (-expected-bytes
+    :initform nil
+    :accessor jsonrpc--expected-bytes
+    :documentation "How many bytes declared by server.")
+   (-on-shutdown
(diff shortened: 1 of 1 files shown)


cosmicexplorer:
Spending some time diving into the rust code today!! If successful, will likely be developing (1) a description of data flow (2) some testing tools.

blinry:
Good luck! Your questions are really good, and will probably help improve the documentation a lot. We're just not very active on weekends, usually.

Very briefly: Having ownership means being allowed to write to the file. If an editor opens a file, we don't want the daemon to write over it, otherwise, editors are usually unhappy :P

And you can get the daemon to output stuff by connecting a second daemon. This is a bit tricky, unfortunately. You could either use a second machine, connect to the daemon on the first machine, and use your Emacs plugin (or the Neovim plugin) to edit a file.

Or, on the same machine, you can use a command like `ethersync daemon --peer localhost:<port> --socket-path /tmp/etherbonk` to start a second daemon. If you then set the `ETHERSYNC_SOCKET` environment variable, `ethersync client` will connect to that daemon's socket instead, allowing you to trigger edits.

Hope that helps a bit!

cosmicexplorer:
Thanks so much!! <3

cosmicexplorer:
> We're just not very active on weekends, usually.

Thanks for the note! I really appreciate the incredibly prompt and useful response!

cosmicexplorer:
> Very briefly: Having ownership means being allowed to write to the file. If an editor opens a file, we don't want the daemon to write over it, otherwise, editors are usually unhappy :P

Thanks so much for this brief and effective description, that part made a lot of sense to me. Emacs has a ["shadow file"](https://www.gnu.org/software/emacs/manual/html_node/emacs/File-Shadowing.html) feature which syncs whole files across machines, but it's not live and doesn't keep track of individual edits. I mention this just because it's an interesting thought experiment, and I haven't found any way to actually synthesize it with ethersync's live editing via CRDT. And definitely, if file content changes while a user is actively editing it, there is really no way to put them back where they want to be, especially if e.g. a ton of content is added to the beginning of the file.

So it totally makes sense now that putting a file under editor control means taking it out of that file-based paradigm and bringing it into our synchronized editing paradigm!

I think I hadn't really considered at all before the feature you've already implemented (file syncing via the daemon in the background). It makes perfect sense when you consider an ethersync directory as a repository of synchronized state: editor actions for foreground, and raw file syncing with the daemon when the editor has closed the file. What this provides of course is a really really nice UX, where individual users don't need to share a link or negotiate a key for every single document they're working on, but rather they can share a keyphrase just once, and then allow the daemon to do the work of syncing.

I can see why NGI funded this work. This is a really fascinating approach that I had never once considered in my previous attempt at a similar piece of software ([emdocs](https://github.com/cosmicexplorer/emdocs)). Thanks so much for taking the time to explain it; I was confused because I was thinking in the mode that every single new file must be negotiated, whereas ethersync's model maps more closely to the way humans actually work and code. Very very very cool project.

cosmicexplorer:
I also just noticed the `docs/decisions/` directory -- thanks so much for that, will be checking that out as well.

cosmicexplorer:
@blinry two questions on this part:
> If you then set the `ETHERSYNC_SOCKET` environment variable, `ethersync client` will connect to that daemon's socket instead, allowing you to trigger edits.

1. Is `ETHERSYNC_SOCKET` the same thing as providing `ethersync client --socket-path <...>`?
2. Where is `ETHERSYNC_SOCKET` being read from? I know [`clap` can read from environment variables](https://docs.rs/clap/latest/clap/struct.Arg.html#method.env), but I don't see where that's being done here, except in the neovim plugin itself? Would it be useful to add that functionality (filling in `--socket-path` from `ETHERSYNC_SOCKET`) to the ethersync clap argument parsing itself, instead of patching it into the editor plugin?

zormit:
Regarding `docs/decisions`, we definitely don't have all decisions documented in there, but it's a good find, hopefully helpful and let us know if not :)

Regarding the socket questions: The environment variable is use it [the nvim plugin](https://github.com/ethersync/ethersync/blob/main/vim-plugin/plugin/ethersync.lua#L69-L72). It was the easiest way for us to influence its behavior without having to fiddle with the lua config of the plugin every time.

So it would be nice if you would implement a similar behavior that `ETHERSYNC_SOCKET` gets used as the parameter to `ethersync client --scoket-path` in emacs.

Here's a little sketch that I just drew with the hope it makes things even clearer:
image

Left side is default, right side is using the environment variable.

cosmicexplorer:
Broke out #127 for reading `ETHERSYNC_SOCKET` in the `ethersync` binary, let me know if I've misunderstood!

cosmicexplorer:
To a similar point, I think it would be quite useful to consider a request-response method for the client to request the current list of daemon-tracked files, or perhaps just send it immediately to the client upon startup (although I would prefer the former). This would be similar in spirit to the proposed ["status" subcommand in the CLI UI doc](https://github.com/ethersync/ethersync/blob/main/docs/ethersync-cli-ui.md#subcommands).

While it doesn't appear that there's currently any logic to ignore files in the tracked directory, if we do end up wanting to support that feature, it would avoid having to duplicate that logic in the editor plugins. I was going to say that "instead, the editor could just send an `open` message and the daemon would let it know whether the file will be tracked", but I realize that changing the `open` message from a "notification" into a "request" would be a breaking API change. I think it could make sense, though, especially if you're using ethersync for code collaboration and you want to avoid it syncing changes to build output (does ethersync currently avoid this? how?).

blinry:
> Having some difficulty getting two daemon instances on the same node (different unix sockets), to talk to one another, largely around "maximum frame length exceeded", "timeout", and "unexpected eof":

Honestly not sure what the issue is there/how to fix it. We haven't seen that in our user testing so far.

Have you been able to run the fuzzer locally yet? It conceptually does a very similar thing.

> This is the same ethersync binary started up, listening to the same directory. Is there some sort of resource contention on the `.ethersync` directory (the two are both pointing to the same directory on the local filesystem) that I might be causing? Should I be testing these against two separate directories and letting ethersync handle syncing those?

I'm starting a new experiment: For each of the (very good and valid) questions you ask, I'll try to improve the documentation, and link you to it. Please let us know if the docs still aren't clear or could be expanded on!

Yes, we've always been using a separate directory per peer, see: https://github.com/ethersync/ethersync/blob/main/docs/editor-plugin-dev-guide.md#starting-two-local-daemons

This could well be related to your connection problems. Especially, the daemon will persist its public/private key pair in `.ethersync/key`, and it's probably not good if both daemon share it...

> More generally, how will two ethersync daemons resolve conflicts if the files aren't the same upon startup? Does the one providing the remote address decide to accept the remote as the source of truth upon startup?

https://github.com/ethersync/ethersync/blob/main/README.md#can-i-make-changes-to-a-shared-directory-while-the-daemon-isnt-running

> To a similar point, I think it would be quite useful to consider a request-response method for the client to request the current list of daemon-tracked files, or perhaps just send it immediately to the client upon startup (although I would prefer the former). This would be similar in spirit to the proposed ["status" subcommand in the CLI UI doc](https://github.com/ethersync/ethersync/blob/main/docs/ethersync-cli-ui.md?rgh-link-date=2024-08-15T20%3A12%3A47Z#subcommands).

Yeah, having something like that in the editor protocol in the long term seems useful, especially once we get closer to multi-project daemons (#134).

> While it doesn't appear that there's currently any logic to ignore files in the tracked directory, if we do end up wanting to support that feature, it would avoid having to duplicate that logic in the editor plugins. I was going to say that "instead, the editor could just send an open message and the daemon would let it know whether the file will be tracked", but I realize that changing the open message from a "notification" into a "request" would be a breaking API change.

We've actually started using JSON-RPC requests a couple of weeks ago exactly for that reason. I've updated the spec in 1544dca1029ef4ae2e2486ca25f6e1e00b3fd3d8!

> I think it could make sense, though, especially if you're using ethersync for code collaboration and you want to avoid it syncing changes to build output (does ethersync currently avoid this? how?).

We try to! Ethersync uses the "ignore" crate and [`sandbox::ignored()`](https://github.com/ethersync/ethersync/blob/watch-file-creation-and-modification/daemon/src/sandbox.rs#L56-L89) to determine whether files should be ignored. We will reject "open"-ing them.

Untracked files, however, currently *are* shared, which we've noticed in our dogfeeding is not great.

Again, thanks for your questions! :)

zormit:
> This is the same ethersync binary started up, listening to the same directory. Is there some sort of resource contention on the .ethersync directory (the two are both pointing to the same directory on the local filesystem) that I might be causing? Should I be testing these against two separate directories and letting ethersync handle syncing those?

blinry has already answered this question, I thought it might still be helpful to elaborate on it a bit from my perspective: Yeah, you definitely need to use two different directories! Starting two daemons on the same directory is probably quite undefined behavior and resource contention might be a good word for that :D

When you think of the two daemons, think of them as two users who have their own files. I usually do it like this (which is as blinry explains, but a tiny bit more hands on. I didn't test these commands now, so still take it as an example):

### Terminal 1 (first daemon)

  ```bash
mkdir -p playground1/.ethersync
ethersync daemon playground1 --secret locally-whatever
```
=> take multiaddress as peer in the next step, as it seems you already did.

### Terminal 2 (second daemon)

```bash
# optional: If you set a different $USER envvar here, you will be able to have even a different cursor name for that peer :) At least when testing with nvim.
mkdir -p playground2/.ethersync
ethersync daemon --peer <multiaddr from above> -s /tmp/ethertwo playground2 --secret locally-whatever
```

### Terminal 3 (first daemon-client)
```bash
cd playground1
vim file
```

### Terminal 4 (second daemon-client)
```bash
cd playground2
export ETHERSYNC_SOCKET=/tmp/ethertwo
vim file
```

PR #124, opened 2024-08-12
Speed up CI by dropping Nix and using "rust-cache"

blinry:
Also, to fix the CI tests, add another sleep to give the first daemon time to boot.


blinry:
Sigh…



But giving various steps in the fuzzer 5 seconds to boot seems to help now!

blinry:
[Narrator voice] "…it didn't"

PR #126, opened 2024-08-13
Instead of using a CancellationToken, directly abort the SyncActor task

### LINKED ISSUE #125: Fuzzer is broken
> I can also reproduce the failing fuzzer locally sometimes (1/5 runs or so?).
>
> Ideas for how to proceed:
>
> - [ ] Enable debugging in the fuzzer, and look hard at the output?
> - [ ] Log errors in the Vim plugin, and see if there's a failing JSON-RPC request there?
> - [ ] Check if it's always the case that daemon == nvim and peer == nvim2 (is it a peering problem?)
> - [x] Change the fuzzer so that it runs relatively fast locally (decrease sleeps), but increase number of steps, and with that configuration, bisect and see where it starts failing. Seemed to be around the new "open" behaviour, right?
> - [ ] Consider that's is a real, deep issue that leads to incorrect results. It's just surprising that we started seeing this a couple of days ago, and the fuzzer seemed fine before that.

_Originally posted by @blinry in https://github.com/ethersync/ethersync/issues/109#issuecomment-2284792055_

blinry:
It's unclear to me why, but using .cancelled() in the select! seems to block the other branches? Maybe using it in a loop is bad for some reason?

Hopefully fixes #125.

### DIFF
--- a/daemon/src/bin/fuzzer.rs
+++ b/daemon/src/bin/fuzzer.rs
@@ -64,6 +64,8 @@ async fn main() {
     sleep(std::time::Duration::from_millis(5000)).await;
 
     let nvim = Neovim::new(file).await;
+    // Give the editor time to process the open.
+    sleep(std::time::Duration::from_millis(5000)).await;
 
     let peer = Daemon::new(
         PeerConnectionInfo {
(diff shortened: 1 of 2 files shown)


zormit:
Thanks for figuring this out! Cool to see another way of cancelling / aborting a task :)

PR #127, opened 2024-08-14
make --socket-path pull ETHERSYNC_SOCKET from env via clap

cosmicexplorer:
# Problem
It's much easier to set `ETHERSYNC_SOCKET` via the environment for many use cases instead of modifying the command line, as noted by @zormit in https://github.com/ethersync/ethersync/pull/112#issuecomment-2287008800. We currently do this by hand in the nvim plugin: https://github.com/ethersync/ethersync/blob/a41304aa0543077bcf8908d912dabf8bc91069cf/vim-plugin/plugin/ethersync.lua#L69-L72

We could do the same for the emacs plugin in #112, but it might be nice for the `ethersync` binary to support this itself.

# Solution
- Add `"env"` feature to our `clap` dependency.
- Set [`Arg::env()`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.env) for the global `socket_path: PathBuf` argument.
    - A command-line `--socket-path` argument overrides this value; the env var is only read if the argument is not provided.
    - *Note that the daemon is now also affected by `ETHERSYNC_SOCKET` in addition to the client.*
- Remove code in nvim plugin to translate the `ETHERSYNC_SOCKET` env var into `--socket-path`.

# Result
All editor plugins (and the fuzzer) can expect to be able to set the `ETHERSYNC_SOCKET` environment variable as a separate way to parameterize the ethersync client, and users can set `ETHERSYNC_SOCKET` in their environment when invoking the daemon.

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -1,5 +1,5 @@
 use anyhow::{Context, Result};
-use clap::{Parser, Subcommand};
+use clap::{parser::ValueSource, CommandFactory, FromArgMatches, Parser, Subcommand};
 use ethersync::peer::PeerConnectionInfo;
 use ethersync::{daemon::Daemon, logging, sandbox};
 use std::path::{Path, PathBuf};
@@ -11,6 +11,7 @@ mod jsonrpc_forwarder;
 const DEFAULT_SOCKET_PATH: &str = "/tmp/ethersync";
 const ETHERSYNC_CONFIG_DIR: &str = ".ethersync";
 const ETHERSYNC_CONFIG_FILE: &str = "config";
+const ETHERSYNC_SOCKET_ENV_VAR: &str = "ETHERSYNC_SOCKET";
 
 #[derive(Parser)]
 #[command(version, about, long_about = None)]
@@ -19,10 +20,14 @@ struct Cli {
     #[command(subcommand)]
     command: Commands,
     /// Path to the Unix domain socket to use for communication between daemon and editors.
-    #[arg(short, long, global = true, default_value = DEFAULT_SOCKET_PATH)]
+    #[arg(
+      short, long, global = true,
+      default_value = DEFAULT_SOCKET_PATH,
+      env = ETHERSYNC_SOCKET_ENV_VAR,
+    )]
     socket_path: PathBuf,
     /// Enable verbose debug output.
-    #[arg(short, long, global = true, action)]
+    #[arg(short, long, global = true)]
     debug: bool,
 }
 
@@ -64,7 +69,11 @@ async fn main() -> Result<()> {
         std::process::exit(1);
     }));
 
-    let cli = Cli::parse();
+    let matches = Cli::command().get_matches();
+    let cli = match Cli::from_arg_matches(&matches) {
+        Ok(cli) => cli,
+        Err(e) => e.exit(),
+    };
 
     logging::initialize(cli.debug);
 
@@ -78,6 +87,14 @@ async fn main() -> Result<()> {
             mut secret,
             init,
         } => {
+            if matches.value_source("socket_path").unwrap() == ValueSource::EnvVariable {
+                info!(
+                    "Using socket path {} from env var {}",
+                    socket_path.display(),
+                    ETHERSYNC_SOCKET_ENV_VAR
+                );
+            }
+
             let directory = directory
                 .unwrap_or_else(|| {
                     std::env::current_dir().expect("Could not access current directory")
--- a/vim-plugin/plugin/ethersync.lua
+++ b/vim-plugin/plugin/ethersync.lua
@@ -66,11 +66,6 @@ local function connect()
 
     local params = { "client" }
 
-    local socket_path = os.getenv("ETHERSYNC_SOCKET")
-    if socket_path then
-        table.insert(params, "--socket-path=" .. socket_path)
-    end
-
     local dispatchers = {
         notification = function(method, notification_params)
             process_operation_for_editor(method, notification_params)
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -7,7 +7,7 @@ default-run = "ethersync"
 
 [dependencies]
 serde_json = { version = "1" }
-clap = { version = "4.5.3", features = ["derive"] }
+clap = { version = "4.5.3", features = ["derive", "env"] }
 automerge = "0.5.9"
 rand = "0.8.5"
 tokio = { version = "1", features = ["full"] }


daemon/src/main.rs
```diff
@@ -19,10 +19,14 @@ struct Cli {
...
     /// Path to the Unix domain socket to use for communication between daemon and editors.
-    #[arg(short, long, global = true, default_value = DEFAULT_SOCKET_PATH)]
+    #[arg(
+      short, long, global = true,
+      default_value = DEFAULT_SOCKET_PATH,
+      env = "ETHERSYNC_SOCKET",
+    )]
     socket_path: PathBuf,
     /// Enable verbose debug output.
-    #[arg(short, long, global = true, action)]
```
cosmicexplorer, line 25: This doesn't seem to change behavior, as can be seen by the output of `ethersync --help`. If the intent was to allow `--debug=true`, I think you would need something like `action(clap::ArgAction::Set), default_value = "false"`, but I'm having difficulty figuring out how to make that work.

If the intent was just to make a simple boolean flag `-d`/`--debug`, I think this `action` argument is redundant. Let me know if I'm mistaken!

blinry: You seem right! Thanks for removing the redundancy! :)

daemon/src/main.rs
```diff
@@ -69,6 +73,7 @@ async fn main() -> Result<()> {
     logging::initialize(cli.debug);
 
     let socket_path = cli.socket_path;
+    info!("Using socket path {}", socket_path.display());
```
blinry: I noticed that `ethersync client` will also print this info log. While this is helpful for humans, because editor plugins use the command to "speak" JSON-RPC, I think it'd be good to not show it for the `client` subcommand.

For the daemon, maybe we could print it only if read from the environment variable? (That's the "surprising" case.) Not sure if it's easy to figure that out after Clap has run.

vim-plugin/plugin/ethersync.lua
```diff
@@ -66,11 +66,6 @@ local function connect()
 
     local params = { "client" }
 
-    local socket_path = os.getenv("ETHERSYNC_SOCKET")
-    if socket_path then
-        table.insert(params, "--socket-path=" .. socket_path)
-    end
-
```
blinry, line 73: I really like that we can get rid of this! :) Props to you for noticing that. Takes away a bit of complexity in the editor plugins.

daemon/src/actors.rs
```diff
@@ -25,6 +25,7 @@ impl Neovim {
     /// Will panic if Neovim cannot be started or if the file cannot be opened.
     pub async fn new(file_path: PathBuf) -> Self {
         let handler = Dummy::new();
+        /* TODO: consider a process to bootstrap nvim for testing in CI. */
```
blinry: There is a (slightly involved) way: We provide a Nix flake that'll make a Neovim with Ethersync plugin available.

You can try it using `nix run github:ethersync/ethersync#neovim`, or put it in your PATH using `nix shell github:ethersync/ethersync#neovim`.

But it's true that it'd be nice to allow people to just run all tests without having to think about this.

daemon/src/bin/fuzzer.rs
```diff
@@ -33,6 +33,8 @@ fn initialize_project() -> (temp_dir::TempDir, PathBuf) {
     (dir, file)
 }
 
+/* TODO: consider moving fuzzer to separate crate to avoid adding test code to ethersync binary. */
+/* TODO: consider using cargo-fuzz and exposing a stdin interface to avoid network overhead. */
```
blinry: I think it's nice that we can test the whole "stack" here – it's our "integration test", if you will. So testing this through the network is part of the point here! :)

daemon/src/bin/fuzzer.rs
```diff
@@ -33,6 +33,8 @@ fn initialize_project() -> (temp_dir::TempDir, PathBuf) {
     (dir, file)
 }
 
+/* TODO: consider moving fuzzer to separate crate to avoid adding test code to ethersync binary. */
```
blinry: Maybe an actual [integration test](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html) would be the correct place for this! This would also avoid the problem of having a "fuzzer" binary in our output if you run `cargo build`. I created #128 to track this. Feel free to remove this TODO :)

blinry (changes requested):
Thanks for the thoughtful PR! Very cool that clap has a built-in feature for this!

daemon/src/main.rs
zormit: Good catch. The intent was the latter, I'm not sure why we added it in the first place if it does nothing. I think a possible alternative is to make it a count by repeating it, like [here](https://docs.rs/clap/latest/clap/_derive/_tutorial/chapter_0/index.html).

daemon/src/main.rs
cosmicexplorer: Not quite *after* clap has run, but rather as part of it, we can extract `ArgMatches`! I've done this (only for the daemon subcommand). Let me know what you think ^_^!

vim-plugin/plugin/ethersync.lua
cosmicexplorer: ^\_^\_^\_^\_^\_^\_^\_^\_^!!!!

daemon/src/actors.rs
cosmicexplorer: I also didn't notice that your github actions workflow installs it from a tarball!! I first actually left a TODO mentioning [`spack-rs`](https://docs.rs/spack-rs) ([repo](https://github.com/cosmicexplorer/spack-rs)), my cargo wrapper for [`spack`](https://github.com/spack/spack) which can bootstrap anything spack can, including neovim. However, there's some [optimization for bootstrapping](https://github.com/spack/spack/discussions/44714) I have yet to do, and `spack-rs` in particular is for linking libraries--no reason to use it  over `spack` itself for executables.

spack is a really interesting and useful project (build any codebase on any platform without any installation requirements, install multiple versions of each package, [really cool dependency "spec" syntax](https://spack.readthedocs.io/en/latest/basic_usage.html#specs-dependencies)). You can think of it as similar to nix but with a dependency resolver and no install requirements. But since I realized we're not linking any libraries and nvim is rather easy to install, I can't immediately see where spack would be useful for ethersync at this moment.

I have removed this comment, because I didn't realize CI already bootstraps nvim. If `ethersync` ever exports a library instead of just a binary, or it wants to depend on any C/C++ library, then it may become relevant.

Please feel free to ignore this comment, I like spack quite a lot but I think it's not useful right now. ^_^

daemon/src/bin/fuzzer.rs
cosmicexplorer: I ended up spending some time doing almost exactly this, but had created a subcrate instead of an integration test (https://github.com/ethersync/ethersync/compare/main...cosmicexplorer:use-cargo-test-for-integration-tests?expand=1). I was literally about to post a separate PR for it when I saw your review of this one -- I'll update that PR to make it an integration test instead!

daemon/src/bin/fuzzer.rs
cosmicexplorer: So the reason I made this comment is because I believe the term "fuzz" has a relatively specific meaning in cargo, including special support with `cargo fuzz`: https://rust-fuzz.github.io/book/cargo-fuzz.html. We have a separate subcrate for fuzzing in the `zip` crate (https://github.com/zip-rs/zip2/blob/af33ed343da5769e0587d638e435d67151990f10/fuzz/Cargo.toml#L8-L9), because it often requires different dependencies, and `cargo fuzz` links the binary against clang's ASan (which causes errors if it's in the same workspace as our `zip` library). I have contributed some changes to `zip`'s fuzzing -- I would *love* to do the same here (in a separate change).

In https://github.com/ethersync/ethersync/compare/main...cosmicexplorer:use-cargo-test-for-integration-tests?expand=1 (which I am going to move into an integration test before posting as a PR), I rename the binary from `fuzzer.rs`, because I thought it would be useful to make clear it's not using the *cargo fuzz* framework. However, I think that change is actually not necessary, since it definitely is still performing a sort of fuzzing, so I will change its name back before posting the integration test PR.

daemon/src/bin/fuzzer.rs
cosmicexplorer: Done! https://github.com/ethersync/ethersync/pull/129

daemon/src/main.rs
cosmicexplorer: I happened to have read through most of `clap`'s documentation literally two days ago for the `zip` crate (zip-rs/zip2#235), so it was a fortunate coincidence -- `clap` is very powerful, perhaps a little *too* powerful.

daemon/src/bin/fuzzer.rs
cosmicexplorer: Broken out into #130!

daemon/src/bin/fuzzer.rs
zormit, line 1: (Leaving another single comment while looking at your PRs)

It seems to me this mode/permission change should not be part of this PR?

blinry (dismissed):
This seems really good to me! :sparkles:

The last thing I'd request is undoing the mode change on fuzzer.rs spotted by @zormit :)

After that, it'd be great if you could squash the small "fixup" commits into the others, and then we can merge this! :tada:

daemon/src/bin/fuzzer.rs
cosmicexplorer: Thank you!! I'm very sorry, there is an issue with my editor configuration that happened a couple months ago that keeps causing this, I believe related to magit somehow but not sure. Haven't fixed it yet. Thanks so much, will check this in the future.

cosmicexplorer: Just set a pre-commit hook to make sure to avoid it in the future ^_^!

zormit (approved):

zormit:
Thanks for this, @cosmicexplorer! =)

PR #131, opened 2024-08-14
add optimization flags to --release profile

cosmicexplorer:
It's possible to set [optimization flags](https://doc.rust-lang.org/cargo/reference/profiles.html) in cargo for the `--release` profile. This may make it slightly faster, and symbol stripping drastically reduces binary size (14M -> 7.2M):
```shell
> git checkout main
> cargo build --release
> ls target/release/ethersync
-rwxr-xr-x 2 cosmicexplorer wheel 14M Aug 14 08:14 target/release/ethersync*
> git checkout -
> cargo build --release
> ls target/release/ethersync
-rwxr-xr-x 2 cosmicexplorer wheel 7.2M Aug 14 08:08 target/release/ethersync*
```

Note that `opt-level` is [already set to 3 by default](https://doc.rust-lang.org/cargo/reference/profiles.html#release).

### DIFF
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -44,3 +44,22 @@ rust-ini = "0.21.1"
 assert_matches = "1.5"
 serial_test = "3.1.1"
 tracing-test = "0.2.4"
+
+[profile.release]
+# Stripping symbols (not debuginfo) reduces binary size by ~1.5x.
+strip = true
+opt-level = 3
+
+# 2-3x slower to compile, but produces a ~1.2x smaller binary.
+# TODO: benchmark this!
+[profile.release-lto]
+inherits = "release"
+lto = true
+codegen-units = 1
+
+# Also quite slow to compile, but produces a ~1.6x smaller binary.
+[profile.release-size]
+inherits = "release"
+lto = true
+codegen-units = 1
+opt-level = "s"
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -45,13 +45,16 @@ assert_matches = "1.5"
 serial_test = "3.1.1"
 tracing-test = "0.2.4"
 
-[profile.release]
 # Stripping symbols (not debuginfo) reduces binary size by ~1.5x.
+# This is the default profile for `cargo install`:
+# https://doc.rust-lang.org/cargo/reference/profiles.html#profile-selection.
+# TODO: We may want to make [profile.release-size] the default
+# once we're out of the fast iteration development phase.
+[profile.release]
 strip = true
 opt-level = 3
 
 # 2-3x slower to compile, but produces a ~1.2x smaller binary.
-# TODO: benchmark this!
 [profile.release-lto]
 inherits = "release"
 lto = true


cosmicexplorer:
You're so right! I believe LTO is the main culprit here. Since `--release` is the easiest to type and we haven't benchmarked the effect of LTO yet, I decided to just add `strip = true` to `[profile.release]`, and added separate profiles `[profile.release-lto]` and `[profile.release-size]` for other optimizations. The reason for this too is that `cargo install` by default uses `--release`, so any user installing ethersync with `cargo install --ethersync` will also have to wait a long time to install it if we add LTO to the `[profile.release]` configuration. Since we haven't benchmarked the effect of LTO or other optimizations yet, it doesn't seem appropriate to prematurely optimize here. Please let me know if that makes sense! ^_^

daemon/Cargo.toml
```diff
@@ -44,3 +44,22 @@ rust-ini = "0.21.1"
...
 serial_test = "3.1.1"
 tracing-test = "0.2.4"
+
+[profile.release]
+# Stripping symbols (not debuginfo) reduces binary size by ~1.5x.
+strip = true
+opt-level = 3
+
+# 2-3x slower to compile, but produces a ~1.2x smaller binary.
+# TODO: benchmark this!
```
zormit: I'm not sure if we should leave the TODO in... What would you want to benchmark there? I think for this use-case the more handwavy "it's probably LTO" could be enough :-)
```suggestion
```

daemon/Cargo.toml
```diff
@@ -44,3 +44,22 @@ rust-ini = "0.21.1"
 assert_matches = "1.5"
 serial_test = "3.1.1"
 tracing-test = "0.2.4"
+
+[profile.release]
+# Stripping symbols (not debuginfo) reduces binary size by ~1.5x.
```
zormit:
```suggestion
# Stripping symbols (not debuginfo) reduces binary size by ~1.5x.
# This is the default profile for `cargo install`:
# https://doc.rust-lang.org/cargo/reference/profiles.html#profile-selection.
# TODO: We may want to make [profile.release-size] the default
# once we're out of the fast iteration development phase.
[profile.release]
```

zormit: I personally think for *installation* it would later be acceptable that it's building a bit slower, but also agree with you that `--release` is the easiest to type, so let's keep it like this for now :)

zormit (commented):
Thanks for splitting this up! I think with this we can merge it and then see how to make use of it the other profiles when we automate building a (small) downloadable binary.

I've suggested two comment changes, what do you think?

daemon/Cargo.toml
cosmicexplorer: Yes! I was mostly putting that in there in case someone wondered why the code was there if it's not being used! Didn't want to leave little bits of nonfunctional code around, but I like the other comment you added quite a bit and should give any reader the clues they need to see how to use it ^_^!

blinry:
Merging this :) Having the option of building a 5.6 MB binary in the release-size profile is cool, we could pick that profile when we have the CI build binaries for releases, for example!

Issue #105, opened 2024-08-08
File watcher should forward files changed by external tools to CRDT


blinry:
Examples where this would be useful:

- `cargo add <package>`
- `git checkout <file>`
- `mv a b`

Note: This would enable a "save to share" use case for unsupported editors. But it's a bit finicky – as changes by other people are not forwarded into your open editor, you need to be careful not to overwrite other people's changes.

Might be related to #103.

blinry:
Implemented by #133.

Issue #103, opened 2024-08-08
File watcher should pick up newly created files, and add them to the CRDT


blinry:
Similarly to how we already notice when files are deleted, the same should happen when files are being created.

If you copy a file into the shared directory, its content should make it to the CRDT.

blinry:
Implemented by #133.

PR #133, opened 2024-08-15
Watch for file creation and modification

blinry:
This picks up when files are added or modified, and updates the CRDT.

Manual testing seemed to work well, but there are probably some bugs left.

Addresses #103 and #105.

Remaining tasks:

- [x] Build support for `mv` on the mac
	- Make stateful?
	- Or look up whether file exists at the point of the event)
	- Pick a different watcher? inotify?
- [x] Write tests

### DIFF
--- a/daemon/src/daemon.rs
+++ b/daemon/src/daemon.rs
@@ -41,6 +41,10 @@ pub enum DocMessage {
     RemoveFile {
         file_path: String,
     },
+    CreateFile {
+        file_path: String,
+        content: String,
+    },
     Persist,
     RandomEdit,
     ReceiveSyncMessage {
@@ -64,6 +68,7 @@ impl fmt::Debug for DocMessage {
                 format!("open/close/edit/... message from editor #{id}")
             }
             DocMessage::RemoveFile { .. } => "delete file".to_string(),
+            DocMessage::CreateFile { .. } => "create file".to_string(),
             DocMessage::Persist => "persist".to_string(),
             DocMessage::RandomEdit => "random edit".to_string(),
             DocMessage::ReceiveSyncMessage { .. } => "<automerge internal sync rcv>".to_string(),
@@ -179,6 +184,13 @@ impl DocumentActor {
                     self.crdt_doc.remove_text(&file_path);
                     let _ = self.doc_changed_ping_tx.send(());
                 }
+            }
+            DocMessage::CreateFile { file_path, content } => {
+                let file_path = self
+                    .file_path_for_uri(&file_path)
+                    .expect("Could not determine file path when trying to create file");
+                self.crdt_doc.initialize_text(&content, &file_path);
+                let _ = self.doc_changed_ping_tx.send(());
             }
             DocMessage::Persist => {
                 debug!("Persisting!");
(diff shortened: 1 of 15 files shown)


daemon/src/daemon.rs
```diff
@@ -830,24 +837,111 @@ impl Daemon {
 }
 
 async fn spawn_file_watcher(base_dir: PathBuf, document_handle: DocumentActorHandle) {
-    // Beware of the file watching spaghetti triangle monster.
```
zormit, line 903: 🍝

daemon/src/daemon.rs
```diff
@@ -41,6 +41,14 @@ pub enum DocMessage {
     RemoveFile {
         file_path: String,
     },
+    CreateFile {
```
zormit: We could group removal, creation and update to a "sub enum", similar to `FromEditor`, wdyt? Or something more dynamic:
```rust
enum FileOperationKind {
	Creation,
	Removal,
	Update,
}
pub enum DocMessage {
...
	FileOperation {
		kind: FileOperationKind,
		file_path: String,
		content: Option<String>,
	}
...
}
```

daemon/src/daemon.rs
zormit: I think this is _still_ an impressive triangle that might benefit from refactoring. Which I _think_ this comment wanted to say ;-P
image

daemon/src/daemon.rs
```diff
@@ -830,24 +837,111 @@ impl Daemon {
...
+                            notify::event::RenameMode::Both,
+                        )) => {
+                            let from_path = &event.paths[0];
+                            let to_path = &event.paths[1];
+
+                            // TODO: Avoid repeating the effects here.
+                            // TODO: Is there a cleverer way to do renames in the CRDT?
+
                             document_handle
                                 .send_message(DocMessage::RemoveFile {
```
zormit: haha this diff confused me. "logically" this is the old removal event above, but that got added instead... just leaving this comment in case we come back here :)

zormit (commented):
Didn't test it yet, but reading the code it looks fine to me. I'll check out if I can break it :) Left some unimportant code-quality comments.

daemon/src/daemon.rs
```diff
@@ -830,24 +837,111 @@ impl Daemon {
...
+                                            .into(),
+                                        content,
+                                    })
+                                    .await;
+                            }
+                        }
+                        e => {
+                            // Don't handle other events.
+                            // But log them! I'm curious what they are!
+                            info!("{:?}: {e:?}", event.paths);
```
zormit: I think this is a bit too noisy for actual use. Maybe put it to `debug!`?

blinry: Or yeah, we can also just drop them at this point. The remaining once I see are Access-type events that seem to happen after each write.

daemon/src/daemon.rs
```diff
@@ -830,24 +837,111 @@ impl Daemon {
...
+                                                .expect("Failed to convert path to string")
+                                                .into(),
+                                            content,
+                                        })
+                                        .await;
+                                }
+                            }
+                        }
+                        notify::event::EventKind::Modify(notify::event::ModifyKind::Name(
+                            notify::event::RenameMode::Both,
```
zormit: For me this doesn't work. When I `mv` a file (on MacOS) I get

```
14:02:18 INFO ThreadId(14) ethersync::daemon: ["/Users/mn/code/ethersync/ethersync/playground/file2"]: Modify(Name(Any))
14:02:18 INFO ThreadId(14) ethersync::daemon: ["/Users/mn/code/ethersync/ethersync/playground/file4"]: Modify(Name(Any))
```

daemon/src/daemon.rs
zormit: Well, it's still interesting for now. I do see for example a couple of different `Modify(...)` events that were not matched above. See also my comment above.

blinry:
I started writing a nicer Watcher abstraction with tests in `watcher.rs`. Not quite done yet, you're welcome to pick this up, @zormit!

blinry:
Works for me, and is a bit less spaghetti monster! :P Over to you, @zormit.

blinry:
Oops, seems I never pushed my spaghetti disentanglement! Done now.

blinry:
Status: We're confused by why some of the tests seem to endless-loop on macOS.

The block_on could be part of the problem. Maybe replace the internal watcher with [this one](https://github.com/justinrubek/async-watcher/blob/main/examples/simple/src/main.rs) (which already comes with async support)?

blinry:
Something we noticed: When we only track creation and deletion, when the events are synced over, and "realized" in the peer, it will re-trigger the event and sync it back. (Not really bad, but unnecessary.)

Approaches:

- On a Create file event, check in the CRDT if a file of the same name with the same content (or a different content?) exists, and then do *something clever*

blinry:
The notify crate docs say that file events are not 100% reliable. Approaches:

- Do a full scan occasionally.
	- Heuristic: If a lot of file events happened in a short amount of time , do a full scan?
	- Every minute?
- Hook into `git add`?
- Implement `ethersync add`?
- Document and accept it.

zormit:
- [ ] For some reason `cargo init` does not create the `src/main.rs` file on the other side 🤔

blinry:
Note: We managed to "lose" external changes that happen quickly, by running `for i in {1..10000}; do echo i >> file; done` in both peers in bash. This is a tradeoff we're willing to accept.

- [ ] Creating ~200 new files crashes the peer, "write to file failed"

blinry:
- [ ] Log when getting new files via peer.

PR #136, opened 2024-08-16
Rename flake outputs, to better adhere to nixpkgs naming conventions

blinry:
As suggested by @mangoiv: https://github.com/ethersync/ethersync/issues/132#issuecomment-2288620383

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ Again, we have several options of how to install it:
   For testing purposes, you can run an Ethersync-enabled Neovim like this:
 
   ```bash
-  nix run github:ethersync/ethersync#neovim
+  nix run github:ethersync/ethersync#neovim-with-ethersync
   ```
 </details>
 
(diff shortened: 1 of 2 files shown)


MangoIV:
thank you so much :)

MangoIV (approved):
I think it was generally fine to keep the `neovim` the same name, though it would probably be unexpected that it would completely overwrite it - the user would apply the overlay and would basically get a neovim that could only do ethersync. 

I think I would generally avoid using overlays at all, although it may be nice to have it a convenience thing to add the neovim plugin itself and the daemon to the package set, the "try it out" - neovim I would probably not add to the overlay at all, though it doesn't hurt in its current, renamed form as you will never build it if you don't explicitly ask it to be part of your system closure. 

If you want to avoid these boilerplate-y things in the beginning of the flake, I can recommend having a look at https://flake.parts/ which I tend to use for all my flakes.

MangoIV:
If you want me to, I can open an MR to change the flake, shouldn't be much work. :)

blinry:
Closing in favor of #137.

PR #137, opened 2024-08-16
refactoring of the nix code

MangoIV:
- factored out the nix expressions for the packages into a ./nix directory
- factored out the overlay into ./nix/overlay.nix
- factored out the devShell expression into ./nix/shell.nix which can also be called with legacy nix commands (nix-shell nix/shell.nix)
- refactored flake to use flake-parts and import all the packages defined before and to reexport the overlay
- added nix outputs to the .gitignore
- added a new workflow that checks whether the nix flake evaluates on both mac and linux and whether the patched nvim builds and works 
- unbreak nix build on darwin
- unbreak devShell on darwin 
- add default.nix to use with legacy nix 
- remove dependency on naersk and instead use nixpkgs' rustPlatform


MangoIV:
I also wanted to add a `default.nix` but I was unsure on how to avoid excessive boilerplate and at the same time don't make the users of pre-flake nix mad, for now I think it suffices to be able to `callPackage` all the newly defined packages yourself.

MangoIV:
As far as I understand, ethersync doesn't currently build on nix darwin, is that known?
 https://github.com/MangoIV/ethersync/actions/runs/10418413323/job/28854519943

MangoIV:
should be more or less ready, checked that the devShell on darwin works again and the build on darwin should work now as well.

blinry:
Thanks for your work on this, @MangoIV! Seems very clean and de-composed, happy being able to benefit from your experience! We'll check it out when we get the chance.

MangoIV:
now we're back to things not working on darwin, on it. :D

MangoIV:
should be good now ... phew

MangoIV:
Btw; how do you feel about a formatter? I could add one and format everything with it. Would even simplify upstreaming if you were to, as there's now an official nixpkgs formatter: 
https://github.com/NixOS/nixfmt

MangoIV:
let me know if I can do something. to make this progress :)

.github/workflows/general.yml
```diff
@@ -55,6 +55,20 @@ jobs:
           run: cargo test --release -- --include-ignored
         - name: Run fuzzer
           run: cargo run --release --bin=fuzzer
+  check_nix:
+    name: Check nix code
+    strategy: 
+      matrix:
```
zormit: nitpick: can we drop the trailing spaces?
```suggestion
    strategy:
      matrix:
```

.github/workflows/general.yml
```diff
@@ -55,6 +55,20 @@ jobs:
           run: cargo test --release -- --include-ignored
         - name: Run fuzzer
           run: cargo run --release --bin=fuzzer
+  check_nix:
+    name: Check nix code
+    strategy: 
+      matrix: 
+        os: [ubuntu-latest, macos-latest]
```
zormit: Helpful idea to iterate on operating systems in the CI! I think we could do that for the fuzzer as well (which is independent of this PR, but just pointing it out to ourselves). Do the different systems run in parallel?

MangoIV:
@zormit we don’t want to export neovim twice, one of them was the derivation for the plugin :3

MangoIV:
Ah I see why it’s failing. I’ll fix. 🙃

.gitignore
```diff
@@ -0,0 +1,5 @@
+# nix outputs
+result*
+
+# direnv
```
zormit: `direnv` seems to be a tool of yours? Maybe we can remove it from here for now (we probably need a better/extended project `.gitignore`, but it seems to be independent of this PR).

If you want to keep it ignored for yourself, you could write it in `.git/info/exclude`

MangoIV:
@zormit sorry for all the fuzz :3

zormit (commented):
I keep forgetting that we need to trigger the CI to be executed... I'm not sure what I'm doing here (nix beginner), but I think I fixed it (or maybe not I see that you did add more commits).

Notes on this PR, so far:
- style-nitpick: We don't have contributor guidelines and probably it doesn't matter for a while, but if you don't mind, could you please fix the commit message format? The format we're trying to follow is "fix: Add ..." (instead of "[fix] add ...", so the label without square brackets, with a colon, and then capital letter.


I'll post a half-baked review as I might stop reviewing this for now (to keep it somewhat timeboxed, it doesn't have a very high priority, this refactoring)

.github/workflows/general.yml
MangoIV: yes, they do, it generates a matrix that will run on different runners in parallel :)

zormit:
> let me know if I can do something. to make this progress :)

So from my perspective (and as indicated above), we just need some patience from your side as it's not a "core thing", but a "nice to have". I spent a bit of time looking at it above, unfortunately didn't get through.

But any time, please ping us if we forget to click "run CI"... It should be a at least green build when we get to review this :D

.gitignore
MangoIV: oh, I'm kinda used to `direnv` being a kinda ubiquitous tool to use with nix, e.g. I personally have `use flake . -Lv` in the `.envrc` file, with `nix-direnv`, this makes loading the devShell instant, it really makes working with nix much nicer :) 

but yes, I can remove it for the matter of this PR :)

MangoIV:
>  it's not a "core thing", but a "nice to have" 

sure, no worries, if you said "nothing, no time for this", this is also a fine answer for me :D

.github/workflows/general.yml
zormit: Nice! We'll follow-up on this in #153.

.gitignore
zormit: Okay interesting point. @blinry might have a different opinion than me, my exposure to nix tooling is "mainly this project".

MangoIV:
@zormit I also just noticed, if you want a static binary for ethersync, for now it looks like it's as easy as 
`pkgsStatic.callPackage ./nix/ethersync.nix {};`
it's fat as hell (23.5 Mb) but you can maybe reduce this with `upx` and then you can distribute that :)

MangoIV:
I have also reworded all the commits :3

blinry:
Some quick thoughts from me:

- I'd like enforcing a formatter :) I'm used to Alejandra, but I'm fine with nixfmt, too!
- It might make sense to move the `nix` directory to a subfolder like `contrib`, as the `nix` directory feels more "supplemental", compared to the very integral components `daemon` and `vim-plugin`. Neovim [used to keep the files (even the flake) in `contrib`](https://github.com/neovim/neovim/pull/28863). In the long term, we could maybe upstream it to nixpkgs :)
- It's cool that the MacOS build works now! \o/

MangoIV:
@blinry are you happy to merge without putting everything in `contrib`?

MangoIV:
I've also exported the static nix build - you could e.g. upload it when merging to `main` and people could just download a binary from the project website instead of having to actually build the project ;)

MangoIV:
it's a bit thick, but the size can be reduced with `upx`...

MangoIV:
is the `android` derivation interesting to you? I don't have android so I can't test it and it's also trivial to just build it yourself, but if you wish I can add it as an outpu...

blinry:
The static build idea is indeed something we might consider, to speed up installation! This is also mentioned in the [Cargo documentation](https://rust-cli.github.io/book/tutorial/packaging.html#distributing-binaries). I guess we'd need an action that uploads the artifact to a reasonable place? At least for releases.

I've moved `nix/` into `contrib/` for now, and split out the test into a separate CI workflow, so that it runs when the Nix code changes!

Merging this for now – feel free to send more PRs later if you see opportunities for refinement!

Thanks so much for working on this! :green_heart:

MangoIV:
@blinry see for example https://github.com/MangoIV/cabal-audit/blob/b1897d26b69fd54790ed68a01657a10e76c9209e/.github/workflows/cabal-audit.yml#L49 (it’s not perfect but it works well enough)

PR #140, opened 2024-08-16
Add a changelog, and back-date some releases

blinry:
Are we happy with this changelog, @zormit?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,41 @@
+# Unreleased features
+
+(None yet)
+
+# 0.3.0 (2024-08-13)
+
+New features:
+
+- Integrated libp2p for transport encryption and authentication of peers.
+- Added a password protection – peers must specify the same `--secret` to connect.
+- You can now provide default startup arguments in `.ethersync/config`.
+- When editors send
+
+Bug fixes:
+
+- The daemon is now more resilient against misbehaving peers and editors, and won't crash as easily.
+
+# 0.2.2 (2024-08-01)
+
+New features:
+
+- Persist the CRDT to `.ethersync/doc`, so that peers can run independently, and sync back later.
+- When restarting the daemon, calculate file diffs since the daemon was last online, and apply them to the CRDT.
+- Dynamically create files when they are opened.
+- When files are deleted, delete them for all peers.
+- Sandbox file I/O to be restricted into the shared project folder.
+
+# 0.2.1 (2024-07-26)
+
+New features:
+
+- Share multiple files per directory.
+- Transmit cursor positions, and display them in the Neovim plugin.
+
+# 0.2.0 (2024-05-02)
+
+Rewrite of the initial prototype in Rust.
+
+# 0.1.0 (2024-01-22)
+
+An initial prototype in Typescript. It only supports sharing a single file.


CHANGELOG.md
```diff
@@ -0,0 +1,41 @@
...
+- Dynamically create files when they are opened.
+- When files are deleted, delete them for all peers.
+- Sandbox file I/O to be restricted into the shared project folder.
+
+# 0.2.1 (2024-07-26)
+
+New features:
+
+- Share multiple files per directory.
+- Transmit cursor positions, and display them in the Neovim plugin.
```
zormit, line 41: Even though not a core feature, we could mention adding a nix flake here?

zormit: Also it seems we have done some improvements to the daemon-editor protocol specification (it might be good practice to actively mention protocol tweaks and changes in the dev log until we have an actual versioning of that protocol)

See 23246a3

CHANGELOG.md
```diff
@@ -0,0 +1,41 @@
...
+
+# 0.2.2 (2024-08-01)
+
+New features:
+
+- Persist the CRDT to `.ethersync/doc`, so that peers can run independently, and sync back later.
+- When restarting the daemon, calculate file diffs since the daemon was last online, and apply them to the CRDT.
+- Dynamically create files when they are opened.
+- When files are deleted, delete them for all peers.
+- Sandbox file I/O to be restricted into the shared project folder.
```
zormit, line 33: The editor plugin can now send JSON-RPC requests and get feedback on whether the intended operation worked or not (#74). Or less "internal": nvim plugin can now process feedback on whether opening a file was successful (#87)

btw, we could also add PRs to this changelog? 🤔 probably overkill to backdate, but going forward... there's probably an app for that? :D (i.e. a tool to scrape our git(hub) log and generate a summary)

CHANGELOG.md
```diff
@@ -0,0 +1,41 @@
...
+(None yet)
+
+# 0.3.0 (2024-08-13)
+
+New features:
+
+- Integrated libp2p for transport encryption and authentication of peers.
+- Added a password protection – peers must specify the same `--secret` to connect.
+- You can now provide default startup arguments in `.ethersync/config`.
+- When editors send
```
zormit: when editors send! :D love letters? 💌 

do you remember what you wanted to write here? could it be in a similar direction to what I added in 0.2.2 regarding editor `open`ing? we have certainly improved this between 0.2.2 and 0.3.0 as well, so we can also just place it here.

CHANGELOG.md
```diff
@@ -0,0 +1,41 @@
+# Unreleased features
```
zormit: What's the plan with this section? Is it a running list of what's on `main`?

zormit: - we now have an mdbook! =)
- #127

zormit (commented):
I checked it and it looks good to me. I've left some ideas for more bullets.

Some potential follow up:
- Should we make sure that a release exists for each of those tags? It seems that Github allow turning [tags](https://github.com/ethersync/ethersync/tags) into a [release](https://github.com/ethersync/ethersync/releases).

CHANGELOG.md
blinry: > Should we make sure that a release exists for each of those tags?

I think probably not – because we didn't update the version in the Cargo.toml, it will show the wrong version. Except if we start little release branches, and hotfix that there…

I'd be fine with updating the version from now on, when we do the next release! Which could be soon! :)

blinry: > protocol tweaks and changes

Oh yeah! We dropped the revisioning for the "cursor" messages from the protocol at some point! That's one of the changes I can think of, in addition to supporting JSON-RPC requests now.

CHANGELOG.md
blinry: > there's probably an app for that

Yeah, like https://github.com/MarcoIeni/release-plz :D (Which follows https://www.conventionalcommits.org/en/v1.0.0/)

CHANGELOG.md
blinry: I don't remember :cry: #74 seems to have been part of 0.2.2, so I'm not sure what else to add here.

CHANGELOG.md
blinry: > Is it a running list of what's on main?

Yeah, that's what I had in mind there! But I think it's also fine to just add a new section once we make a release!

blinry:
Tried adding all your suggestions to the CHANGELOG! :)

PR #141, opened 2024-08-16
Support multiple local editors

### LINKED ISSUE #63: Fix handling multiple editor connection
Either by allowing or disallow (right now it's kind of undefined behavior). Allowing would be nicer, but we might back off? :)

blinry:
Fixes #63.

This needs a surprising amount of changes in our different daemon functions.

- [x] Update the tests.
- [x] Look into the "Failed to write to socket" error – can/should we avoid that? But I think I even got this error sometimes before this PR.

### DIFF
--- a/daemon/src/daemon.rs
+++ b/daemon/src/daemon.rs
@@ -575,7 +575,7 @@ impl DocumentActor {
         } in cursor_states
         {
             let message = EditorProtocolObject::Request(EditorProtocolMessageToEditor::Cursor {
-                cursor_id,
+                userid: cursor_id,
                 name,
                 uri: format!("file://{}", self.absolute_path_for_file_path(&file_path)),
                 ranges,
--- a/daemon/src/daemon.rs
+++ b/daemon/src/daemon.rs
@@ -141,8 +141,8 @@ impl DocumentActor {
     /// If any editor has an `ot_server` for a given file,
     /// it means that the daemon doesn't have ownership.
     fn owns(&mut self, file_path: &str) -> bool {
-        for (_, ot_servers) in self.ot_servers.iter_mut() {
-            if ot_servers.get_mut(file_path).is_some() {
+        for (_, ot_servers) in self.ot_servers.iter() {
+            if ot_servers.get(file_path).is_some() {
                 return false;
             }
         }
--- a/daemon/src/daemon.rs
+++ b/daemon/src/daemon.rs
@@ -241,7 +241,7 @@ impl DocumentActor {
             DocMessage::CloseEditorConnection(editor_id) => {
                 self.editor_clients.remove(&editor_id);
 
-                let userid = self.crdt_doc.actor_id() + "-" + editor_id.to_string().as_str();
+                let userid = self.cursor_id(editor_id);
                 debug!("Deleting cursor {userid}");
                 self.maybe_delete_cursor_position(&userid).await;
             }
@@ -385,7 +385,7 @@ impl DocumentActor {
                 let file_path = self
                     .file_path_for_uri(&uri)
                     .map_err(anyhow_err_to_protocol_err)?;
-                let userid = self.crdt_doc.actor_id() + "-" + editor_id.to_string().as_str();
+                let userid = self.cursor_id(editor_id);
                 self.store_cursor_position(&userid, file_path.clone(), ranges.clone());
 
                 let cursor_state = CursorState {
@@ -400,6 +400,10 @@ impl DocumentActor {
             }
         }
         Ok(())
+    }
+
+    fn cursor_id(&self, editor_id: EditorId) -> String {
+        self.crdt_doc.actor_id() + "-" + editor_id.to_string().as_str()
     }
 
     async fn handle_message_from_editor(&mut self, editor_id: EditorId, message: String) {
--- a/daemon/src/daemon.rs
+++ b/daemon/src/daemon.rs
@@ -141,11 +141,11 @@ impl DocumentActor {
 
     fn owns(&mut self, file_path: &str) -> bool {
         for (_, ot_servers) in self.ot_servers.iter_mut() {
-            if let Some(_) = ot_servers.get_mut(file_path) {
-                return true;
+            if ot_servers.get_mut(file_path).is_some() {
+                return false;
             }
         }
-        return false;
+        true
     }
 
     async fn handle_message(&mut self, message: DocMessage) {
@@ -649,7 +649,7 @@ impl DocumentActor {
 
     fn maybe_write_file(&mut self, file_path: &str) {
         // Only write to the file if editor *doesn't* have the file open.
-        if !self.owns(file_path) {
+        if self.owns(file_path) {
             if let Ok(text) = self.current_file_content(file_path) {
                 let abs_path = self.absolute_path_for_file_path(file_path);
                 debug!("Writing to {abs_path}.");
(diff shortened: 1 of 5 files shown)


blinry:
#142 avoids the "Failed to write to socket" problem by simplifying things a lot. Sending to the EditorHandle can still fail, but I think that can't be avoided, if we try to write in a bad moment, while the socket connection is stopped. But the caller will know from the `Result`.

daemon/src/daemon.rs
```diff
@@ -5,8 +5,8 @@ use crate::peer;
 use crate::sandbox;
 use crate::types::{
     CursorState, EditorProtocolMessageError, EditorProtocolMessageFromEditor,
-    EditorProtocolMessageToEditor, EditorProtocolObject, EditorTextDelta, FileTextDelta,
-    JSONRPCFromEditor, JSONRPCResponse, PatchEffect, Range, RevisionedEditorTextDelta, TextDelta,
+    EditorProtocolMessageToEditor, EditorProtocolObject, FileTextDelta, JSONRPCFromEditor,
+    JSONRPCResponse, PatchEffect, Range, RevisionedEditorTextDelta, TextDelta,
```
zormit, line 9: So many types, makes diff hard to read 😆 
Ok, basically we remove `EditorTextDelta`. Got it. :)

daemon/src/daemon.rs
```diff
@@ -509,76 +531,95 @@ impl DocumentActor {
...
-                debug!("Applying incoming CRDT patch for {file_path}");
-                let rev_text_delta_for_editor = ot_server.apply_crdt_change(delta);
-                self.send_to_editors(rev_text_delta_for_editor, &file_path)
-                    .await;
+    async fn process_crdt_file_deltas_in_ot_servers(
+        &mut self,
+        exclude_id: Option<EditorId>,
+        file_deltas: Vec<FileTextDelta>,
+    ) {
+        let editor_ids: Vec<EditorId> = self.ot_servers.keys().cloned().collect();
```
zormit, line 540: Ugh, that's unfortunate that we can't just use an iterator due to ownership. I think there's an opportunity for refactoring here. I'm not there yet, but I'm thinking of splitting off the part that handles sending to the editor (everything around `editor_clients`). Will probably have to be some follow-up work though as I want to wrap this review up and merge :-D

https://users.rust-lang.org/t/iterating-over-a-self-some-vec-while-mutating-self/51135

daemon/src/daemon.rs
zormit: But, after looking into the whole of it: Nice that we can get rid of it as a dependency to daemon! =) I like the more simplified interface and that we just do the conversion in `document.rs`.

daemon/src/daemon.rs
```diff
@@ -509,76 +531,93 @@ impl DocumentActor {
...
-    async fn send_to_editor_clients(&mut self, message: EditorProtocolMessageToEditor) {
-        let mut to_remove = Vec::new();
-        for (id, handle) in &mut self.editor_clients.iter_mut() {
-            if handle
-                .send(EditorProtocolObject::Request(message.clone()))
-                .await
-                .is_err()
-            {
+                warn!(
+                    "Sending to editor client failed: We don't have a client registered with id #{editor_id:?}"
```
zormit: I think this is a confusing warning. Maybe a reason why I would hold back with approval until I understand better what your idea was behind this part of the change.

Before the warning was in the `else` of the `Some(handle)`, now it became part of the `is_err`.

daemon/src/daemon.rs
```diff
@@ -509,76 +531,93 @@ impl DocumentActor {
...
+                );
                 // Remove this client.
-                to_remove.push(*id);
+                self.editor_clients.remove(editor_id);
```
zormit, line 619: It seems we can directly remove it now, because we're not _iterating_ over `editor_clients` at this point?

otoh this mutation might be one of the pieces that blocks a proper "iter_mut" [further above](https://github.com/ethersync/ethersync/pull/141#discussion_r1727201789)?

zormit (commented):
Gosh this got somewhat complex now 😅 But I can confirm it working with some basic manual testing, so that's great! :)

I'm wondering if we can separate concerns better to make it more readable. I'm a bit going back and forth whether we should merge it like this or work on making it nicer before that.

I'm slightly for merging to prevent merge conflicts with the other open branch, and then do a bigger refactoring once we're down on the number of PRs.

Only the `warn!` needs to be fixed. I would do it, but got to go now :) Either you can pick it up or I'll look at it at some point later.

zormit:
Also the fuzzer seems to not come to an end. Did one of my changes cause this? 🤔

blinry:
I tried bisecting:

```
~/wip/ethersync git bisect good 3671c5f
status: waiting for bad commit, 1 good commit known
~/wip/ethersync git bisect bad 3822631
Bisecting: 2 revisions left to test after this (roughly 1 step)
[22d1582d3ab0d9e60f596f044e2ed31e640e920e] refactor: Drop unnecessary _mut
~/wip/ethersync git bisect good
Bisecting: 0 revisions left to test after this (roughly 1 step)
[e6ccb499e659c8239c0413a98a528082efa5409a] refactor: Rename userid to cursor_id in all places
~/wip/ethersync git bisect bad
Bisecting: 0 revisions left to test after this (roughly 0 steps)
[e3ad06628e44f3e396452ed7ad2c5823f1e72432] refactor: Pull out little helper to deduplicate id building logic
~/wip/ethersync git bisect good
e6ccb499e659c8239c0413a98a528082efa5409a is the first bad commit
commit e6ccb499e659c8239c0413a98a528082efa5409a
Author: Moritz Neeb <nt4u@kpvn.de>
Date:   Thu Aug 22 16:14:44 2024 +0200

    refactor: Rename userid to cursor_id in all places
```

That commit seems to be the one that introduces endless hangs during the "Performing edits" step...

blinry:
Ah! The commit is also renaming the "userid" field of the "cursor" message in the editor protocol! Should we re-think our terminology there?

Instead of "cursor_id" in the code, I feel like "editor_id" could be a good name – as the active editor is the determining factor for the ID.

In the protocol, I still like "userid" right now. A second attached editor on a system kind of behaves like another "user".

blinry:
Ah, editor_id is already the integer ID of the editor connection, right… yeah, let's go with cursor_id then.

daemon/src/daemon.rs
blinry: Tried addressing that in ed09de7, does that make more sense? The warn! message was indeed not fitting. I moved the original one to the place it belongs, and made this one better (and lowered it to an info!, because it can happen during regular editor disconnect).

blinry:
I'd also favor merging this now, and refactoring it later.

It would be really nice to have a single component that has the EditorHandle and its associated OTServers built into it.

I'm also not super happy with the need to have the `exclude_id: Option<EditorId>` in multiple places. I think at its core, there are multiple components that are plugged together here – the editor connections, and the CRDT document part. And each time we get a message (FileTextDelta or CursorState) from one of them, we want to forward it to all the others. We can probably write that more elegantly.

zormit:
Okay last thing before we merge (potential blocker): Is it sufficiently clear already that a 2nd (and following) editor might have an out of date buffer? I feel like it's easy to shoot yourself in the foot with this feature as it is.

Terminal 1
- `vim file` (reads content from disk)
- edit file (can even be an external edit by peer, user can't reliably control that)

Terminal 2
- `vim file` (reads *outdated* content from disk)
- edit...? 🤔 => daemon can crash! (see also #29)

```log
10:18:12  INFO ThreadId(03) ethersync::editor: Client #0 connected
10:18:19  INFO ThreadId(11) ethersync::editor: Client #1 connected
thread 'tokio-runtime-worker' panicked at /Users/mn/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ropey-1.6.1/src/rope.rs:826:13:
Attempt to index past end of Rope: line index 4, Rope line length 2
```

Potential options:
- accept this risk and complexity and just merge it. File more bugs if we shoot ourselves in the foot too much.
- continue to work on this (e.g. adapt the way we do the `open`. Allow to double check. "I am opening with this content, can I participate in your sync loop?")
- reject this feature (which I think is a valid option too. It feels a bit like feature creep 😅 also I'm not super happy with the complexity we get to support this, as discussed above. Do we have to support this? what's the motivation again?)

blinry:
Oh. That's a really good point. Didn't think about that.

My main motivation is to make it easier for plugin developers to try their own plugin (by just opening two editors on a file).

It sounds a bit like when the second editor opens the file, the *daemon should tell it*: "Here, this is the current real buffer content!"

I think I'd favor merging + filing new bugs :) If we don't tell anyone that this is supposed to work, nothing would change for them. But maybe we can put the feature in the docs, together with the current restrictions.

zormit:
Let's gooo then! =) Move vast and rake things. Or something like that. 😈

PR #142, opened 2024-08-16
Simplify the EditorHandle

blinry:
Removing one or two layers of abstraction. While possibly using another: Tokio's `Framed` and `Codec` traits! :green_heart:

```
3 files changed, 45 insertions(+), 163 deletions(-)
```


zormit (approved):
I don't understand how this works yet, will need to read up on the concepts, but it seems to work and I like that we can remove those layers of abstractions from our codebase.

What do you think about moving all of the `connect.rs` back into `editor.rs` now that it's so reduced anyways and connect doesn't do the peer stuff anymore?

PR #143, opened 2024-08-18
Set up mdBook skeleton for documentation

blinry:
This makes the built book available at https://ethersync.github.io/ethersync/.

Before we merge:

- [x] Remove workflow trigger for PRs (only the one for main should stay)
- [x] Very visibly link to the book from the README/remove stale link to dev guide
- [x] Remove most content from the README?

I'd like it if:

- [ ] We could keep the sidebar open for narrower screens than currently (doesn't show by default for narrow windows).
- [ ] The deployed URLs didn't contain "/foo.html", but "/foo/". Is there an option for that?


.github/workflows/mdbook.yml
```diff
@@ -0,0 +1,58 @@
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Install latest mdbook
+        run: |
+          tag=$(curl 'https://api.github.com/repos/rust-lang/mdbook/releases/latest' | jq -r '.tag_name')
+          url="https://github.com/rust-lang/mdbook/releases/download/${tag}/mdbook-${tag}-x86_64-unknown-linux-gnu.tar.gz"
+          mkdir mdbook
+          curl -sSL $url | tar -xz --directory=./mdbook
+          echo `pwd`/mdbook >> $GITHUB_PATH
```
zormit, line 32: do we always need latest? otherwise a bit simpler way (according to [this](https://github.com/actions/starter-workflows/blob/main/pages/mdbook.yml)) could be:
```suggestion
      - name: Install mdBook
        run: |
          curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf -y | sh
          rustup update
          cargo install --version ${MDBOOK_VERSION} mdbook
```

where we have previously defined a version
```yml
    env:
      MDBOOK_VERSION: 0.4.36
```

book/src/introduction.md
```diff
@@ -0,0 +1,3 @@
+# Why Ethersync?
+
+Why not?
```
zormit: 😎

.github/workflows/mdbook.yml
```diff
@@ -0,0 +1,58 @@
...
+      - [$default-branch]
+    paths:
+      - 'book/**'
+      - '.github/workflows/mdbook.yml'
+  pull_request:
+    paths:
+      - 'book/**'
+      - '.github/workflows/mdbook.yml'
+
+# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
```
zormit, line 9: I have added one commit, lmk if you like this :)

zormit (commented):
Regarding some of your open questions/todos:

> Remove most content from the README?

Even though it's maybe some overhead, I would like to have the basic quickstart in the README. I think it's nice to come to a repository and be able to get started without having to dive into the docs. On the other hand it's maybe some overhead to maintain both places.

> We could keep the sidebar open for narrower screens than currently (doesn't show by default for narrow windows).

It seems to me it's determined on reload, so probably a JS option rather than plain CSS responsiveness? Do we need to create our own theme?

> The deployed URLs didn't contain "/foo.html", but "/foo/". Is there an option for that?

I _think_ that might be independent of rust book, but dependent on github pages? I have tried to figure out how e.g. https://yarr.fyi/setup has done it, but that's a whole different [way of deployment](https://git.sr.ht/~ntietz/yet-another-rust-resource/tree/main/item/Makefile): On cloudflare not gh-pages using npx wrangler.

zormit:
Regarding removing `.html`: Ah, maybe it *is* an mdbook option, because:
https://stackoverflow.com/questions/21244910/remove-html-extension-from-github-pages

So it seems we already can use https://ethersync.github.io/ethersync/editor-plugin-dev-guide, but the table of content links it as https://ethersync.github.io/ethersync/editor-plugin-dev-guide.html.

It seems there's an open issue for this: https://github.com/rust-lang/mdBook/issues/1563

book/.gitignore
zormit, line 1: I'm wondering regarding the path of this in the repo... will we still have a use for `docs`? I was thinking that the ADRs could also go into this book format, in the Development/Contributing chapter? So maybe we can declare `docs` to be a book?

blinry: I also thought so in the beginning… but I think we *might* still have docs that are so internal that they don't really fit in the book. I'd say let's keep that distinction for now. If docs really becomes empty, we can consider renaming back :)

I saw another project doing just that, but I don't remember which one…

PR #144, opened 2024-08-20
ADR 08: How to secure connections?

blinry:
This would probably do as-is; but I think it could also still be expanded a bit. Let me know if you see "blind spots".

Ah, I didn't integrate your comment about the key durations yet, for example!


zormit:
@blinry thanks for preparing this! =)
I added my thoughts. Merging, but feel free to iterate on it later.

PR #148, opened 2024-08-22
Allow manual Action for deploying book

zormit:


blinry:
In which situation would this be useful? To build + publish it quickly from another branch?

zormit:
tbh, the _reason_ I added this was because I was a bit desperate, because merging to main did not trigger the workflow.
Probably this `[$default-branch]` thingy didn't properly work, but I didn't suspect that at this point. See also #149, where I fixed that (and the indentation I got wrong here).
image
(you see: No ✅  until #149 was merged.)

But I am also wondering what's a good way to _check_ a PR (to the book) before merging it to main... Is there a way to build temporary drafts? We probably would need to split the build and deploy as I also tried earlier somewhere, and then *build* everytime, but only deploy when merging to main.

This manual action could be a somewhat hacky workaround: Before merging a PR we could manually deploy the branches book and see if it looks okay. If not, rollback to main. If it looks good, go ahead and merge.

PR #152, opened 2024-08-26
Fix initialization logic of Document

### LINKED ISSUE #150: when no file is in the peering daemon's directory, the daemon panicks
When running the daemon in `--peer` mode and there is no file in the ethersync enabled directory, the daemon crashes with 

```
13:55:12  INFO ThreadId(01) ethersync::peer: No config file found, please provide everything through CLI options
13:55:12  INFO ThreadId(01) ethersync: Starting Ethersync on /home/mangoiv/playground
13:55:12  INFO ThreadId(01) ethersync::daemon: Loading persisted CRDT document from "/home/mangoiv/playground/.ethersync/doc"
13:55:12  INFO ThreadId(01) ethersync::daemon: Done.
thread 'main' panicked at src/document.rs:204:14:
Failed to get files store from document: Automerge document doesn't have a files Map object

Stack backtrace:
   0: anyhow::error::<impl anyhow::Error>::msg
   1: anyhow::__private::format_err
   2: ethersync::document::Document::top_level_map_obj
   3: ethersync::document::Document::files
   4: ethersync::daemon::DocumentActorHandle::new
   5: ethersync::daemon::Daemon::new
   6: tokio::runtime::park::CachedParkThread::block_on
   7: tokio::runtime::runtime::Runtime::block_on
   8: ethersync::main
   9: std::sys_common::backtrace::__rust_begin_short_backtrace
  10: std::rt::lang_start::{{closure}}
  11: std::panicking::try
  12: std::rt::lang_start_internal
  13: main
  14: __libc_start_call_main
  15: __libc_start_main@@GLIBC_2.34
  16: _start
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: ethersync::document::Document::files
   4: ethersync::daemon::DocumentActorHandle::new
   5: ethersync::daemon::Daemon::new
   6: tokio::runtime::park::CachedParkThread::block_on
   7: tokio::runtime::runtime::Runtime::block_on
   8: ethersync::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Sometimes I cannot reproduce it, then it helps to restart the remote daemon.

zormit:
fixes #150

### DIFF
--- a/daemon/src/daemon.rs
+++ b/daemon/src/daemon.rs
@@ -114,9 +114,14 @@ impl DocumentActor {
                 .unwrap_or_else(|_| panic!("Could not read file '{persistence_file:?}'"));
             Document::load(&bytes)
         } else {
-            info!("Initializing a new CRDT document");
-            let init = !persistence_file_exists && is_host;
-            Document::new(init)
+            // Here we either do not have a persistence_file_exists
+            // or we want to force recreating it with the --init flag.
+            //
+            // If we are the host, we are responsible for setting up
+            // the top level structure of the document, a peer *must not* do that.
+            let role = if is_host { "host" } else { "peer" };
+            info!("Initializing a new CRDT document (as {role})",);
+            Document::new(is_host)
         };
         info!("Done.");
 
--- a/daemon/src/document.rs
+++ b/daemon/src/document.rs
@@ -199,10 +199,11 @@ impl Document {
     }
 
     pub fn files(&self) -> Vec<String> {
-        let file_map = self
-            .top_level_map_obj("files")
-            .expect("Failed to get files store from document");
-        self.doc.keys(file_map).collect()
+        if let Ok(file_map) = self.top_level_map_obj("files") {
+            self.doc.keys(file_map).collect()
+        } else {
+            vec![]
+        }
     }
 
     pub fn file_exists(&self, file_path: &str) -> bool {
--- a/daemon/src/document.rs
+++ b/daemon/src/document.rs
@@ -24,12 +24,12 @@ pub struct Document {
 }
 
 impl Document {
-    pub fn new(init: bool) -> Self {
+    pub fn new(init_structure: bool) -> Self {
         let mut s = Self {
             doc: AutoCommit::default(),
         };
 
-        if init {
+        if init_structure {
             s.initialize_top_level_maps();
         }
 


daemon/src/daemon.rs
```diff
@@ -114,9 +114,14 @@ impl DocumentActor {
                 .unwrap_or_else(|_| panic!("Could not read file '{persistence_file:?}'"));
             Document::load(&bytes)
         } else {
-            info!("Initializing a new CRDT document");
-            let init = !persistence_file_exists && is_host;
```
zormit, line 118: What if we are the host and the persistence file exists? I think in this else-branch we still want to recreate it, now, because of `--init`.

zormit:
Actually I just realized that I fixed something else, but related. A fix for the actual issue still has to be done, so I'm turning this PR into a draft 🙃

daemon/src/document.rs
```diff
@@ -24,12 +24,12 @@ pub struct Document {
 }
 
 impl Document {
-    pub fn new(init: bool) -> Self {
+    pub fn new(init_structure: bool) -> Self {
```
zormit, line 27: `init` is overloaded, so I'm trying to specify what exactly.

reasoning: I thought it's misleading to call this parameter `init`, as the flag.
Initializing a new document (this whole constructor) != initializing the automerge objects. The latter is an optional sub operation of the former.

daemon/src/document.rs
```diff
@@ -199,10 +199,11 @@ impl Document {
...
 
     pub fn files(&self) -> Vec<String> {
-        let file_map = self
-            .top_level_map_obj("files")
-            .expect("Failed to get files store from document");
-        self.doc.keys(file_map).collect()
+        if let Ok(file_map) = self.top_level_map_obj("files") {
+            self.doc.keys(file_map).collect()
+        } else {
+            vec![]
```
zormit, line 205: I think it's fair and somewhat elegant to respond with the empty vector for a getter? That makes it possible to iterate over `files` here independent of whether a file exists or not:

https://github.com/ethersync/ethersync/blob/64cf875ca82f481e380237bd00e3e5954e668683/daemon/src/daemon.rs#L674

Also it's abstracting a bit better over the internals, I'd say... Nothing outside of `Document` should have to care/worry about the automerge structure.

blinry:
Note: Once (and if) we expand our "peer protocol" to allow sending cursor messages there (instead of storing them in the "states" dict), all of this will become much nicer! It will again become possible that both peers independently start adding files, and then, later, sync them together!

daemon/src/document.rs
blinry: Indeed!

daemon/src/document.rs
blinry: I like it!

daemon/src/daemon.rs
blinry: True!

blinry (approved):
Looks good to me, thanks for the fix! :)

PR #155, opened 2024-08-27
ADR 09: How to keep directory structures in sync?

blinry:
What I wrote on the train. Stopped writing it half-way, needs proof-reading and probably more content.

### DIFF
--- a/docs/decisions/09-multi-file.md
+++ b/docs/decisions/09-multi-file.md
@@ -1,6 +1,6 @@
 ---
-status: draft
-date: 2024-08-22
+status: accepted
+date: 2024-09-13
 ---
 # How to keep directory structures in sync?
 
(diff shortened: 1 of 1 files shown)


docs/decisions/09-multi-file.md
```diff
@@ -0,0 +1,81 @@
...
+status: draft
+date: 2024-08-22
+---
+# How to keep directory structures in sync?
+
+## Context and Problem Statement
+
+Instead of only syncing a static, single file, we want to be able to collaborate on entire directory structures.
+
+What should our core data structure look like, and how should we approach edge cases?
```
zormit, line 11: A snippet from my notes, just dropping it here, so I can remove the file in favor of this :)

```
There are multiple edge cases when allowing file operations and we need to decide how we want to deal with them.
-  When a peer deletes a file on the filesystem, it gets removed from the CRDT. When someone still has the buffer open, it should be handled gracefully.
```

zormit: This is in my view covered by "Case 2"

PR #156, opened 2024-08-27
Run tests on MacOS in addition to Ubuntu

zormit:
This PR ensures that the tests run for both of our supported operating systems.

### DIFF
--- a/.github/workflows/general.yml
+++ b/.github/workflows/general.yml
@@ -8,13 +8,13 @@ on:
       - 'daemon/**'
       - 'vim-plugin/**'
       - 'flake.*'
-      - '.github/workflows/integration.yml'
+      - '.github/workflows/general.yml'
   pull_request:
     paths:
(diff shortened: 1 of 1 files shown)


.github/workflows/general.yml
```diff
@@ -23,38 +23,40 @@ env:
...
+          workspaces: "daemon"
+      - name: Build Ethersync
+        run: cargo build --release
+      - name: Add ethersync binary to PATH
+        run: echo "$PWD/target/release" >> $GITHUB_PATH
+      - run: ethersync --version
+      - uses: rhysd/action-setup-vim@v1
+        with:
+          neovim: true
+          version: v0.10.1
```
zormit, line 48: besides spacing this is the main change of the steps (see individual commits)

blinry:
Pretty neat! I like that the tests can happen in parallel, so the CI feedback loop doesn't get (much) longer!

PR #157, opened 2024-08-28
Refactor the multi-editor feature

blinry:
I felt responsible for cleaning up the rather ugly code produced by the multi-editor feature.

This is a rather unwieldy diff…

- Refactor out a `EditorConnection` type, which knows how to write messages to the editor, and holds the OT servers (combines what previously were the `ot_servers` and `editor_clients` fields of the daemon, now a single `editor_connections` field).
- Pull out the ignore code to `ignore.rs`.
- Introduce a `InsideMessage` type (better name pending?), that represents unversioned actions on relative file_paths. It's what the editors need to pass to each other, and to the CRDT document. The logic is: If any editor, or the CRDT document produces an `InsideMessage`, pass it to all the other components. The exception is a random edit – pass that to *all* components.

Fuzzer and tests run locally, let's see if the CI is happy about this? I haven't checked my work very carefully, and could have introduced subtle bugs.

Two unit tests would need to be rewritten, I have commented them out for now. And need to get some sleep. :D


daemon/src/daemon.rs
```diff
@@ -537,95 +410,16 @@ impl DocumentActor {
...
-            warn!("Sending to editor client failed: We don't have a client registered with id #{editor_id:?}");
-        }
+        let handle = self
+            .editor_handles
+            .get_mut(editor_id)
+            .expect("Could not get editor handle");
+
+        handle.send(message).await.unwrap_or_else(|err| {
+            error!("Failed to send message to editor: {err} Removing editor.");
+            self.editor_connections.remove(editor_id);
```
zormit, line 475: I think we should also remove from `editor_handles` here.

daemon/src/daemon.rs
```diff
@@ -85,9 +83,8 @@ type DocChangedReceiver = broadcast::Receiver<()>;
 pub struct DocumentActor {
     doc_message_rx: mpsc::Receiver<DocMessage>,
     doc_changed_ping_tx: DocChangedSender,
-    editor_clients: HashMap<EditorId, EditorHandle>,
-    /// There's one OTServer per buffer. Same file in a different editor is a different buffer.
-    ot_servers: HashMap<EditorId, HashMap<String, OTServer>>,
+    editor_handles: HashMap<EditorId, EditorHandle>,
+    editor_connections: HashMap<EditorId, EditorConnection>,
```
zormit: I'm wondering if we can summarize these two as one kind of entity inside here, such that we're forced to keep then in sync?

daemon/src/daemon.rs
```diff
@@ -729,45 +504,83 @@ impl DocumentActor {
...
         // Send cursor delete to local peers.
-        if let Some(file_path) = self.crdt_doc.files().first() {
-            let cursor_states = vec![CursorState {
+        self.broadcast_to_editors(
+            None,
+            &InsideMessage::Cursor {
                 cursor_id: cursor_id.to_string(),
                 name: None,
-                file_path: file_path.to_string(),
+                file_path: "".to_string(),
```
zormit, line 664: We can get away with `""` in the deletion case? :)

daemon/src/editor_connection.rs
```diff
@@ -0,0 +1,306 @@
...
+            base_dir: base_dir.to_owned(),
+            ot_servers: HashMap::new(),
+        }
+    }
+
+    pub fn owns(&self, file_path: &str) -> bool {
+        self.ot_servers.contains_key(file_path)
+    }
+
+    pub fn message_from_inside(
```
zormit: This name is a bit confusing to me. This method seems to do two things:
- a side effect, but only for an edit
- generate a transformed message to the editor (but only for edits and cursor changes)

I'm wondering if it makes sense to split these two things, as it would be cool if the latter is side-effect free?

zormit: The latter could then be done through an `InsideMessage -> Vec<EditorProtocolMessageToEditor>`-`From` trait. Ah no, it still needs the `rev_text_delta`, but that could be passed in.

zormit (commented):
I have again ripped apart your commits and put them together differently :D I think I'm slowly getting a feel for this PR. The EditorConnection that [was async](https://github.com/ethersync/ethersync/commit/c54eba2cc4b87c34b4779a9ba6890d949486a0fa) is now not part of the history anymore (was easier for me to squash it as it felt like a half revert).

I've left some comments too, but overall it seems to look fine.

daemon/src/editor_connection.rs
zormit: I think I was not aware, that the edit does the transformation as well.

PR #167, opened 2024-09-08
book: make all links relative

zormit:


zormit:
huh! This PR was created in a mild panic where I thought I did break all links by suggesting we use the "absolute" version (without `./`) but that might not be entirely true...

The symptom/bug of dead links is more subtle: If you go to a dead link, like https://ethersync.github.io/ethersync/doesnotexist.html, _then_ (and only then) the table of contents has broken links.

It seems this PR doesn't make a difference for that... It's the [currently deployed version](https://github.com/ethersync/ethersync/actions/runs/10762345189) and the symptom is still there.

blinry:
Fixed in b13f2a8.

Should we still keep the relative links? Probably no need, right?

As you're always curious how I find stuff: I first googled for "mdbook deploy in subdirectory", which gave me the keyword of a "site-url" setting. Then, I googled for "mdbook site-url 404", and the [solution](https://rust-lang.github.io/mdBook/continuous-integration.html#404-handling) was the 5th hit for me…

blinry:
I rebased to only keep the site-url option.

PR #182, opened 2024-09-13
Support Neovim >=0.7

### LINKED ISSUE #56: Support Neovim versions older than 0.10
This would require at least a polyfill for `nvim_create_autocmd` and `vim.fs.root`.

blinry:
The thing blocking us from using olding Neovim versions now are the `nvim_create_autocmd` functions.

Neovim v0.7 (the minimum required version) was released over three years ago in April 2022.

Lazy, for example, requires v8.0.

Fixes #56.

### DIFF
--- a/book/src/neovim.md
+++ b/book/src/neovim.md
@@ -1,9 +1,5 @@
 # Neovim Plugin
 
-> ❗ **Important**
->
-> The plugin currently requires Neovim v0.10.
-
 Again, we have several options of how to install the Neovim plugin:
 
 ## Lazy
--- a/vim-plugin/plugin/ethersync.lua
+++ b/vim-plugin/plugin/ethersync.lua
@@ -94,7 +94,20 @@ end
 
 local function is_ethersync_enabled(filename)
     -- Recusively scan up directories. If we find an .ethersync directory on any level, return true.
-    return vim.fs.root(filename, ".ethersync") ~= nil
+    if vim.version().api_level < 12 then
+        -- In Vim 0.9, do it manually.
+        local path = filename
+        while path ~= "/" do
+            if vim.fn.isdirectory(path .. "/.ethersync") == 1 then
+                return true
+            end
+            path = vim.fn.fnamemodify(path, ":h")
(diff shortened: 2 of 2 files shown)


zormit (approved):
Looks good to me. Should we document our minimum version somewhere, even though it's an old one? Is there a standard for vim plugins to document it? Maybe in a README of `vim-plugin` (it might be nice to add a readme to it anyways)?

Some other plugin related thoughts that came up (might not be necessary to solve in this PR)
- should we think about an automated way to mirror to https://github.com/ethersync/ethersync-vim (i see that you have pushed to it for the 0.4.0 release, nice!)
- is it worth adding different nvim versions to our CI to prevent a regression from compatibility? It's probably possible to add it to the [matrix](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow).

blinry:
> Looks good to me. Should we document our minimum version somewhere, even though it's an old one? Is there a standard for vim plugins to document it? Maybe in a README of `vim-plugin` (it might be nice to add a readme to it anyways)?

Oh totally! Or a `:help` entry! (I'll open an issue!) I feel like it's not important to document the version if we support the past three years.

>     * should we think about an automated way to mirror to [ethersync/ethersync-vim](https://github.com/ethersync/ethersync-vim) (i see that you have pushed to it for the 0.4.0 release, nice!)

It *would* make it easier to install for many people. Just having the plugin in its own repo seems pretty "standard". We could even think about splitting off the plugins at this point, and develop them on their own. The editor protocol is the only "defining interface" now anyways. I'll open an issue to research that!

>     * is it worth adding different nvim versions to our CI to prevent a regression from compatibility? It's probably possible to add it to the [matrix](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow).

I think I'd like that for at least Neovim 0.9! Hopefully, should not make the CI run much longer than now. I'll open an issue!

Issue #94, opened 2024-08-06
Daemon crashes on trying to read non-utf8 file


zormit:
"Could not read file as UTF-8" @ src/daemon.rs:642.

zormit:
Closing as fixed. Follow up in #192.

PR #186, opened 2024-09-13
fix: ignore files that could not be parsed as UTF-8 string

zormit:
A first attempt to approach #94. I guess we could then close the bug and open an enhancement issue that ask for syncing binary files.

Tested on our main repo. We see:
> INFO ThreadId(01) ethersync::daemon: Skipping non-text file /[...]/ethersync/book/src/git-integration-areas.png

### DIFF
--- a/daemon/src/daemon.rs
+++ b/daemon/src/daemon.rs
@@ -737,11 +737,14 @@ impl DocumentActor {
                                     .expect("Could not convert PathBuf to str"),
                             )
                             .expect("Could not convert uri to file path");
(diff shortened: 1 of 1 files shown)


blinry:
I like the approach you're describing! Made some changes:

- Also ignore non-UTF-8 files when they're created or changed while the daemon is running (covering all uses of `from_utf8`).
- Raise the log to a `warn!` – I think here it is reasonable, because it's unexpected behaviour.
- Use the relative file path in the logs, to be consistent with other messages about files.

zormit:
Thanks for catching the other places! 💯

PR #188, opened 2024-09-13
Also allow two newlines when sending to ethersync client

### LINKED ISSUE #187: ethersync client should be less picky on line separators
When testing/developing it's a struggle to get the `ethersync client` message right.

The [JSON Lines](https://jsonlines.org/) spec says:

> 3. Line Separator is '\n'
> This means '\r\n' is also supported because surrounding white space is implicitly ignored when parsing JSON values.
>
> The last character in the file may be a line separator, and it will be treated the same as if there was no line separator present.

but the [LSP Base Protocol spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#baseProtocol) expects `\r\n` between the header and the JSONL:
>  The header and content part are separated by a ‘\r\n’.

This can be confusing (and also inconvenient, because it's hard to see).

We're not building the next internet protocol, just some local "as stupid as possible" inter process communication,  so I would say we could be less strict, and allow simple `\n` separators too?

zormit:
Closes #187.

### DIFF
--- a/daemon/src/jsonrpc_forwarder.rs
+++ b/daemon/src/jsonrpc_forwarder.rs
@@ -70,12 +70,20 @@ impl Decoder for ContentLengthCodec {
         };
 
         // Find the end of the line after that.
-        let end_of_line = match src[start_of_header + c.len()..]
+        let (end_of_line, end_of_line_bytes) = match src[start_of_header + c.len()..]
             .windows(4)
             .position(|window| window == b"\r\n\r\n")
         {
-            Some(pos) => pos,
-            None => return Ok(None),
+            Some(pos) => (pos, 4),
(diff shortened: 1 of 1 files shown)


blinry:
The double matching is not particularly elegant, but I don't have better ideas.

Maybe we should expand the comment to explain what the goal of the code is.

Have you manually tested that the result actually works with \n\n?

zormit:
It was the fastest to get it working without digging too deep into how the Decoder API works. An alternative could be to "match for `\r` => expect `\n\r\n`" / "match for `\n` => expect `\n`, like a [trie state machine](https://craftinginterpreters.com/scanning-on-demand.html#identifiers-and-keywords).

Yup, I tried both cases manually.

PR #193, opened 2024-09-16
Test both Neovim 0.9 and 0.10 in the CI

### LINKED ISSUE #191: Test both Neovim 0.9 and 0.10 in the CI
At least for Neovim's "unit tests". Maybe also for the fuzzer? (That *would* take longer.)

zormit:
Closes #191.


.github/workflows/general.yml
```diff
@@ -57,11 +58,11 @@ jobs:
       - name: Compile integration tests
         working-directory: ./daemon/integration-tests
         run: cargo test --no-run --locked
-      - if: ${{ matrix.os == 'ubuntu-latest' }}
+      - if: ${{ matrix.os == 'ubuntu-latest' && matrix.nvim-version == 'v0.10.1' }}
         name: Run integration tests
         working-directory: ./daemon/integration-tests
         run: cargo test
-      - if: ${{ matrix.os == 'macos-latest' }}
+      - if: ${{ matrix.os == 'macos-latest' || matrix.nvim-version == 'v0.9.5' }}
```
zormit, line 65: We do now have two-by-two = four CI scenarios and *one* of them runs the whole fuzzer. That seems reasonable to me? :)

zormit:
Interestingly this hangs for both OS's for v0.9.5,
> test vim_processes_deltas_correctly has been running for over 60 seconds
> test vim_sends_correct_delta has been running for over 60 seconds

I stopped it. So unfortunately not a "quick win" :D but also not the highest priority to investigate this.

zormit:
Fixed by #210, merging.

PR #216, opened 2024-09-26
Make "edit" messages contain revision and delta on the top level

blinry:
This makes both creating and consuming the "edit" messages easier.


zormit (approved):
good delta!

PR #222, opened 2024-10-04
Path newtypes

zormit:


daemon/src/path.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+/// Paths like these are relative to the project directory.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Eq, Hash, Deref, AsRef)]
+#[as_ref(Path)]
+pub struct RelativePath(PathBuf);
+
+impl RelativePath {
+    // TODO: This doesn't check the parameter in any way. Should it?
```
zormit, line 52: I think it should be indeed a relative path related to the base dir? Can we check that 
somehow when `new` is called?

daemon/src/editor_connection.rs
```diff
@@ -95,52 +99,63 @@ impl EditorConnection {
...
 
                 // We only want to process these messages for files that are not ignored.
-                if sandbox::ignored(&self.base_dir, absolute_file_path)
+                if sandbox::ignored(&self.base_dir, &absolute_path)
                     .expect("Could not check ignore status of opened file")
                 {
                     return Err(EditorProtocolMessageError {
                         code: -1,
-                        message: format!("File '{absolute_file_path:?}' is ignored"),
+                        message: format!("File '{absolute_path:?}' is ignored"),
```
zormit: nitpick: The single quotes were dropped for some of the messages for newtypes, but not in this one. Should it be dropped here as well?

zormit: Tested it, it seems it's quoted twice in this case

```
[ethersync] Error for 'open': File '"/Users/mn/code/ethersync/ethersync/daemon/playground/.ethersync/config"' is ignored (This file should not be shared with other peers)
```

(byy the way, this is a strange error message to show to the user :D but that's another issue)

daemon/src/daemon.rs
```diff
@@ -366,21 +352,16 @@ impl DocumentActor {
...
+                    .expect("Watcher event should have a path within the base directory");
                 if self.owns(&relative_file_path) {
                     if !self.crdt_doc.file_exists(&relative_file_path) {
                         let content = sandbox::read_file(&self.base_dir, Path::new(&file_path))
                             .expect("Failed to read newly created file");
                         if let Ok(content) = String::from_utf8(content) {
                             self.crdt_doc.initialize_text(&content, &relative_file_path);
                         } else {
-                            warn!("Ignoring newly created non-UTF-8 file '{relative_file_path}'");
+                            warn!("Ignoring newly created non-UTF-8 file {relative_file_path:?}");
```
zormit: I'm not a big fan of using the Debug representation for user facing messages. It feels somewhat hacky and potentially brittle.

>[fmt::Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html) implementations should be implemented for all public types. Output will typically represent the internal state as faithfully as possible. The purpose of the [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html) trait is to facilitate debugging Rust code. [...]

What if that "internal state" represents itself differently at some point? It seems outside our control unless we implement the trait ourselves (and not derive it).

But I get that it's pragmatic, so at this point I could still accept it :) There's also some effort to at least unify the approach if we do it, see the other comments.

daemon/src/path.rs
```diff
@@ -0,0 +1,157 @@
...
+        let path = Path::new(&self.0[7..]);
+        AbsolutePath::try_from(path.to_path_buf())
+            .expect("File URI should contain an absolute path")
+    }
+}
+
+impl TryFrom<String> for FileUri {
+    type Error = anyhow::Error;
+
+    fn try_from(string: String) -> Result<Self, Self::Error> {
```
zormit, line 104: This could be taking a &str, probably, then the caller doesn't need to clone?

daemon/src/path.rs
```diff
@@ -0,0 +1,157 @@
+use anyhow::{bail, Context};
+use automerge::Prop;
+use derive_more::{AsRef, Deref};
+use serde::{Deserialize, Serialize};
+use std::path::{self, Path, PathBuf};
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Eq, Hash, Deref, AsRef)]
+#[as_ref(Path)]
+pub struct AbsolutePath(PathBuf);
```
zormit, line 11: docstring could be nice :)

daemon/src/path.rs
```diff
@@ -0,0 +1,157 @@
...
+#[as_ref(Path)]
+pub struct RelativePath(PathBuf);
+
+impl RelativePath {
+    // TODO: This doesn't check the parameter in any way. Should it?
+    pub fn new(path: &str) -> Self {
+        Self(path.into())
+    }
+
+    pub fn try_from_absolute(path: &AbsolutePath, base_dir: &Path) -> Result<Self, anyhow::Error> {
```
zormit: nitpick: The parameters to this are the other way around as the one in `AbsolutePath::from_parts`. Should we unify that?

zormit (commented):
I think this already looks like a good start. We could use this pattern in more places, I assume. Also it seems that due to all the `TryFrom` implementations we now have lot's more Results to unpack. I'm wondering if there's a way around that.

daemon/src/daemon.rs
blinry: Another downside is that (I think), the Debug output would escape spaces, which doesn't seem desirable in user-facing strings. I agree that we should implement the `Display` trait instead.

daemon/src/path.rs
blinry: Absolutely!

daemon/src/path.rs
blinry: Left to the reader as an exercise (a future refactoring), I guess!

daemon/src/path.rs
blinry: In our case, then the conversion has to clone. I *think* it is good practice to make that explicit to the caller. I remember reading that, if the parameter is to be stored somewhere, like in this case, it should be owned already.

zormit: Interesting, that's contrary to my heuristics, which is something like "parameters should always be &str, if possible, to make the interface more widely applicable".

Digging a bit into it, let's take [this personal rules of thumb](https://users.rust-lang.org/t/returning-strings-string-vs-str/86286/7):
> 1. "&str for parameters; String for return types". This always works, and has no borrowck complications, but is sometimes inefficient.
>
> 2. the small tweak to add "If you always need to to_owned/to_string the &str parameter, change it to a String parameter". That's more efficient when your caller happens to already have a String, no less efficient if they don't, and still doesn't require thinking about lifetimes.
>
> 3. if you're always returning a substring of one of the arguments, add a lifetime generic and return a &'a str, updating the corresponding parameter from &str to &'a str as well.

So it seems I was having rule 1 in mind, while you are maybe talking about something related to rule 2? I think it makes sense: A potential caller could already have a `String`, that way we prevent an unnecessary clone of that inside `try_from`.

=> resolving and accepting your approach :)

daemon/src/path.rs
zormit: Hehe yeah, I don't see a good way to do that with this setup, and I think that's why I came to the solution that I drafted to you (using a `Project` that is the only place that can construct `ProjectPath`s).

PR #223, opened 2024-10-04
Drop --random flag (used only in development) and repower fuzzer

zormit:
- **bump ethersync version in Cargo.lock**
- **Revert "add random flag for manual fuzzing"**

### DIFF
--- a/daemon/Cargo.lock
+++ b/daemon/Cargo.lock
@@ -802,7 +802,7 @@ dependencies = [
 
 [[package]]
 name = "ethersync"
-version = "0.4.0"
+version = "0.5.0"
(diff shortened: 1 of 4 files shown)


daemon/Cargo.lock
```diff
@@ -802,7 +802,7 @@ dependencies = [
 
 [[package]]
 name = "ethersync"
-version = "0.4.0"
```
zormit, line 805: Not sure why this version was still on 0.4.0, but I think it can/should be bumped too. Some magic (probably rust analyzer) did that for me and I was wondering why I'm getting an unclean git state.

PR #257, opened 2024-12-22
Intellij Plugin

### LINKED ISSUE #256: Intellij Plugin
First of all really great project and I love [your talk](https://talks.mrmcd.net/2024/talk/V7BRVP/). :heart_eyes: And also even more fantastic that this project supports Neovim out of the box. Love it.

Now, here comes the thing: I work at a Java shop and many Java people use Intellij, so I was wondering if there is a plan to support Intellij because I assume ethersync would be very useful at work?

FYI: I know that you asking for support in the talk. I created this issue to reflect that in your issue tracker.

schrieveslaach:
This PR provides an Intellij Plugin for ethersync.

https://github.com/user-attachments/assets/e9276de2-b78b-4d2d-921b-39e663fd272d

Fixes #256


intellij-plugin/src/main/kotlin/io/github/ethersync/protocol/CursorEvent.kt
```diff
@@ -0,0 +1,15 @@
+package io.github.ethersync.protocol
+
+import com.google.gson.annotations.SerializedName
+import org.eclipse.lsp4j.Range
+
+data class CursorEvent(
+   @SerializedName("userid")
+   val userId: String,
```
schrieveslaach, line 8: The editor dev guide says

> #### `"cursor" {userid: integer, name?: string, uri: DocumentUri, ranges: Range[]}`

However, I've got a string. @blinry, is that a typo in the doc?

intellij-plugin/src/main/kotlin/io/github/ethersync/protocol/EthersyncEditorProtocol.kt
```diff
@@ -0,0 +1,12 @@
+package io.github.ethersync.protocol
+
+import com.google.gson.JsonObject
+import org.eclipse.lsp4j.jsonrpc.services.JsonNotification
+import org.eclipse.lsp4j.jsonrpc.services.JsonRequest
+import java.util.concurrent.CompletableFuture
+
+interface EthersyncEditorProtocol {
+   @JsonNotification("cursor")
+   fun cursor(cursorEvent: CursorEvent)
```
schrieveslaach, line 8: @blinry, is it possible to be notified about the own user id? This would make it a bit easier to filter out, the own cursor events.

schrieveslaach:
@blinry, I think this PR is ready for review and testing.

intellij-plugin/src/main/kotlin/io/github/ethersync/protocol/CursorEvent.kt
blinry: Indeed, looks like we've been using strings so far! Fixed the documentation in #261, but I'm wondering whether we should actually switch to integers, which seem to make better IDs…

intellij-plugin/src/main/kotlin/io/github/ethersync/protocol/EthersyncEditorProtocol.kt
blinry: Can you briefly explain how this would help you/how you would use it?

Would you like to know your userid, or actually get back "cursor" messages for your own cursor movements?

intellij-plugin/src/main/kotlin/io/github/ethersync/protocol/CursorEvent.kt
schrieveslaach: If I remember correctly, the `userId` contains the local peer ID + the client ID (which is a increasing integer). So, String makes sense.

intellij-plugin/src/main/kotlin/io/github/ethersync/protocol/EthersyncEditorProtocol.kt
schrieveslaach: I had a situation where I got my own cursor events and thus IntelliJ rendered my cursor as a remote cursor. I could be that this was an testing artifact but I was thinking that I maybe, need to filter my own events too.

blinry: I think this must've been an artifact – you normally shouldn't get back your own cursors. If you notice that again, I'd consider it a bug :)

schrieveslaach:
@blinry, there is now DEVELOPMENT.md containing build and install instructions. What do you think?

blinry:
Thanks for the README! :) I tried to get the plugin to build on NixOS, but I'm giving up for today.

Some references (for my future self):

- https://github.com/NixOS/nix-idea/issues/30
- https://github.com/NixOS/nix-idea/pull/31
- https://discourse.nixos.org/t/how-to-create-a-development-environment-with-intellij-idea-and-openjdk/10153
- https://gitlab.uni-marburg.de/langbeid/powersort

@schrieveslaach For now, could you attach a compiled ZIP plugin for testing? :sweat_smile:

blinry:
Ah, I made it work, using the following hack:

```bash
mv ~/.gradle/caches/8.10.2/transforms/c491cd297192d77eda951eddd460124d/transformed/ideaIC-2024.3.1.1 /tmp/
ln -s /nix/store/vr4jmrihg57ss2z5pqmnchqwflhnclji-idea-community-2024.3.1.1/idea-community ~/.gradle/caches/8.10.2/transforms/c491cd297192d77eda951eddd460124d/transformed/ideaIC-2024.3.1.1
```

Testing tomorrow!

schrieveslaach:
> Testing tomorrow!

Exciting. Looking forward to your response.

blinry:
It works! :tada: Very exciting, this is the first contributed plugin that I'm trying out!

Many things seem to work really well, like any UTF-8 shenanigans I tried to throw at it! :D Fantastic work, @schrieveslaach!

Didn't look much at the code yet, here's some things I noticed from a "user's perspective":

- When selecting something "backwards" in Neovim (so that the "end" of the selection is in front of the "start"), the selection is not displayed in Intellij. (We decided to allow those selections, so that editors could mark the "active end" of the selection.)
- When selecting something in Intellij, Neovim seems to receive only the current cursor position, not the full selection.
- When receiving a cursor position on an empty line, the Intellij plugin doesn't render it.
- Starting the daemon from within Intellij, it uses a `--socket-name ethersync-playground-socket`, which was unexpected – at first, I tried to use the default socket with another editor. The socket is our workaround for having more than one Ethersync daemon at a time – in the long term, we might add multi-project support to the daemon. But in this plugin, it might make sense to use the default socket.
- The icon in the menu looks rather big on my setup, which might be hard to reproduce (using the niri window manager on Wayland with some screen scaling):

image

- Is there a way to stop the daemon spawned by Intellij?
- Say I wanted to follow the current "standard model", where one launches the daemon on its own, outside of any editor. Could I currently do it using the plugin?
- It sometimes makes sense to use Ethersync with a password, so others in the same network can't join. We use the [configuration files](https://ethersync.github.io/ethersync/configuration.html) to store them. I wonder how this feature could be smoothly integrated into the current plugin. Again, I feel a bit uneasy about integrating the daemon launch into the plugin, because I expect it will change quite a bit. In our thinking so far, `ethersync daemon` is not part of the "editor interface", instead, `ethersync client` and the editor sync protocol is. (For the daemon, we might switch to an `ethersync share`/`ethersync join` interface at some point). But I get your reasoning that people might expect such a functionality. I'm happy about experimenting on this front, and finding the best solutions!
- I noticed this warning in the output of my `idea-community` process:

image

Again, thanks for all the work you put into this! Do you have time and would you enjoy taking care of some of the issues above? If you'd prefer, you could also put this plugin into your own repo first, and we could iterate on it using an issue tracker.

schrieveslaach:
Thank you very much for the feedback and thanks for collecting the things you noticed. I'm very happy and I would like to continue working on it. Ideally, in a new repo in which I could push the commits of the sub-directory. Do you mind to set this up?

Maybe, the things you collected could end up as issues there?

When we have that we could have a GitHub action that builds the latest main line version and offers a download of the plugin. I have some people at work who would like to test it.

> Also, I'd love to chat some more on your impressions of the JSON-RPC protocol, and what was easy/hard when implementing it, if you'd enjoy that!

I'm still open to have a video-chat on the protocol. Sure, write me an [DM on Mastodon](https://fosstodon.org/@schrieveslaach) or grep the e-mail from my commits. Then we could arrange something to chat. Maybe, the daemon handling is also worthwhile having a chat.

schrieveslaach:
As discussed in our video chat the plugin code is now located https://github.com/ethersync/ethersync-jetbrains and I created a set of issues to track your observations. Closing this PR.

Issue #258, opened 2024-12-29
Windows support


d0mmi:
can the daemon run on windows?
i tried installing it with "cargo install ethersync"
but got this error:
"could not find `unix` in `os`"

d0mmi:
will be fixed with #260

alerque:
@d0mmi Would you (or anyone listening!) happen to be around to test building on Windows? For example I have [this branch](https://github.com/alerque/teamtype/tree/windoze) that may or may not fix this issue but I'm basically shooting in the dark. I believe it could be tested as simply as this:

```console
$ cargo install --locked --git https://github.com/alerque/teamtype --branch windoze
```

zormit:
closing in favor of #485

PR #260, opened 2025-01-10
Adding windows support to daemon and vscode extension

d0mmi:
- adjusted the daemon to use pipes instead of unix streams when building for windows
- adjusted daemon and vscode extension to be able to handle unix and windows file paths

### DIFF
--- a/daemon/src/editor.rs
+++ b/daemon/src/editor.rs
@@ -1,18 +1,13 @@
 //! This module is all about daemon to editor communication.
 use crate::daemon::{DocMessage, DocumentActorHandle};
-use crate::sandbox;
 use crate::types::EditorProtocolObject;
-use anyhow::{bail, Context, Result};
+use anyhow::{Result};
 use futures::StreamExt;
 use std::{
-    fs, io,
-    os::unix::fs::PermissionsExt,
+    io,
     path::{Path, PathBuf},
 };
-use tokio::{
-    io::WriteHalf,
-    net::{UnixListener, UnixStream},
-};
+use tokio::io::WriteHalf;
 use tokio_util::{
     bytes::BytesMut,
     codec::{Encoder, FramedRead, FramedWrite, LinesCodec},
@@ -20,8 +15,6 @@ use tokio_util::{
 use tracing::info;
 
 pub type EditorId = usize;
-
-pub type EditorWriter = FramedWrite<WriteHalf<UnixStream>, EditorProtocolCodec>;
 
 pub struct EditorProtocolCodec;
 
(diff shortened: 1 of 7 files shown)


d0mmi:
related to #258

blinry:
Hello! Thanks for your work on this! First-class Windows support would be neat to have. I'll try to get a proper code review in soon.

On first glance, I'm a bit unhappy with the many feature-flagged duplications, which makes it harder to maintain. Building a modular solution here would be great! Ideally, there could be a trait defining an interface for communication with the daemon, which can then be implemented by the "socket approach" and the "named pipe approach", I think. Would you enjoy working on that/keep refining the PR?

From a security standpoint, for the Unix sockets, we make very sure that they're only accessible by the user who launches the daemon. Do named pipes have a similar property? For example, if one user starts the daemon, and then a second user logs into the same computer, could that second user connect to the pipe?

Are there other projects which have a similar client-daemon system, where we could look at how they solved it?

d0mmi:
> On first glance, I'm a bit unhappy with the many feature-flagged duplications, which makes it harder to maintain. Building a modular solution here would be great! Ideally, there could be a trait defining an interface for communication with the daemon, which can then be implemented by the "socket approach" and the "named pipe approach", I think. Would you enjoy working on that/keep refining the PR?


That makes sense its currently more like a proof of concept and refactoring that for better maintainace makes a lot of sense.
In the current implementation i just tried to touch as little places as possible. But i will have a look on how this could be integrated in a cleaner way.


> From a security standpoint, for the Unix sockets, we make very sure that they're only accessible by the user who launches the daemon. Do named pipes have a similar property? For example, if one user starts the daemon, and then a second user logs into the same computer, could that second user connect to the pipe?


Im also very new to windows pipes so i will have to check that


> Are there other projects which have a similar client-daemon system, where we could look at how they solved it?


Im very new to this client-daemon system and dont know any other projects that use something like this.
Usually i work on games or boring java business logic ^^

blinry:
I like the direction this is heading in a lot! :sparkles: Let us know once you're ready for more feedback or in need of guidance!

d0mmi:
I was struggling a lot trying to figure out how those "interfaces" work in rust 😅. But i think I learned some stuff now and will continue cleaning up more of the code when I find the time ^^

d0mmi:
just wanted to check if i could rebase my change on the master version as i have not looked at it in a longer time but there were so many changes im not sure if it would be faster to start from 0 😅

zormit:
Hi @d0mmi my answer is a vague: "Sure, if you want to! :)" which means that we are happy about you working on this, but we can not promise to put a lot of effort ourselves into Windows support right now. Happy to review what you come up with, but not able to contribute a lot of brainpower into it, as we are focusing on other features right now. This might change in the future, ultimately we'd love to be able to support Windows.

As for merge conflicts vs. rewriting from scratch, some hints: Since January and now we have changed the underlying library how peers are connecting (see #265 for details), also we have changed the CLI interface (see #273), which means that it might not even be possible anymore to use your code in those files.

From looking at what you have so far we liked the modular approach that you used in 82fb986.

Also there's still the [open question](https://github.com/ethersync/ethersync/pull/260#issuecomment-2588126101) of the security properties of Windows pipes. As our software is quite powerful (write permissions to disk via exposed interfaces), we want to do our best to prevent widening the attack surface.

PowerUser64:
Is there any way to test this out?

d0mmi:
> Is there any way to test this out?

im currently working on a rework based on the new changes once this is done i will push it to this pull request

blinry:
Another option, if you just want to try Ethersync on Windows, @PowerUser64, would be to run it in the WSL. Especially the Neovim plugin should work well there.

zormit:
We discussed this and would focus on a refactoring first, which makes most sense in a new PR.

PR #263, opened 2025-02-04
Update neovim.md

Nils-Magnus:
Point out minimal requirement of 0.7.0 for Neovim and fix a minor inconsistency with the error message.


blinry:
Thanks a lot, Nils-Magnus, that's a good thing to point out! :)

PR #265, opened 2025-06-02
Switch from libp2p to Iroh

blinry:
I'm mostly curious whether the integration test will succeed here…


daemon/src/peer.rs
```diff
@@ -230,15 +149,29 @@ impl P2PActor {
...
+
+            let bytes = secret_key.to_bytes();
+
+            let mut rng = rand::rngs::OsRng;
+            let passphrase: String = (0..64)
+                .map(|_| rng.sample(rand::distributions::Alphanumeric) as char)
+                .collect();
 
             let mut file = OpenOptions::new()
                 .create(true)
```
zormit, line 194: we noticed that the truncate of the existing code might not be necessary, because there's an existence check anyways

daemon/src/peer.rs
```diff
@@ -100,125 +76,68 @@ impl P2PActor {
...
-                    debug!("{:?}", error);
-                }
-                libp2p::swarm::SwarmEvent::IncomingConnectionError { error, .. } => {
-                    error!("Someone tried to connect to you, but failed. The secret they provided might be wrong?");
-                    debug!("{:?}", error);
-                }
-                event => debug!(?event),
+        let endpoint = iroh::Endpoint::builder()
+            .secret_key(secret_key)
+            .alpns(vec![b"ethersync".to_vec()])
```
zormit: - [x] is this the correct ALPNS format?

daemon/src/peer.rs
```diff
@@ -100,125 +76,68 @@ impl P2PActor {
...
+
+        let address = format!("{}#{}", endpoint.node_id(), my_passphrase);
+
+        info!(
+            "Others can connect with:\n\n\tethersync daemon --peer\n{}\n",
+            address
+        );
+
+        // TODO: Use the error?
+        let _ = address_tx.send(address);
```
zormit: ( hack to get the address to the fuzzer )

daemon/src/peer.rs
```diff
@@ -8,40 +8,23 @@
...
 use std::mem;
-use std::net::Ipv4Addr;
 use std::os::unix::fs::{OpenOptionsExt, PermissionsExt};
 use std::path::{Path, PathBuf};
+use std::str::FromStr;
 use tokio::sync::{broadcast, mpsc, oneshot};
-use tokio::time::Duration;
 use tracing::{debug, error, info, warn};
 
-const ETHERSYNC_PROTOCOL: StreamProtocol = StreamProtocol::new("/ethersync");
```
zormit, line 34: reuse this for the ALPN

daemon/src/peer.rs
blinry: I tried it, and it is indeed not necessary. Removed.

daemon/src/peer.rs
blinry: Iroh uses "/[protocol]/[version]" – and even plan on supporting multiple versions at once!

blinry:
We discussed the quality of randomness produced by `rand::rngs::OsRng`. It implements `CryptoRng`, a marker trait only implemented by randomness sources that are supposed to be cryptographically secure.

To my understanding, `OsRng` refers to the operating system's best source of randomness. So I think this is fine. Iroh also [advises](https://docs.rs/iroh/latest/iroh/struct.SecretKey.html#method.generate) to use it.

book/src/connection-making.md
```diff
@@ -9,9 +9,9 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 
 In order to make sense of how Ethersync daemons connect to each other a little bit of Networking background (IP addresses, TCP ports) is helpful. You should still be able to get going within one local network (such as two computers in the same Wi-Fi) by just copy pasting things, but connecting to other peers over the internet might currently require some configurations. We're aiming to give you the right keywords to look for in case you've not encountered that yet.
 
-## Multiaddress
+## Addressing and authenticating with the peer
```
zormit, line 10: We need a crisp name for "node id + secret key". It's definitely not a multiaddress anymore, but as you can see I felt like I have to "describe it" in all the places.

book/src/connection-making.md
```diff
@@ -9,9 +9,9 @@ SPDX-License-Identifier: CC-BY-SA-4.0
...
 
-Ethersync uses libp2p for making a connection. To identify another daemon we're using a connection address like `/ip4/192.168.23.42/tcp/58063/p2p/12D3KooWPNj7mom3X2D6NiSyxbFa5hHfzxDFP98ZL52yYnkEVmDv`. This is what libp2p calls a [multiaddress](https://docs.libp2p.io/concepts/fundamentals/addressing/) – it contains your IP address, the TCP port, and a "peer ID" (which is used by connecting peers to make sure that they're actually connecting to the correct peer, and not to a "man in the middle").
+Since version 0.7.0 Ethersync uses iroh for making a connection. To identify another daemon we're using a combination of the iroh [Node Identifier](https://www.iroh.computer/docs/concepts/endpoint#node-identifiers) and a secret key which, smashed together, looks like `429e94...0e9819#32374e...4a6789`. Treat it like a password and share it with your peer through a secret channel. In the meantime, we're working on an approach which doesn't require sharing such a long random and sensitive string.
```
zormit: doc versioning would be overkill, I guess? 😬

zormit: Also, different topic, same line: note last sentence. Overpromised? We will do something with whisper links soon, won't we? :)

book/src/configuration.md
```diff
@@ -13,17 +13,16 @@ There are two ways to configure what an Ethersync daemon will do.
...
 
 ## Configuration files
 
-If you keep starting Ethersync with the same options, you can also put the following options into a configuration file at `.ethersync/config`. Here, you have the option to specify a secret; otherwise, a default password will be used, which is only recommended for testing:
+If you keep starting Ethersync with the same options, you can also put the following options into a configuration file at `.ethersync/config`:
 
 ```ini
-peer = <multiaddr you want to try connecting to>
+peer = <peer "id#secret" you want to try connecting to>
 port = <port for your daemon>
```
zormit: Oh, I think port isn't needed anymore either! 😮 
- [x] remove port from docs

daemon/src/main.rs
```diff
@@ -43,12 +44,9 @@ enum Commands {
     Daemon {
         /// The directory to sync. Defaults to current directory.
         directory: Option<PathBuf>,
-        /// Multiaddr of a peer to connect to.
+        /// Read <node_id>#<passphrase> from stdin.
```
zormit: This is user visible documentation, isn't it? So I'd rephrase it to be a bit less technical:
```suggestion
        /// Connect to another peer. Will ask for <node_id>#<passphrase> to connect to.
```

book/src/connection-making.md
blinry: Iroh calls it a "ticket": https://www.iroh.computer/docs/concepts/tickets

And I think Ink & Switch have made good experiences with sharing URLs…

Maybe "node address" would work?

daemon/src/main.rs
blinry: I like that!

book/src/connection-making.md
zormit: I think my (maybe nitpicky) point is that it's a ticket/node address _plus_ a passphrase/key. When writing the docs I just wanted to make sure that it's very clear that it's sensitive information to be handled with care. "Node Address" or "ticket" alone doesn't sound overly sensitive.

OTOH let's not overthink it, I assume the docs might change again soon when we will integrate .

PR #268, opened 2025-06-09
When a connection times out, reconnect

blinry:
These reconnections seem useful if you basically want to "stay connected" to your peer forever.

This PR restructures the peer code around a "ConnectionManager", which spawns an "EndpointActor", which can receive requests to make connections. When the EndpointActor decides that a reconnect makes sense, it will send such a message to itself.

It's the first time I needed to do some more "proper" error handling, because we want to be able to tell whether the reason for a failure was a timeout, or something else a couple of layers higher up. It's not really pretty.

## Ideas for how to make stuff prettier

(But maybe that's for a later PR?)

The main other failure reason I see is an authorization failure. The joining peer will not get any feedback on that failure, unfortunately – its next communication attempt will just fail. We could modify the peer protocol a bit so that we can tell earlier when an auth failure happens, and then do two steps:

1. Try auth, if it fails, always stop
2. Start regular connection, if it fails, always reconnect?

### DIFF
--- a/daemon/src/peer.rs
+++ b/daemon/src/peer.rs
@@ -3,7 +3,7 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-//! A peer is another daemon. This module is all about daemon to daemon communication.
+//! This module provides a ConnectionManager, which can be used to connect to other daemons.
 
 use self::sync::SyncActor;
 use crate::daemon::DocumentActorHandle;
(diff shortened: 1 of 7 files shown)


daemon/src/peer.rs
```diff
@@ -178,225 +161,315 @@ impl P2PActor {
...
+        tokio::spawn(async move {
+            let _ = Self::read_loop(receive, message_tx).await;
+        });
+
+        Ok(Self { send, message_rx })
+    }
+
+    async fn read_loop(
+        mut receive: RecvStream,
+        message_tx: mpsc::Sender<Result<PeerMessage, ConnectionError>>,
```
zormit, line 405: [variable naming] Would it be okay to rename `message_tx/rx` to `endpoint_message_tx/rx` and `peer_message_tx/rx` respectively?

zormit (approved):
I don't understand the whole of it, but it seems fine to merge and iterate on it :)

daemon/src/peer.rs
zormit: actually I tried it and it was a bit ugly to have

```
struct EndpointActor {
    endpoint: iroh::Endpoint,
    endpoint_message_rx: mpsc::Receiver<EndpointMessage>,
    endpoint_message_tx: mpsc::Sender<EndpointMessage>,
    my_passphrase: SecretKey,
}
```
so I'd leave it as it is for now.

PR #269, opened 2025-06-09
Update flakes & crates, support + test Neovim 11

blinry:
Let's see if this passes the CI?

- [x] Add Neovim 11 to the test matrix?

### DIFF
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -36,7 +36,7 @@ required-features = ["executable-deps"]
 
 [dependencies]
 anyhow = "1.0.81"
-automerge = "0.5.9"
+automerge = "0.6.1"
 clap = { version = "4.5.3", features = ["derive", "env"], optional = true }
 constant_time_eq = "0.3.1"
 derive_more = { version = "2.0.1", features = ["as_ref", "deref", "display"] }
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -38,14 +38,14 @@ required-features = ["executable-deps"]
 anyhow = "1.0.81"
 automerge = "0.5.9"
 clap = { version = "4.5.3", features = ["derive", "env"], optional = true }
-constant_time_eq = "0.3.1"
-derive_more = { version = "1.0.0", features = ["as_ref", "deref", "display"] }
+constant_time_eq = "0.4.2"
+derive_more = { version = "2.0.1", features = ["as_ref", "deref", "display"] }
 dissimilar = "1.0.9"
 futures = { version = "0.3.30", default-features = false }
 ignore = "0.4.22"
 iroh = "0.35.0"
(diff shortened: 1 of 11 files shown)


blinry:
Insight from checking all updates step-by-step: All updates pose no problems – except the update to Automerge 1.0.0-beta, which breaks our fuzzer! :O

So I'll not include that update in this PR for now. Options:

- Try to debug why it breaks – maybe their (new) `TextRepresentation::String(TextEncoding::UnicodeCodePoint)` handling has bugs? Or our usage needs to be adapted more? This would probably require getting the fuzzer to run locally again...
- Wait for the 1.0.0 release. We wouldn't get those sweet, sweet performance boost yet, though :/

daemon/Cargo.toml
```diff
@@ -38,14 +38,14 @@ required-features = ["executable-deps"]
...
 dissimilar = "1.0.9"
 futures = { version = "0.3.30", default-features = false }
 ignore = "0.4.22"
 iroh = "0.35.0"
 magic-wormhole = "0.7.6"
-notify = { version = "6.1.1", default-features = false, features = [
+notify = { version = "8.0.0", default-features = false, features = [
     "macos_fsevent",
 ] }
 operational-transform = "0.6.1"
```
zormit, line 51: Seems we can also upgrade `rand` (in line 53) to 0.9.1?

daemon/Cargo.toml
```diff
@@ -38,14 +38,14 @@ required-features = ["executable-deps"]
 anyhow = "1.0.81"
 automerge = "0.5.9"
```
zormit: if not 1.0, should we at least bump the minor version to 0.6.1?

flake.lock
zormit, line 1: For my understanding/learning: How was this file upgraded?

zormit:
Since I was talking about the fact that `strip=false`, maybe we can fix this TODO:

https://github.com/ethersync/ethersync/blob/b4ed43b5353ac90af639c2796bb14096f73337e6/daemon/Cargo.toml#L87

Which came up in https://github.com/ethersync/ethersync/issues/244 (solved). I saw it while thinking about the MacOS build issue, but I guess we can fix it here and try to solve the problem independently.

daemon/Cargo.toml
zormit: I found your commit message of a2bbe704, which explains it:

> The only other crate we cannot update is the rand crate: iroh currently requires v0.8.

daemon/Cargo.toml
zormit: i think it's possible ✅

flake.lock
zormit: `nix flake update` bumps the dependencies ("inputs") of the flake https://nix.dev/manual/nix/2.25/command-ref/new-cli/nix3-flake-update

Issue #271, opened 2025-06-11
Don't display secrets on command line


blinry:
I just noticed that the iroh PR introduced a security regression: The daemon's full "connection address" (Iroh calls them [tickets](https://www.iroh.computer/docs/concepts/tickets)) is displayed. Previously, we displayed something like "your peers will need the secret from your config file".

Displaying the full ticket is **good**, because:

- It's much easier to just copy-paste it from there, compared to… Ethersync writing the ticket into `.ethersync/config`? (Also doesn't make much sense.)

It is **bad**, because:

- If someone were to launch a daemon in front of an audience (on a stream, in a talk), everyone would be able to connect to it. This could be done accidentally, and Ethersync doesn't warn users that this will happen.
- It seems nice to build tools that are secure by default.

It is **neutral**, because:

- That's how other tools (magic-wormhole, screensy, pads with obscure document names) work, as well. So maybe people are used to "don't show sensitive things on screen"? (Google Docs, on the other hand, requires an additional click to get the share-link.)

Here's an **alternative option**: For now, we could *directly put the ticket into the user's clipboard*, and print a message about it. If the user needs to access it later again, they will need to restart the daemon.

If we build a CLI interface to configure daemon sessions later, they could have an `ethersync status --show-secrets` command! :D

zormit:
As of #273 it's now possible to steer the behaviour of how secrets are displayed: The full connection address isn't shown by default, but can be displayed with a flag (`--show-secret-address`). Also the less critical join code can be disabled with `--no-join-code`, e.g. in a screensharing situation where the watchers should not see the code.

PR #273, opened 2025-06-11
Redesigning the CLI UI

zormit:
- [x] rewrite docs / book
- [x] #271


blinry (approved):
I'm happy with this redesign! I documented the new config options, and prevented `ethersync share` from ever attempting peer connections (even if there's one in the config file). Let's try this and see how we like it!

PR #276, opened 2025-06-16
Use .ethersync/socket as daemon socket

blinry:
This has the advantage that two daemons started on different directories will never clash - useful for multi-daemon scenarios, and for local testing.

One downside is that the editor plugins need to do a bit more work: They need to figure out which directory to start `ethersync client` in. However, our two plugins already had all the code required to find it, because they needed to scan for the .ethersync directory anyway.

The --directory flag would be global for all subcommands.

What do you think? I think this is worth the additional editor code, and simplifies things a lot.

- [x] Update documentation (aka delete all references to sockets ever, hopefully!)
- [x] Provide a smooth upgrade path? (Currently, it's important to do `chmod go-rwx .ethersync`.)
	- Maybe we should drop this permission requirement now? I think basically all other programs leave it up to the user to set permissions on files in their own directories like they want... I did that for now in this PR.

Possible refactorings:

- [x] Move the constants like DEFAULT_SOCKET_NAME out into the `ethersync` crate, and use them everywhere else.

### DIFF
--- a/daemon/integration-tests/src/actors.rs
+++ b/daemon/integration-tests/src/actors.rs
@@ -159,15 +159,18 @@ fn rand_usize_inclusive(start: usize, end: usize) -> usize {
 }
 
 impl Neovim {
-    pub async fn new_ethersync_enabled(initial_content: &str) -> (Self, PathBuf) {
+    pub async fn new_ethersync_enabled(initial_content: &str) -> (Self, PathBuf, PathBuf) {
         let dir = TempDir::new().unwrap();
         let ethersync_dir = dir.child(".ethersync");
         sandbox::create_dir(dir.path(), &ethersync_dir).unwrap();
+
         let mut file_path = dir.child("test");
         sandbox::write_file(dir.path(), &file_path, initial_content.as_bytes())
             .expect("Failed to write initial file content");
         file_path = fs::canonicalize(file_path).expect("Could not canonicalize");
 
-        (Self::new(file_path.clone()).await, file_path)
+        let socket_path = dir.child("socket");
+
+        (Self::new(file_path.clone()).await, file_path, socket_path)
     }
 }
--- a/daemon/integration-tests/src/actors.rs
+++ b/daemon/integration-tests/src/actors.rs
@@ -3,6 +3,8 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
+use crate::socket::*;
+
 use ethersync::daemon::Daemon;
 use ethersync::sandbox;
 
@@ -42,6 +44,7 @@ impl Neovim {
         nvim.command(&format!("edit! {}", file_path.display()))
             .await
             .expect("Opening file in nvim failed");
+
         let buffer = nvim.get_current_buf().await.unwrap();
 
         Self { nvim, buffer }
@@ -160,7 +163,9 @@ fn rand_usize_inclusive(start: usize, end: usize) -> usize {
 
 impl Neovim {
     // The caller should store the TempDir, so that it is not garbage collected.
-    pub async fn new_ethersync_enabled(initial_content: &str) -> (Self, PathBuf, PathBuf, TempDir) {
+    pub async fn new_ethersync_enabled(
+        initial_content: &str,
+    ) -> (Self, PathBuf, MockSocket, TempDir) {
         let dir = TempDir::new().unwrap();
         let ethersync_dir = dir.child(".ethersync");
         let file_path = dir.child("test");
@@ -173,10 +178,12 @@ impl Neovim {
 
         let canonicalized_file_path = fs::canonicalize(&file_path).expect("Could not canonicalize");
 
+        let socket = MockSocket::new(&socket_path);
+
         (
             Self::new(file_path.clone()).await,
             canonicalized_file_path,
-            socket_path,
+            socket,
             dir,
         )
     }
--- a/daemon/integration-tests/src/actors.rs
+++ b/daemon/integration-tests/src/actors.rs
@@ -159,18 +159,25 @@ fn rand_usize_inclusive(start: usize, end: usize) -> usize {
 }
 
 impl Neovim {
-    pub async fn new_ethersync_enabled(initial_content: &str) -> (Self, PathBuf, PathBuf) {
+    // The caller should store the TempDir, so that it is not garbage collected.
+    pub async fn new_ethersync_enabled(initial_content: &str) -> (Self, PathBuf, PathBuf, TempDir) {
         let dir = TempDir::new().unwrap();
         let ethersync_dir = dir.child(".ethersync");
+        let file_path = dir.child("test");
+        let socket_path = ethersync_dir.clone().join("socket");
+
         sandbox::create_dir(dir.path(), &ethersync_dir).unwrap();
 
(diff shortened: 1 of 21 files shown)


vscode-plugin/src/extension.ts
```diff
@@ -260,8 +266,17 @@ async function applyEdit(document: vscode.TextDocument, edit: Edit): Promise<boo
...
 
-// TODO: check if belongs to project.
 async function processUserOpen(document: vscode.TextDocument) {
+    const filename = document.fileName
+    const directory = findEthersyncDirectory(path.dirname(filename))
+    if (!directory) {
+        return
+    }
+
+    if (!ethersyncClient) {
```
zormit, line 276: Just to reflect on this, as I'm not sure.

What happens if I (accidentally) open a file from a second Ethersync-enabled directory. Then we would not reconnect, but re-use the (wrong) ethersync client, bound to the first directory. Is it a problem? I think the edits would just not be tracked, right? Would it lead to any confusion?

daemon/integration-tests/src/actors.rs
```diff
@@ -159,15 +165,29 @@ fn rand_usize_inclusive(start: usize, end: usize) -> usize {
 }
 
 impl Neovim {
-    pub async fn new_ethersync_enabled(initial_content: &str) -> (Self, PathBuf) {
+    // The caller should store the TempDir, so that it is not garbage collected.
```
zormit, line 168: It's maybe a weird place, but what if we store it in the `Neovim` struct? Would it then be garbage collected at the correct time?

daemon/integration-tests/tests/vim-plugin.rs
```diff
@@ -5,116 +5,14 @@
...
 use tokio::time::{timeout, Duration};
-use tokio::{
-    io::{split, AsyncBufReadExt, AsyncWriteExt, BufReader, BufWriter},
-    net::UnixListener,
-    sync::mpsc,
-};
-
-use std::path::Path;
-
-struct MockSocket {
```
zormit, line 27: Nice refactoring to pull this out! :)

daemon/src/editor.rs
```diff
@@ -108,12 +68,13 @@ pub async fn spawn_socket_listener(
     document_handle: DocumentActorHandle,
 ) -> Result<()> {
     // Make sure the parent directory of the socket is only accessible by the current user.
-    if let Err(description) = is_user_readable_only(&socket_path) {
+    /*if let Err(description) = is_user_readable_only(&socket_path) {
```
zormit: You wrote in the PR:
> Maybe we should drop this permission requirement now? I think basically all other programs leave it up to the user to set permissions on files in their own directories like they want...

I have some "Bauchschmerzen" with this. We introduced this feature as part of an external security review. At least we should warn the user.

Though my thoughts go in the direction of owning the `.ethersync` directory: We are currently seeing at as flaw anyways that the user has to create `.ethersync` themselves, so maybe it's a good motivation to either create an `ethersync init` or to have a dialogue that asks the user if they want us to create `.ethersync`, if it doesn't exist when they call `ethersync share`.

vscode-plugin/src/extension.ts
blinry: Probably! It's like that on `main` as well, though. So at least not a regression... :P

But the proper way to do it would be to have a dictionary of daemon connections, keyed via their directories, probably? Same thing in the Vim plugin.

I'd favor documenting this in an issue (and maybe in the docs?), but I'd also be open to do it properly in this PR!

daemon/integration-tests/src/actors.rs
blinry: I think it would be! It'd need to be an `Option`, because we can also construct Neovims without a TempDir inside it (with `new()`). It *does* feel a bit weird to put it there. Maybe I'd prefer it like it is?

daemon/src/editor.rs
blinry: I think my main "program to compare to" was Git, which also just seems to create its files in `.git` with the user's default permissions (which could be rw-r--r--).

My thinking was also that currently, a "local attacker" (another user on the same system) could just modify the text files in the shared directory, instead of going through the daemon. So to me, it seems that the attacker can't really do "more harm" by connecting to the socket. (Assuming our sandbox works correctly.)

Are there other references we could look at? Software that creates keyfiles in a project directory?

Setting the permissions for the key file, on the other hand, seems good to force to rw-------. That's also what ssh does, for example.

Offering to create `.ethersync` ourselves seems good in any case :) Then we might just do it with proper permissions, I guess, that's true!

daemon/integration-tests/src/actors.rs
zormit: Seems fine. Until we fall into the trap to forget it :-D but I think it will just fail fast, so should be not a mean trap, but an easy one to get out of.

vscode-plugin/src/extension.ts
zormit: Yeah we don't have to fix it here, I just got the idea when reading the changes. I would say we can handle it when thinking about multi-project support (through one daemon) and just delay it until then?

zormit: #279 created

daemon/src/editor.rs
blinry: I added a prompt that offers to create `.ethersync` with rwx------ permissions, if it doesn't exist, and an explanation for how to set the permissions yourself, if it does.

daemon/src/main.rs
```diff
@@ -141,11 +144,22 @@ fn get_directory(directory: Option<PathBuf>) -> Result<PathBuf> {
...
         );
+
+        if ask(&format!(
+            "Create an {}/ directory to allow live collaboration?",
+            config::CONFIG_DIR
+        ))? {
+            sandbox::create_dir(&directory, &ethersync_dir)?;
+            info!("Created! Resuming launch.");
+        } else {
+            panic!("Aborting launch.");
```
zormit: This seems to be a bit harsh, to panic on a "no" (or anything else that isn't a yes). Maybe we can shutdown more gracefully and give the user more guidance on why it shut down?

daemon/src/main.rs
```diff
@@ -159,3 +173,17 @@ async fn wait_for_ctrl_c() {
         }
     }
 }
+
+fn ask(question: &str) -> Result<bool> {
+    print!("{} (y/n): ", question);
```
zormit: the default will be no, which I _think_ is symbolized by a capital letter? 
```suggestion
    print!("{} (y/N): ", question);
```

daemon/src/main.rs
```diff
@@ -159,3 +173,17 @@ async fn wait_for_ctrl_c() {
...
+    std::io::stdout().flush()?;
+    let mut lines = std::io::stdin().lines();
+    if let Some(Ok(line)) = lines.next() {
+        match line.as_str() {
```
zormit: Should we accept e.g. a `Y` as well? I would say so :)
```suggestion
        match line.to_lowercase().as_str() {
```

zormit (approved):
Looks good. Thanks for implementing the `.ethersync` creation. I've left some nitpicks on the usage of it, but we can also merge as it is :)

Also I really like the improvements to the docs and that they simplify with this :)

PR #283, opened 2025-06-24
Update more stuff :)

zormit:
Update Automerge to 1.0.0-beta.3 and constant_time_eq to 0.4.2


blinry:
Yeah, the Automerge update seems to break our fuzzer :/

It might be a performance issue! When I run the fuzzer locally on this branch, and limit it to 100 (instead of 500) edits per actor, it needs 18.81 seconds to pass.

On `main`, for 100 edits, it needs 4.7 seconds.

Maybe we can try again when we've removed the cursor storage? That *might* be the bottleneck? (We could also test that by not sending cursor positions!)

PR #284, opened 2025-06-25
Small fixes while browsing the code

zormit:
- **vim-plugin: fix: Replace deprecated nvim_err_writeln**
- **refactor: clippy: inline format args**


zormit:
`vim.api.nvim_echo({ { error_message } }, true, { err = true })` is only supported in nvim 0.11, but we want to support older versions as well. Version specific code isn't worth it here :)

Issue #44, opened 2024-07-04
Cursors should show immediately when buffer is opened


zormit:
The editor only gets ephemeral "moved" updates, but can't see the "past" yet. Would be nice, to get the latest cursor state from the daemon when opening a file.

PR #291, opened 2025-06-27
Ephemeral cursors, part 2

blinry:
This is based on #286, and adds some remaining bits and pieces:

- Cache cursor positions known to a peer, using the cursor ID and a (incrementing) sequence number.
- When receiving a cursor information from a peer, only do something if it's newer than the last known one.
- Send known cursor positions to editor when they join (#44).

I refactored the involved types a bit, thinking that neither the cursor ID (maybe another name would be better? client ID?) nor the sequence number really belong to the "payload" = the cursor selection information. But I'm not too sure/confident about this change.

### DIFF
--- a/daemon/src/daemon.rs
+++ b/daemon/src/daemon.rs
@@ -11,9 +11,9 @@ use crate::path::{AbsolutePath, RelativePath};
 use crate::peer;
 use crate::sandbox;
 use crate::types::{
-    ComponentMessage, CursorState, EditorProtocolMessageError, EditorProtocolMessageFromEditor,
-    EditorProtocolObject, FileTextDelta, JSONRPCFromEditor, JSONRPCResponse, PatchEffect,
-    TextDelta,
+    ComponentMessage, CursorId, CursorStateWithSequenceNumber, EditorProtocolMessageError,
+    EditorProtocolMessageFromEditor, EditorProtocolObject, FileTextDelta, JSONRPCFromEditor,
+    JSONRPCResponse, PatchEffect, TextDelta,
 };
 use crate::watcher::Watcher;
 use crate::watcher::WatcherEvent;
@@ -61,7 +61,7 @@ pub enum DocMessage {
     },
     NewEditorConnection(EditorId, EditorWriter),
     CloseEditorConnection(EditorId),
-    ReceiveEphemeral(CursorState),
+    ReceiveEphemeral(CursorStateWithSequenceNumber),
 }
 
 // TODO: Refactor this, to make use of sub-debug traits (default match).
@@ -88,8 +88,8 @@ impl fmt::Debug for DocMessage {
 type DocMessageSender = mpsc::Sender<DocMessage>;
 type DocChangedSender = broadcast::Sender<()>;
 type DocChangedReceiver = broadcast::Receiver<()>;
-type EphemeralMessageSender = broadcast::Sender<CursorState>;
-type EphemeralMessageReceiver = broadcast::Receiver<CursorState>;
+type EphemeralMessageSender = broadcast::Sender<CursorStateWithSequenceNumber>;
+type EphemeralMessageReceiver = broadcast::Receiver<CursorStateWithSequenceNumber>;
 
 /// This Actor is responsible for applying changes to the document asynchronously.
 ///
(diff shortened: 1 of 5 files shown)


daemon/src/daemon.rs
```diff
@@ -621,30 +651,63 @@ impl DocumentActor {
...
-            name: None,
-            file_path: RelativePath::new(""), // TODO: Fix by changing the "cursor" message?
-            ranges: vec![],
-        });
+    async fn maybe_delete_cursor_position(&mut self, cursor_id: &CursorId) {
+        let message = ComponentMessage::Cursor {
+            cursor_id: cursor_id.clone(),
+            cursor_state: CursorState {
+                name: None,
+                file_path: RelativePath::new(""),
```
zormit, line 711: You decided to drop the TODO, which seems fair. I don't think we'll adapt the protocol in the near future. We could still explain, what's happening here:
```suggestion
                file_path: RelativePath::new(""), // NOTE: The "cursor" message doesn't have a specific
                // deletion mechanism. We get around this by setting it to an empty file path,
                // which means it will disappear from any previous file path.

```

daemon/src/daemon.rs
```diff
@@ -621,30 +651,63 @@ impl DocumentActor {
...
+            )
+            .await;
+        }
+    }
+
+    async fn react_to_ephemeral_message(&mut self, new_ephemeral_message: EphemeralMessage) {
+        let cursor_id = new_ephemeral_message.cursor_id.clone();
+        if let Some(existing_state) = self.ephemeral_states.get_mut(&cursor_id) {
+            if new_ephemeral_message.sequence_number <= existing_state.sequence_number {
+                // We've already seen an older ephemeral message for this cursor_id.
```
zormit: Isn't it the other way around?
```suggestion
                // We've already seen a newer ephemeral message for this cursor_id, thus ignoring this older one.
```

daemon/src/daemon.rs
```diff
@@ -621,30 +651,63 @@ impl DocumentActor {
...
+        let _ = self
+            .ephemeral_message_tx
+            .send(new_ephemeral_message.clone());
+
+        // Broadcast to editors.
+        self.broadcast_to_editors(
+            None,
+            &ComponentMessage::Cursor {
+                cursor_id: new_ephemeral_message.cursor_id,
+                cursor_state: new_ephemeral_message.cursor_state.clone(),
```
zormit: Here we can just pass ownership - no need to clone anymore at the end.
```suggestion
                cursor_state: new_ephemeral_message.cursor_state,
```

Not sure it makes any difference or if the compiler optimizes that away anyways.

daemon/src/daemon.rs
```diff
@@ -99,6 +99,7 @@ pub struct DocumentActor {
     doc_changed_ping_tx: DocChangedSender,
     ephemeral_message_tx: EphemeralMessageSender,
     editor_connections: HashMap<EditorId, (EditorConnection, EditorWriter)>,
+    ephemeral_states: HashMap<CursorId, EphemeralMessage>,
```
zormit, line 102: Will we ever remove an entry? This is probably premature optimization, but if a long running daemon has lots of different peers connecting, this HashMap will grow indefinitely (in a very slow way, because we only store the _latest_ `EphemeralMessage`, so that's not a lot of data)

daemon/src/types.rs
zormit, line 1: When testing this with an older `doc` I realized that we have definitely created an incompatibility in the previous PR #286, by deleting the handling code of `states` [here](https://github.com/ethersync/ethersync/pull/286/files#diff-85aca4b73d07d66ef4e61ae50920b9e7d08812cb0f83d9611823f775cf7494d8L519).

```
thread 'tokio-runtime-worker' panicked at src/types.rs:128:21:
Failed to convert patch to delta: "Unexpected path in Automerge patch, expected it to begin with 'files'"
```

I think it would be nice, if older documents wouldn't crash, but I also understand that we don't want the code rot.

For now I would accept that it's failing and just advise users in the release notes that they need to delete their document. If this software becomes more popular we can transition a bit smoother.

zormit (approved):
Seems to be working fine, thank you :)

I've left some comments, some of them can be applied, but none of the issues are really blocking.

daemon/src/daemon.rs
blinry: We won't remove entries, and I agree that this might be nice! But I also agree that it's not urgent.

One thing we can easily do: Delete the state whenever a cursor disconnets, in `maybe_delete_cursor_position`.

daemon/src/daemon.rs
blinry: You're right!

daemon/src/daemon.rs
blinry: I like that.

daemon/src/types.rs
blinry: I started building a "compatibility" feature, but then realized that basically all users will *want* to delete their document (which contains many, many cursors...) :D So let's leave it as-is.

PR #292, opened 2025-06-30
vim: When pressing o on a line, edit should be placed in the new line

### LINKED ISSUE #288: Pressing o in Vim while someone else is editing that line
Opening a new line with `o` in Vim while someone else is editing that line can push that user into the next line. That's quite annoying.

We have this test:

```
assert_vim_input_yields_replacements("a\n", "o", vec![replace_ed((0, 1), (0, 1), "\n")]).await;
```

It might already help if this would result in `replace_ed((1, 0), (1, 0), "\n"))`.

blinry:
Instead of placing the edit after the previous one. This fixes #288.

Strategy: Add a new "fix" condition, where we check for newlines at the beginning of the replacement. Also, when 'eol' is on, pretend there are empty lines for compute_diff. This leads to prettier diffs overall. We need that differentiation because otherwise, for buffer content "x", when pressing o, we have no chance to insert in the next line (which we may do when 'eol' is on!)

### DIFF
--- a/daemon/integration-tests/tests/vim-plugin.rs
+++ b/daemon/integration-tests/tests/vim-plugin.rs
@@ -326,9 +326,6 @@ async fn vim_sends_correct_delta() {
 
     assert_vim_input_yields_replacements("a", "o", vec![replace_ed((0, 1), (0, 1), "\n")]).await;
 
-    assert_vim_input_yields_replacements("eins\ntwo", "jo", vec![replace_ed((1, 3), (1, 3), "\n")])
-        .await;
-
     assert_vim_input_yields_replacements(
         "eins\ntwo\n",
         "jo",
--- a/vim-plugin/lua/changetracker.lua
+++ b/vim-plugin/lua/changetracker.lua
@@ -46,7 +46,7 @@ function M.track_changes(buffer, callback)
             -- TODO: optimize with a cache
             local curr_lines = get_all_lines_respecting_eol(buffer)
 
-            -- Special case: When deleting the entire content, when of 'eol' is on, there
+            -- Special case: When deleting the entire content, when 'eol' is on, there
             -- will still be a "virtual line" after the current empty line: The file content will be "\n".
             -- So new_last_line should not be 0, but 1!
             if vim.bo[buffer].eol and #curr_lines == 2 and curr_lines[1] == "" and last_line == 1 then
@@ -121,6 +121,7 @@ function M.track_changes(buffer, callback)
                     end
                 end
             else
+                -- The range does not extend past the visible buffer lines (diff.range["end"].line < #prev_lines).
                 -- We might still want to make the delta prettier.
                 -- TODO: Integrate these cases in the above if branches somehow?
                 if
@@ -143,7 +144,6 @@ function M.track_changes(buffer, callback)
                 then
                     -- Range ends on the beginning of a line, but the replacement ends with a newline.
                     -- This newline is redundant, and leads to less-pretty diffs. Remove it.
-
                     diff.text = string.sub(diff.text, 1, -2)
                     diff.range["end"].line = diff.range["end"].line - 1
                     diff.range["end"].character = vim.fn.strchars(prev_lines[diff.range["end"].line + 1])


vim-plugin/lua/changetracker.lua
```diff
@@ -32,9 +43,17 @@ function M.track_changes(buffer, callback)
             -- Line counts that we get called with are zero-based.
             -- last_line and new_last_line are exclusive
 
-            debug({ first_line = first_line, last_line = last_line, new_last_line = new_last_line })
             -- TODO: optimize with a cache
-            local curr_lines = vim.api.nvim_buf_get_lines(buffer, 0, -1, true)
+            local curr_lines = get_all_lines_respecting_eol(buffer)
+
+            -- Special case: When deleting the entire content, when of 'eol' is on, there
```
zormit: "...when of..."? What was it supposed to say? Probably just "when"?
```suggestion
            -- Special case: When deleting the entire content, when 'eol' is on, there
```

vim-plugin/lua/changetracker.lua
```diff
@@ -101,17 +120,30 @@ function M.track_changes(buffer, callback)
...
+                end
+
                 if
                     diff.range["end"].character == 0
                     and string.sub(diff.text, -1) == "\n"
                     and diff.range["start"].line < diff.range["end"].line
                 then
                     -- Range ends on the beginning of a line, but the replacement ends with a newline.
                     -- This newline is redundant, and leads to less-pretty diffs. Remove it.
+
```
zormit: [nitpick] seems a bit inconsistent to add an empty line here? "Remove the newline!" =)
```suggestion
```

daemon/integration-tests/tests/vim-plugin.rs
```diff
@@ -332,16 +332,12 @@ async fn vim_sends_correct_delta() {
...
     )
     .await;
 
-    assert_vim_input_yields_replacements(
-        "eins\ntwo\nthree",
-        "jo",
-        vec![replace_ed((1, 3), (1, 3), "\n")],
-    )
-    .await;
+    assert_vim_input_yields_replacements("eins\ntwo", "jo", vec![replace_ed((1, 3), (1, 3), "\n")])
```
zormit, line 336: This is now the same as line 329. You think "eins\ntwo\nthree" would be redundant? Let's drop it, then!
```suggestion
```

daemon/integration-tests/tests/vim-plugin.rs
```diff
@@ -366,10 +362,18 @@ async fn vim_sends_correct_delta() {
     )
     .await;
 
-    // Visual on multiple lines
+    // Visual on multiple lines.
```
zormit, line 362: 🦅 :)

vim-plugin/lua/changetracker.lua
```diff
@@ -101,17 +120,30 @@ function M.track_changes(buffer, callback)
                         )
                     end
                 end
-            elseif diff.range["end"].line < #prev_lines then
-                -- The range ends before the line after the last visible buffer line.
```
zormit, line 105: I would find a comment helpful to keep track of what conditional branch we're in. Can we re-add something similar?

vim-plugin/lua/changetracker.lua
```diff
@@ -101,17 +120,30 @@ function M.track_changes(buffer, callback)
...
+            else
```
zormit, line 123: The condition isn't necessary? Can `diff.range["end"].line > #prev_lines` ever happen? (being a lazy reviewer: I'm just asking questions for now, maybe I'll try to answer it myself too, but only in the next step :D)

zormit (approved):
Seems legit! :) Let's-o fix-o it! ("it's-a me fixi-mo!" (and fixi-bry? 🤣))

I tested it and it fix-o things indeed, nice! Tbqh, I don't get the code in lines 130ff yet 🤯

PS: Curious, how VS Code behaves in this? Could this even be an issue there?

vim-plugin/lua/changetracker.lua
blinry: Like it, done!

vim-plugin/lua/changetracker.lua
blinry: I think that would be a bad, unfixable situation: The range would cover unknown content. Should we log a warning in this case?

zormit: I suppose it would warn us or crash somewhere else (probably rope error or something), so basically fine by me :) 😈

PR #293, opened 2025-06-30
Update pre-commit to newest version

zormit:
Super tiny unimportant tooling update. I noticed that I hadn't had `pre-commit` installed in my current setup. When I wanted to activate it I was nudged to update it. Here we are. Also a `pre-commit run --all-files` found some whitespace issues which I'll sneak in.


LICENSES/MPL-2.0.txt
```diff
@@ -35,7 +35,7 @@ Mozilla Public License Version 2.0
     means any form of the work other than Source Code Form.
 
 1.7. "Larger Work"
-    means a work that combines Covered Software with other material, in 
+    means a work that combines Covered Software with other material, in
```
blinry: I feel a bit uneasy modifying the license file, as it was downloaded by the `reuse` tool. It might check if it's the correct/unmodified file? The [original](https://www.mozilla.org/media/MPL/2.0/index.txt) also seems to have it.

Could we exclude the LICENSES directory?

blinry (changes requested):

PR #297, opened 2025-07-02
No document pre-baking

zormit:
After reading https://github.com/ethersync/ethersync/issues/69 in combination with https://github.com/ethersync/ethersync/pull/152#issuecomment-2311619197 I thought it might actually be worth to remove the hierarchy and see how far we can get with that.

It simplifies the code and we might have an easier time reasoning about initialization (hopefully).

The main disadvantage I see is that we can not synchronize any metadata other than file changes and patches.


blinry:
I like the simplification this brings! :) I'm now also hesitating a bit, for future extension reasons. Collecting things we might use other keys for:

- A "version" entry, that allows us to do :scream: schema migration :scream: (The version could also be derived from the structure, though)
- Metadata for a good Git integration, like storing which automerge version relates to which commit. This seems like the most important one to me.
- Metadata related to undo?!

For comparison, I looked at what https://patchwork.inkandswitch.com ended up with. Their directory format is like this:

```
"root": {
	"discussions": {}
	"tags": []
	"changeGroupSummaries": {}
	"versionControlMetadataUrl": "automerge:HZsAEA26d9eaUow1YxPEHhJqhEd"
	"title": "Untitled Folder"
	"docs": [
		0: {
			"url": "automerge:2NwxRP55NmeW8WCus3cgfaJcQXYZ"
			"type": "essay"
			"name": "Hehe"
		}
		1: {
			"url": "automerge:46uaHjjUkiPCzyCAF1XbBF9btdbZ"
			"type": "essay"
			"name": "Blubber"
		}
	]
	"@patchwork": {
		"type": "folder"
		"suggestedImportUrl": "@patchwork/folder"
	}
}
```

And per-document, it's like this:

```
"root": {
	"content": "# Hehe\n\n"
	"commentThreads": {}
	"discussions": {}
	"tags": []
	"changeGroupSummaries": {}
	"versionControlMetadataUrl": "automerge:JZQMvofj4JfoepDwnZoQYNrDVMs"
	"@patchwork": {
		"type": "essay"
		"suggestedImportUrl": "@patchwork/essay"
	}
}
```

(Don't really have a takeaway from this, except that they did need a lot of structure.)

My feeling is that we will need sub-keys at some point for the Git integration, and then also the baker. (If we don't switch to [Loro](https://loro.dev) or something.) But we could return to it if we need it, and have simplicity for now. I don't really have a strong opinion. The functionality is the same. Document compatibility doesn't seem very important right now. What do you think?

blinry:
We discussed this in-person, and decided to leave the structure as-is for now :)

Issue #97, opened 2024-08-06
Rework logic for which peer initializes the CRDT


blinry:
Currently, the logic is a bit involved; a host initiates the "files" and "states" maps in the CRDT if there's no persistence file, and if they are the host.

In pairing use cases, we often want to "start from scratch", especially after switching the branch. There should be an easier way to do that.

blinry:
This is related to #69.

zormit:
> They shouldn't need to discuss who's the sharing peer. Both should just be able to use the same command. Maybe `ethersync daemon`? `ethersync connect`? (Is there other software that has terms for this?)

I would think of it as a sort of "friend list" or "known devices". I've seen this in software like Steam, TeamViewer, Parsec, Syncthing. I would take Syncthing as the most fitting example and inspiration, because it has the concept of a device but also the concept of a shared folder, and that is in a way independent. Reminds me of our (partially offline) discussion of #270.

blinry:
Just commenting that I like the UI of having `share` and `join` subcommands, to make the intent of "I want to invite a new person" and "I want to connect to a new peer" explicit. :)

For cases where one wants neither of those things, we could re-introduce `ethersync daemon`?

PR #298, opened 2025-07-03
Detect git remotes to infer pairing use-case

zormit:
Introduces a heuristics which tries to guess when the user is in the "pair-programming use-case", in which persistence is not desired.

Based on the existence of a configured Git remote location, the persistence is turned off. The reasoning is based on https://github.com/ethersync/ethersync/issues/97#issuecomment-3031583472 and assumes that in the note-taking case you wouldn't have a Git remote (which is a wild guess, but it fits to how we currently do things).

### DIFF
--- a/daemon/src/config.rs
+++ b/daemon/src/config.rs
@@ -121,7 +121,7 @@ pub fn store_peer_in_config(directory: &Path, config_file: &Path, peer: &str) ->
 }
 
 pub fn has_git_remote(path: &Path) -> bool {
-    if let Ok(repo) = git2::Repository::open(path) {
+    if let Ok(repo) = git2::Repository::discover(path) {
         if let Ok(remotes) = repo.remotes() {
             return !remotes.is_empty();
         }


daemon/src/daemon.rs
```diff
@@ -115,14 +115,15 @@ impl DocumentActor {
         base_dir: PathBuf,
         init: bool,
         is_host: bool,
+        persist: bool,
```
zormit, line 118: With the 7th constructor parameter it could be worth thinking about, whether we can somehow cluster the parameters into structs.

daemon/src/main.rs
```diff
@@ -192,3 +200,12 @@ fn ask(question: &str) -> Result<bool> {
         bail!("Failed to read answer");
     }
 }
+
+fn has_git_remote(path: &Path) -> bool {
```
zormit: It's not necessarily config, but I could also see this in the `config.rs` module:
- it's about inferring the "configured" behaviour of the application
- we could, later, think about introducing a variable to influence/overwrite the inferred use-case
- if we do this, `AppConfig` could also get a `persist` field

daemon/src/daemon.rs
```diff
@@ -115,14 +115,15 @@ impl DocumentActor {
...
     ) -> Self {
         // If there is a persisted version in base_dir/.ethersync/doc, load it.
         // TODO: Pull out ".ethersync" string into a constant.
         let persistence_file = base_dir.join(".ethersync/doc");
         let persistence_file_exists = sandbox::exists(&base_dir, &persistence_file)
             .expect("Could not check for the existence of the persistence file");
 
-        let crdt_doc = if persistence_file_exists && !init {
+        let crdt_doc = if persistence_file_exists && !init && persist {
```
zormit: Tiny refactoring suggestion could be to pull this out to it's own variable. Just because it getting longer and it'd be more "speaking" that way without having to look into the implementation of the branch.
```suggestion
        let load_crdt_doc = persistence_file_exists && !init && persist;
        let crdt_doc = if load_crdt_doc {
```

daemon/src/main.rs
```diff
@@ -192,3 +200,12 @@ fn ask(question: &str) -> Result<bool> {
...
+    if let Ok(repo) = git2::Repository::open(path) {
```
zormit: I realized that this heuristic doesn't work if someone wants to pair program on a subdirectory of a git repo. `Repository::open` will not "traverse up", like `git` itself does (see also [`Repository::discover`](https://docs.rs/git2/latest/git2/struct.Repository.html#method.discover)), but I think that's good. But sometimes you might have a huge monorepo and only want to pair-program on a sub-tree. In our case, we could also easily "just" pair on `daemon` or so sometimes.

The heuristics will assume it should persist in this case.

daemon/src/daemon.rs
blinry: Right! Especially as `init`, `is_host`, and `persist` are just passed through from the DaemonActorHandle and the Daemon. We could rename `AppConfig` into `PeerConfig` and these three into `CRDTConfig` or `PersistanceConfig`?

daemon/src/main.rs
blinry: Hmm... my feeling is that in the sub-tree-pairing case I would also like Ethersync to not persist, so that the document histories are "ephemeral".

I once had an unusual setup where my home directory *was* a Git directory (for dotfiles), so there that might be a bit weird.

But maybe I'd slightly prefer switching to `discover()`? So we'd check "is our directory part of a Git repo with a remote"?

daemon/src/daemon.rs
blinry: Or switch to a [Builder pattern](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html) at some level? It feels wrong that adding a parameter requires adaptions in the tests...

daemon/src/main.rs
zormit: Yes, I was thinking about this "the whole `HOME` is a repo" indeed.
 
It will also cause a playground subdirectory (for me: somewhere _within_ the repo, but ignored through `.git/info/exclude`), for which we then have to find a different approach if we want to test things about persistence.

If we do `discover()`, I'm keen to already introduce a config variable to overwrite that behaviour, as mentioned [here](https://github.com/ethersync/ethersync/pull/298#discussion_r2183662650).

edit: Just realizing, that the workaround for the playground is to initialize an empty git repo (without remotes) in there, if we want to test persistence. That's alright for me, so I'll keep away from the config variable until this feature stabilized.

daemon/src/daemon.rs
zormit: I checked the Builder pattern and my thought is, that it's not exactly what we need here.

[Motivation](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html#motivation):
> Useful when you would otherwise require many constructors or where construction has side effects.

We don't have many different constructors. So using the builder will, for every `bool` constructor parameter probably result in a conditional that does something like:
```
let builder = ...
if toggle {
    builder.toggle();
}
```

Which seems rather vocal compared to just passing in the `toggle`. But I agree that it's a bummer that we're breaking the test cases that way.

I can still give it a try, just out of curiosity if it would be as bad as I think it is. At least the test can then live off a reasonable default.

zormit: The experiment of that kind of refactor can be found in [`detect-git-remote-builder`](https://github.com/ethersync/ethersync/tree/detect-git-remote-builder). I'm sure it could be done differently, maybe more elegant...

blinry: Hmm, you're right in that it doesn't seem to add much clarity. Maybe having a `DaemonBuilder` (one level higher) would be a bit better – there, we could then "configure" the Daemon step by step, depending on various factors, and then launch it. But I'll first check out #299, as it sounds like it might remove the arguments entirely?

zormit: I suggest we merge this without the refactor and think about the Builder pattern some time later, if needed? #299 builds on top of the idea of this PR.

blinry: (We decided to leave this un-fixed in this PR.)

blinry:
Merged! Noting two things still feel a bit weird:

1. If someone with a remote (A) shares with someone without a remote (B), A will re-initialize, but because B retains its history, both will get conflicts for all files. Could be solved by #299 and/or #97.
2. Sharing from a directory with a remote will always log all files as initialized. Let's see if that's maybe a bit much/a bit spammy.

PR #299, opened 2025-07-07
Detect conflicts

zormit:


daemon/src/daemon.rs
```diff
@@ -148,11 +146,7 @@ impl DocumentActor {
             save_fully: true,
         };
 
-        if persistence_file_exists && persist {
-            s.read_current_content_from_dir(init);
-        } else if is_host {
-            s.read_current_content_from_dir(true);
-        }
```
blinry, line 155: Getting rid of all of this would be lovely indeed! :)

daemon/src/document.rs
```diff
@@ -174,6 +174,33 @@ impl Document {
...
+                } else {
+                    "".into()
+                }
+            });
+            let t1 = texts.next().unwrap();
+            let t2 = texts.next().unwrap();
+            let conflicting = t1 != t2;
+            conflicting
+        } else {
+            panic!("more than 2, whatsthat? :-O");
```
blinry: See the [fuzzer](https://github.com/ethersync/ethersync/blob/main/daemon/integration-tests/tests/fuzzer.rs#L166-L175) for how I once implemented checking that any number of things is the same.

daemon/src/daemon.rs
```diff
@@ -232,6 +226,13 @@ impl DocumentActor {
                                 warn!("Failed to remove file {file_path}: {err}");
                             });
                         }
+                        PatchEffect::Conflict(file_path) => {
+                            if self.crdt_doc.has_true_conflict(&file_path) {
```
blinry, line 222: How sure can we be that the conflicts will still be present in the document here? (I know too little about conflicts to judge.) Is it possible that, because we might have applied several patches at this point already, the conflict is already gone? If not, will conflicts accumulate?

PR #300, opened 2025-07-07
Fix logo heights

zormit:
It seems that the `height` attribute is ignored on GitHub Markdown rendering. Doing some ratio maths and changing it to `width` seems to help.


blinry:
Yeah, CSS heights seem to work in all views. Feel free to squash-and-merge this!

PR #328, opened 2025-07-23
Warn if .ethersync is within a Git repository, but not ignored

### LINKED ISSUE #327: Make it unlikelier for people to check in their `.ethersync/key` file
It seems important that people follow our [Git setup](https://ethersync.github.io/ethersync/git-integration.html#ignoring-ethersync-directories) instructions, or they might accidentally check in their `.ethersync/key` file.

One solution could be to check that `.ethersync/` is in `~/.config/git/ignore` when starting on a directory that is a Git repository.

zormit:
Fixes #327.

### DIFF
--- a/daemon/src/config.rs
+++ b/daemon/src/config.rs
@@ -131,6 +131,16 @@ pub fn has_git_remote(path: &Path) -> bool {
     false
 }
 
+pub fn ethersync_directory_is_ignored(path: &Path) -> bool {
+    if let Ok(repo) = git2::Repository::discover(path) {
+        let ethersync_dir = path.join(CONFIG_DIR);
+        return repo
+            .is_path_ignored(ethersync_dir)
+            .expect("Should have been able to determine ignore state for {path}");
+    }
+    false
+}
+
 pub fn get_username(base_dir: &Path) -> Option<String> {
     local_git_username(base_dir)
         .or_else(|_| global_git_username())
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -86,14 +86,19 @@ async fn main() -> Result<()> {
         .join(config::CONFIG_DIR)
         .join(config::DEFAULT_SOCKET_NAME);
 
-    let persist = !config::has_git_remote(&directory);
-
     match cli.command {
         Commands::Share { .. } | Commands::Join { .. } => {
+            let persist = !config::has_git_remote(&directory);
             if !persist {
                 info!(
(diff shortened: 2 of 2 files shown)


daemon/src/config.rs
```diff
@@ -131,6 +131,16 @@ pub fn has_git_remote(path: &Path) -> bool {
     false
 }
 
+pub fn ethersync_directory_is_ignored(path: &Path) -> bool {
+    if let Ok(repo) = git2::Repository::discover(path) {
```
blinry: I'd suggest pulling out the repository discovery into a function, which is then also used for the "is this pairing use-case" detection, **because** :P that way, if we ever change the logic on how the repository works, it translates over to both places.

daemon/src/main.rs
```diff
@@ -86,16 +86,21 @@ async fn main() -> Result<()> {
...
             if !persist {
                 info!(
                     "Detected a Git remote: Assuming a pair-programming use-case and starting a new history."
                 );
             }
 
+            if !config::ethersync_directory_is_ignored(&directory) {
+                warn!(".ethersync should be ignored in your Git repo, as it contains sensitive secrets. See\n\
+                \thttps://ethersync.github.io/ethersync/git-integration.html#ignoring-ethersync-directories\n\
+                on how to do that.");
```
blinry: I'd like it if Ethersync was more helpful, more secure-by-default and more self-contained here:

- It should give all information required to make the desired change, without referring to external documentation. The documentation could then be deleted! :3 That way, neither the URL nor the documentation will get "out of date" in the future.
    - Ethersync could even ask: "Add '.ethersync/' to your ~/.config/git/ignore? (y/n)", and then make that change itself. Many users will want to make that change, and if we could help them, that would be great! (Could also be a later step.)
- I think Ethersync should actually refuse to start without .ethersync being ignored (if in a Git repo), going with our theme of security-by-default/enforcing safe configuration.

daemon/src/config.rs
```diff
@@ -131,6 +131,16 @@ pub fn has_git_remote(path: &Path) -> bool {
...
+        let ethersync_dir = path.join(CONFIG_DIR);
+        return repo
+            .is_path_ignored(ethersync_dir)
+            .expect("Should have been able to determine ignore state for {path}");
+    }
+    false
```
blinry, line 141: Another thing: If there's no Git repo, I'd say no action is needed. So either rename the function to something like `..._should_be_ignored_but_isnt` or call it in a context when we've detected a Git repo?

blinry (approved):

PR #329, opened 2025-07-27
ci: Build binaries for aarch64-unknown-linux-musl (for Android)

blinry:
(To speed up installation process on Android.)

I verified that the build works [here](https://github.com/ethersync/ethersync/actions/runs/16551213849/job/46806313187#step:3:701).


zormit (approved):
lets do it 📱 droidsync

PR #331, opened 2025-07-28
Polish a couple of things in the README & book

blinry:
Based on our latest user test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 
 Ethersync enables real-time collaborative editing of local text files. You can use it for pair programming or note-taking, for example. It's the missing real-time complement to Git!
 
-![](https://files.blinry.org/ethersync-demo.gif)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -91,7 +91,7 @@ $ ethersync share
 Peer connected: adfa90edd932732ddf242f24dc2dcd6156779e69966d432fcb3b9fe3ae9831ab
 ```
 
-In another directory (this can be on another computer!):
+Another person, in a separate directory (also works on the same computer):
```
zormit, line 93: This is still a bit confusing. So we have three things to convey:
- The usual case is to connect with another person on their computer
- It's also possible to connect with yourself on the same computer
- It has to be a different directory, of course different computer fulfills that automatically

How to bring this into one tiny sentence?
```suggestion
In another directory (usually a different person & computer, but it also works on the same computer):
```

zormit (approved):
Cool! :) (left one suggestion for improvement)

README.md
zormit: after discussing, I like the existing solution :)

zormit (approved):

PR #332, opened 2025-07-28
Add "content" field to the "open" message

### LINKED ISSUE #307: Possible race condition when opening files
## Problem

When opening a file in an editor, it reads the content from disk. It seems possible that this happens before the daemon has finished writing an updated version of the file, so now the buffer content would be wrong.

## To reproduce

Run

```lua
:lua t = vim.uv.new_timer(); t:start(0, 1, vim.schedule_wrap(function() vim.api.nvim_buf_set_text(0, 0, 0, 0, 0, {"x"}) end))
```

in one file (adds an "x" every millisecond), then open the shared version of the same file. Close the first nvim with `ZQ`. Now compare: The second nvim will probably have less x's. (For me, for example 1204 vs 1194 x's.)

## Possible solution

Add a content field to the "open" messages, where the editor reports its current content. If it's not what the daemon expects, it can send a delta back to bring it up to date.

blinry:
The "content" represents the editor buffers' starting content. The daemon will send edits back until it matches the canonical content.

This fixes #307.


blinry:
After this PR, the "reproduction steps" in #307 don't lead to the bug anymore.

However, I could've introduced other subtle bugs? Let's check this carefully.

This was not a very pleasant change to make. I have the feeling that it should be possible to refactor this code very nicely, by introducing a "Component" trait. Both the "CRDT document", as well as the editor connections would then implement it. I thought about it a bit, but it seems like a very big change to make. Maybe let's do a "Refactor the core" sprint soon?

PR #336, opened 2025-07-30
Recan files after watcher events were triggered

### LINKED ISSUE #180: Running `cargo init` doesn't create `src/main.rs` in the CRDT
We observed this behavior on Linux.

`Cargo.toml` *is* being created in the CRDT.

The watcher might be missing file events? (For possible solution ideas, see https://github.com/ethersync/ethersync/pull/133#issuecomment-2340272812)

blinry:
This is so that we won't miss any file changes - the watcher isn't necessarily exhaustive.

Fixes #180.

### DIFF
--- a/daemon/src/daemon.rs
+++ b/daemon/src/daemon.rs
@@ -574,14 +574,16 @@ impl DocumentActor {
                         let relative_file_path =
                             RelativePath::try_from_path(&self.base_dir, file_path)
                                 .expect("Walked file path should be within base directory");
-                        if let Ok(text) = String::from_utf8(bytes) {
-                            if init {
-                                self.crdt_doc.initialize_text(&text, &relative_file_path);
+                        if self.owns(&relative_file_path) {
+                            if let Ok(text) = String::from_utf8(bytes) {
+                                if init {
+                                    self.crdt_doc.initialize_text(&text, &relative_file_path);
+                                } else {
+                                    self.crdt_doc.update_text(&text, &relative_file_path);
+                                }
                             } else {
-                                self.crdt_doc.update_text(&text, &relative_file_path);
+                                warn!("Ignoring non-UTF-8 file {relative_file_path}",)
                             }
-                        } else {
-                            warn!("Ignoring non-UTF-8 file {relative_file_path}",)
                         }
                     }
                     Err(e) => {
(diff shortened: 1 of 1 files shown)


daemon/src/daemon.rs
```diff
@@ -872,13 +877,45 @@ impl Daemon {
...
+
+                        debounce_timer.as_mut().reset(tokio::time::Instant::now() + debounce_duration);
+                        rescan_required = true;
+                    } else {
+                        // Watcher terminated. Seems we're shutting down.
+                        return;
+                    }
+                }
+
+                _ = &mut debounce_timer, if rescan_required => {
```
zormit, line 918: This looks like a confusing bit of syntax 😬 What I'd assume it does given the context of the PR is to catch the event that the `debounce_timer` ran out, guarded by `rescan_required`. So `debounce_timer` is a `Future` and this branch is only activated when it has an actual result. Am I right? :)

daemon/src/daemon.rs
```diff
@@ -872,13 +883,45 @@ impl Daemon {
...
     tokio::spawn(async move {
-        while let Some(watcher_event) = watcher.next().await {
-            document_handle
-                .send_message(DocMessage::FromWatcher(watcher_event))
-                .await;
+        let debounce_duration = Duration::from_millis(100);
+
+        let debounce_timer = tokio::time::sleep(debounce_duration);
+        // Sleep does not implement the Unpin trait, so in order to use it with select!, we have to
+        // pin it first (according to the documentation).
```
zormit:
```suggestion
        // pin it first (according to the documentation https://docs.rs/tokio/latest/tokio/time/struct.Sleep.html).
```
What do you think about linking the documentation? That would make it for me smoother to read if I get back to this later... Downside is, the link might become outdated.

daemon/src/daemon.rs
zormit: I think I'm just not used to the syntax of `select!`... It seems I understood it correctly.

Btw, the [documentation](https://docs.rs/tokio/latest/tokio/time/struct.Sleep.html) uses a different construct:
```rust
() = &mut sleep => {
```
which seems to be equivalent. Both of these comments are just me processing the code while reading it, no action required :)

daemon/src/daemon.rs
```diff
@@ -569,26 +574,32 @@ impl DocumentActor {
                         let relative_file_path =
                             RelativePath::try_from_path(&self.base_dir, file_path)
                                 .expect("Walked file path should be within base directory");
-                        if let Ok(text) = String::from_utf8(bytes) {
-                            if init {
-                                self.crdt_doc.initialize_text(&text, &relative_file_path);
+                        if self.owns(&relative_file_path) {
```
zormit, line 577: Note to any other reader: The diff is confusing, here we just insert this line around the unchanged `if let Ok(text) ...` block.

zormit (approved):
Seems fine to me :) I could not confirm if it solves the bug as I'm not on a Linux system right now, but I'm sure rescanning (with this debounce timer) makes things better :)

daemon/src/daemon.rs
blinry: I also don't really like the complicated `,if` syntax, but I didn't have better ideas on how to construct the debounce timer here…

daemon/src/daemon.rs
blinry: I like linking it! :)

PR #337, opened 2025-07-30
refactor(nvim): insert "ethersync" to each module

aileot:
not to pollute lua package.loaded namespaces with general terms.


blinry:
Ah, I agree that this is much tidier! Thanks for the PR! :)

PR #341, opened 2025-08-02
Prevent accidentally starting multiple daemons on the same project

### LINKED ISSUE #324: Prevent multiple daemons starting for the same directory
Daemon shutdown could delete the socket, and a booting daemon checks if it exists?

zormit:
Refuse to start if socket exists, but also remove it on shutdown.
That way the socket itself is a kind of token to detect a running daemon.

Fixes #324.

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -33,7 +33,7 @@ struct Cli {
 enum Commands {
     /// Share a directory with a new peer.
     Share {
-        /// Re-initialize the history of the shared project. You will loose previous history.
+        /// Re-initialize the history of the shared directory. You will loose previous history.
         #[arg(long)]
         init: bool,
         /// Do not generate a join code. To prevent unintended sharing or simply if you want to
@@ -44,7 +44,7 @@ enum Commands {
         #[arg(long)]
         show_secret_address: bool,
     },
-    /// Join a shared project via join code.
+    /// Join a shared directory via a join code, or connect to the most recent one.
     Join {
         /// Specify to connect to a new peer. Otherwise, try to connect to the most recent peer.
         join_code: Option<String>,
--- a/daemon/integration-tests/tests/fuzzer.rs
+++ b/daemon/integration-tests/tests/fuzzer.rs
@@ -29,7 +29,7 @@ async fn perform_random_edits(actor: &mut (impl Actor + ?Sized)) {
     }
 }
 
-fn initialize_project() -> (temp_dir::TempDir, PathBuf, PathBuf) {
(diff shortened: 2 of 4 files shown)


blinry:
Tried a couple of edge conditions:

The cleanup is not done if the `ethersync` process is killed.

- [ ] Maybe we could even listen for the (nice) kill events, and shutdown more "cleanly"?

I wanted to see what happens when the daemon panics, and added a `panic!("boo")` in a suitable location. The socket was still there. According to the [Rust book](https://doc.rust-lang.org/reference/panic.html), `drop` is called for objects when unwinding the panic, but that probably doesn't happen if sub-tasks panic? I'd be fine with not improving this, as each daemon crash is a bug anyway, and should be fixed by us.

Two more things:

- [ ] Could we set a better `base_dir` for the new `remove_file` instead of "/"?
- [ ] To be more friendly, should we ask a (y/n) question offering the user to still start (if they're sure that no other daemon is running)?

daemon/src/main.rs
```diff
@@ -185,26 +183,15 @@ fn get_directory(directory: Option<PathBuf>) -> Result<PathBuf> {
...
 }
 
-async fn wait_for_ctrl_c() {
-    match signal::ctrl_c().await {
-        Ok(()) => {}
-        Err(err) => {
-            eprintln!("Unable to listen for shutdown signal: {err}");
-            // still shut down.
+async fn wait_for_shutdown() {
+    let mut signal_terminate = signal::unix::signal(signal::unix::SignalKind::terminate())
```
zormit, line 187: If we later would want to add Windows support (e.g. #258), we could handle a bunch more signals https://stackoverflow.com/a/77591939

zormit:
I think the reason we're not properly running the destructor is because we call

https://github.com/ethersync/ethersync/blob/1aa64396f51db543e1d32c5931865e2ccd24f19b/daemon/src/main.rs#L68

Which according to [this](https://doc.rust-lang.org/reference/destructors.html#process-termination-without-unwinding) skips the whole unwinding.

I tried to implement proper unwinding with `catch_unwind` but decided it's too complex for now. If a crash happens, at least the user can decide that the leftover socket will be removed for them.

blinry:
I tried it, and it all works like it should! :sparkles: I tried to clarify the description of what an existing socket means a bit.

And I noticed that some subcommand descriptions used "shared directory", and some "shared project". I unified them all to "shared directory", what do you think? If this seems controversial, feel free to remove that commit again & merge :D

blinry (approved):

PR #352, opened 2025-08-13
fix symlink-related crash for MacOS file-watcher event

zormit:
I now get this debug message instead of the crash when following the steps described [here](https://github.com/ethersync/ethersync/issues/194#issuecomment-3145589270) on MacOS.

```
10:57:57Z DEBUG ThreadId(12) ethersync::watcher: Ignoring modification of '/[...]/ethersync/daemon/playground/.direnv/flake-inputs/1mam1ng960ypaa55kwldgx2v63v285vh-source' because of an error: File path /nix/store/1mam1ng960ypaa55kwldgx2v63v285vh-source is not inside the base directory /[...]/ethersync/ethersync/daemon/playground
```

### DIFF
--- a/daemon/src/watcher.rs
+++ b/daemon/src/watcher.rs
@@ -104,14 +104,24 @@ impl Watcher {
                     notify::event::RenameMode::Any,
                 )) => {
                     assert!(event.paths.len() == 1);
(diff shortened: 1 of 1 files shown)


blinry (approved):
Thanks for splitting the refactors into commits so nicely, that made it very pleasant to review! Merging.

Issue #58, opened 2024-07-18
Implement a "follow mode" in Neovim


blinry:
I'd like to be able to say "I want to lean back and watch this user, wherever they go"! :)

PR #363, opened 2025-08-22
nvim: follow/unfollow participants

MichaelBitard:
This is a first raw implementation of the follow mode, works on my machine :tm: (I'll try it with a coworker monday probably).

#58 

I don't mind changing a lot of stuff (rename, code convention,...) if you want.

Anyway, I can't wait for my coworker to come back from holidays to try ethersync in nvim!!!! 

Awesome project!

## Questions

* I'm not really satisfied with the unfollow, I'd like to automatically unfollow when I type Esc, but I don't know how to do that, is it possible?
* In the readme, I updated the mappings examples but I think the keys I used are used by other plugins (I use Lazyvim and f and u are already mapped), which key combination should we set?

### DIFF
--- a/nvim-plugin/README.md
+++ b/nvim-plugin/README.md
@@ -30,7 +30,11 @@ Usually, you will add the string `"ethersync/ethersync-nvim"` to your plugin man
 ```lua
 {
   "ethersync/ethersync-nvim",
-  keys = { { "<leader>j", "<cmd>EthersyncJumpToCursor<cr>" } },
+  keys = { 
+    { "<leader>ej", "<cmd>EthersyncJumpToCursor<cr>" },
+    { "<leader>ef", "<cmd>EthersyncFollow<cr>" },
+    { "<esc>", mode = { "n" }, "<cmd>EthersyncUnfollow<cr>" },
+  },
   lazy = false,
 }
 ```
@@ -41,7 +45,9 @@ Usually, you will add the string `"ethersync/ethersync-nvim"` to your plugin man
 {
   "ethersync/ethersync-nvim",
   config = function()
-    vim.keymap.set('n', '<leader>j', '<cmd>EthersyncJumpToCursor<cr>')
+    vim.keymap.set('n', '<leader>ej', '<cmd>EthersyncJumpToCursor<cr>')
+    vim.keymap.set('n', '<leader>ef', '<cmd>EthersyncFollow<cr>')
+    vim.keymap.set('n', '<esc>', '<cmd>EthersyncUnfollow<cr>')
   end
 }
 ```
@@ -60,4 +66,4 @@ To confirm that the plugin is installed, try running the `:EthersyncInfo` comman
 
 ## Tips
 
-We recommend creating a mapping for the `:EthersyncJumpToCursor` command (for example, `<Leader>j`, which jumps to another user's cursor.
+We recommend creating mappings for the `:EthersyncJumpToCursor`, `:EthersyncFollow` and `EthersyncUnfollow` command (for example, `<Leader>ej`, which jumps to another user's cursor.
--- a/nvim-plugin/lua/ethersync/cursor.lua
+++ b/nvim-plugin/lua/ethersync/cursor.lua
@@ -1,5 +1,6 @@
 -- SPDX-FileCopyrightText: 2024 blinry <mail@blinry.org>
 -- SPDX-FileCopyrightText: 2024 zormit <nt4u@kpvn.de>
+-- SPDX-FileCopyrightText: 2025 mbitard <code@bitard.fr>
 --
 -- SPDX-License-Identifier: AGPL-3.0-or-later
 
@@ -18,6 +19,7 @@ local user_cursors = {}
 local cursor_namespace = vim.api.nvim_create_namespace("Ethersync")
 local offset_encoding = "utf-32"
 local cursor_timeout_ms = 300 * 1000
+local following_user_id = nil
 
 local function show_cursor_information(name, cursor)
     return name .. " @ " .. vim.uri_to_fname(cursor.uri) .. ":" .. cursor.range.start.line + 1
@@ -36,12 +38,23 @@ local function is_forward(start_row, end_row, start_col, end_col)
     return (start_row < end_row) or (start_row == end_row and start_col <= end_col)
 end
 
-local function jump(location)
-    -- In Neovim 0.11, jump_to_location was deprecated.
-    if vim.version().api_level < 13 then
-        vim.lsp.util.jump_to_location(location, offset_encoding, true)
+local function jump_to_user_id(user_id)
+    local cursors = user_cursors[user_id].cursors
+    local cursor = cursors[#cursors]
+    if cursor == nil then
+        following_user_id = nil
     else
-        vim.lsp.util.show_document(location, offset_encoding)
+        local location = {
+            targetUri = cursor.uri,
+            targetRange = cursor.range,
+            targetSelectionRange = cursor.range,
+        }
+        -- In Neovim 0.11, jump_to_location was deprecated.
+        if vim.version().api_level < 13 then
+            vim.lsp.util.jump_to_location(location, offset_encoding, true)
+        else
+            vim.lsp.util.show_document(location, offset_encoding)
+        end
     end
 end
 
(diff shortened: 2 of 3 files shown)


nvim-plugin/lua/ethersync/cursor.lua
```diff
@@ -311,7 +398,7 @@ function M.list_cursors()
         for _, data in pairs(user_cursors) do
             local name = data.name
             local cursors = data.cursors
-            message = message .. name .. ":"
+            message = message .. (name or "") .. ":"
```
MichaelBitard, line 350: This is a bugfix, when a user enter, then leave, then reenter, it creates a new entry and the old one has a nil name, causing an error if you call :EthersyncInfo

MichaelBitard:
The default change view is quite unreadable around the set_cursor method, don't forget to check "hide whitespace", I just removed the return, put the rest in an else in order to be able to follow in every condition

nvim-plugin/README.md
```diff
@@ -30,7 +30,11 @@ Usually, you will add the string `"ethersync/ethersync-nvim"` to your plugin man
 ```lua
 {
   "ethersync/ethersync-nvim",
-  keys = { { "<leader>j", "<cmd>EthersyncJumpToCursor<cr>" } },
+  keys = { 
+    { "<leader>j", "<cmd>EthersyncJumpToCursor<cr>" },
+    { "<leader>f", "<cmd>EthersyncFollow<cr>" },
+    { "<leader>u", "<cmd>EthersyncStopFollow<cr>" },
```
blinry: Now that we have multiple of these, it feels proper to me to cluster them under a common prefix. Maybe `<leader>e`? Then jump would be `<leader>ej`, for example. That might also help avoid clashes with existing mappings.

nvim-plugin/plugin/ethersync.lua
```diff
@@ -279,3 +279,5 @@ vim.api.nvim_create_autocmd("FileChangedShell", { callback = function() end })
 
 vim.api.nvim_create_user_command("EthersyncInfo", print_info, {})
 vim.api.nvim_create_user_command("EthersyncJumpToCursor", cursor.jump_to_cursor, {})
+vim.api.nvim_create_user_command("EthersyncFollow", cursor.follow_cursor, {})
+vim.api.nvim_create_user_command("EthersyncStopFollow", cursor.stop_follow, {})
```
blinry: Our intuitive feeling of what the "stop follow" command should be called was `:EthersyncUnfollow`, which would also fit the `<leader>u` mapping. What do you think? (zormit and I are both not native English speakers, so our feeling might be off!)

nvim-plugin/lua/ethersync/cursor.lua
```diff
@@ -45,6 +46,17 @@ local function jump(location)
     end
 end
 
+local function follow(user_id)
```
blinry: To make it clearer what this function does, we could rename it to `jump_to_userid`.

And while we're at it, we could rename the existing `jump` function to `jump_to_location`.

nvim-plugin/lua/ethersync/cursor.lua
```diff
@@ -233,6 +247,79 @@ function M.track_cursor(bufnr, callback)
     vim.api.nvim_exec_autocmds("CursorMoved", {})
 end
 
+function M.stop_follow()
+    following_userid = nil
+end
+
+function M.follow_cursor()
```
blinry: The code in this function is similar to the one in `M.jump_to_cursor`. It would be great to have a re-usable method that allows selecting cursors, avoiding this duplication!

It could be called `pick_cursor_menu`, and take a callback that's to be called when a cursor is selected. That callback should probably receive the selected userid, as well as the selected specific cursor data structure.

The message "No user to follow" could be generalized in that method to something like "No other cursors available."?

nvim-plugin/lua/ethersync/cursor.lua
```diff
@@ -233,6 +247,79 @@ function M.track_cursor(bufnr, callback)
...
+            border = "single",
+        }
+        local win = vim.api.nvim_open_win(buf, true, opts)
+
+        vim.api.nvim_buf_set_keymap(buf, "n", "<CR>", "", {
+            callback = function()
+                local line_number = vim.fn.line(".")
+                local location = locations[line_number]
+                vim.api.nvim_win_close(win, true)
+                jump(location)
```
blinry: This callback still performs a jump. Was the intention to also set `following_userid` here, and start following it?

nvim-plugin/lua/ethersync/cursor.lua
```diff
@@ -45,6 +46,17 @@ local function jump(location)
     end
 end
 
+local function follow(user_id)
+    local cursors = user_cursors[user_id].cursors
+    local cursor = cursors[#cursors]
```
blinry, line 43: Here's a subtle bug: When a user that you're currently following disconnects, an error will be thrown here. The reason is that in that case, to clear the user's cursors, we set the list of cursor positions to an empty array. As a result, `cursor` will be nil here.

You could check for that case, and abort the jump here.

I even thought about un-setting `following_userid` in that case, but maybe it's nice to continue the follow after the other user has briefly closed Neovim? You can decide what you like better!

MichaelBitard:
For my future self, we got a bunch of errors:

```
Error executing vim.schedule lua callback: /usr/share/nvim/runtime/lua/vim/lsp/util.lua:947: Cursor position outside buffer
stack traceback:
	[C]: in function 'nvim_win_set_cursor'
	/usr/share/nvim/runtime/lua/vim/lsp/util.lua:947: in function 'show_document'
	...workspace/ethersync/nvim-plugin/lua/ethersync/cursor.lua:45: in function 'jump'
	...workspace/ethersync/nvim-plugin/lua/ethersync/cursor.lua:57: in function 'follow'
	...workspace/ethersync/nvim-plugin/lua/ethersync/cursor.lua:148: in function 'set_cursor'
	/opt/workspace/ethersync/nvim-plugin/plugin/ethersync.lua:60: in function 'process_operation_for_editor'
	/opt/workspace/ethersync/nvim-plugin/plugin/ethersync.lua:83: in function ''
	vim/_editor.lua: in function <vim/_editor.lua:0>

```

Not sure how it happened though...

nvim-plugin/README.md
MichaelBitard: Done!

nvim-plugin/plugin/ethersync.lua
MichaelBitard: I'm not a native english speaker either, but Unfollow makes more sense!

nvim-plugin/lua/ethersync/cursor.lua
MichaelBitard: I removed it completely, it was called only in the new jump_to_user_id

nvim-plugin/lua/ethersync/cursor.lua
MichaelBitard: Yeah, I saw that and was afraid on how callback worked in lua. But it was quite simple

nvim-plugin/lua/ethersync/cursor.lua
MichaelBitard: Yeap, that was a mistake

nvim-plugin/lua/ethersync/cursor.lua
MichaelBitard: Nice catch!

nvim-plugin/README.md
```diff
@@ -31,9 +31,10 @@ Usually, you will add the string `"ethersync/ethersync-nvim"` to your plugin man
 {
   "ethersync/ethersync-nvim",
   keys = { 
-    { "<leader>j", "<cmd>EthersyncJumpToCursor<cr>" },
-    { "<leader>f", "<cmd>EthersyncFollow<cr>" },
-    { "<leader>u", "<cmd>EthersyncStopFollow<cr>" },
+    { "<leader>ej", "<cmd>EthersyncJumpToCursor<cr>" },
+    { "<leader>ef", "<cmd>EthersyncFollow<cr>" },
+    { "<leader>eu", "<cmd>EthersyncUnfollow<cr>" },
```
blinry: My feeling is: If we have the `<esc>` mapping, we don't even need to suggest the `<leader>eu` one. Less mappings to keep track of.

nvim-plugin/lua/ethersync/cursor.lua
blinry: I like how it turned out! :)

blinry (approved):
Works perfectly for me! :) Thank you! I have another suggestion regarding the example mappings, but that's kind of optional.

To wrap this up, could you squash your commits together? (If that's any trouble, I can also do it when merging.)

And feel free to add yourself to the copyright header of cursor.lua – credit where credit is due! :) (We're following [the REUSE spec](https://reuse.software/)!)

MichaelBitard:
Thanks for everything!

I have squashed the commits, added myself to the copyrights and removed the `<leader>eu` keymap!

I hope we'll keep using it and that I'll be able to contribute again on this, it's been a pleasure!

PR #370, opened 2025-08-31
More lints and refactorings

EdJoPaTo:
This mostly enables the `clippy::pedantic` and `clippy::nursery` lint groups. Some lints are still allowed (=ignored) in the `Cargo.toml` where some of them might indicate potential existing bugs. I suspect they should be fixed on their own as it might result in behavior changes (albeit probably fixes).

Just as a heads-up: Coding with pedantic and nursery lints enabled can be a bit annoying at first but is teaching a lot of how to write better Rust code and after a while it ends up the normal code being written.

This also enabled running `cargo clippy` and `cargo doc` from the CI.

Regarding `must_use`, I suspect that a bunch of structs in `types.rs` should be marked as `must_use` (as in, whenever they are returned, they have to be used or explicitly ignored). But tests seem to not use the results of these methods, so I'm not sure what the best way there is.
Therefor, I have explicitly allowed `must_use_candidate` for now.

This does not change the minimal required Rust version. I specified it in the `Cargo.toml` to make sure lints don't increase it by accident.
This should not change the behavior but might be slightly more performant as some changes should be more efficient now (for example less clones).


daemon/src/config.rs
```diff
@@ -172,10 +177,9 @@ pub fn add_ethersync_to_global_gitignore() -> Result<()> {
     let ignore_file_path = git_config_dir.join("ignore");
 
     sandbox::create_dir_all(&git_config_dir, &git_config_dir)?;
-    let bytes_in =
-        sandbox::read_file(&git_config_dir, &ignore_file_path).unwrap_or("".to_string().into());
+    let bytes_in = sandbox::read_file(&git_config_dir, &ignore_file_path).unwrap_or_default();
```
EdJoPaTo, line 180: This created an empty String, then created a `Vec<u8>` from it which is empty as the String was also empty. Using the default is way simpler but not sure its actually the intention here?

daemon/src/document.rs
```diff
@@ -268,28 +274,29 @@ impl Document {
...
-        }
+        self.top_level_map_obj("files")
+            .map(|file_map| {
+                self.doc
+                    .keys(file_map)
+                    .map(|k| RelativePath::new(&k))
+                    .collect()
+            })
+            .unwrap_or_default()
     }
```
EdJoPaTo, line 287: This might be a bit annoying to read but does the same as before, default to an empty Vec.

daemon/Cargo.toml
```diff
@@ -34,6 +35,35 @@ bench = false
...
+map_with_unused_argument_over_ranges = "warn"
+
+filetype_is_file = "allow" # TODO: probably relevant for this project
+needless_continue = "allow" # TODO: probably a bug that needs to be fixed
+too_long_first_doc_paragraph = "allow"
+needless_pass_by_ref_mut = "allow" # TODO: could be a bug, could be intended
+
+cast_possible_truncation = "allow"
+cast_possible_wrap = "allow"
+cast_sign_loss = "allow"
```
EdJoPaTo, line 56: These numeric cast thingies are probably no issue and not that relevant in this project. But it might be a good idea to move these to their actual use with `#[expect(…)]` instead to indicate more clearly about it where it actually is happening.

daemon/Cargo.toml
```diff
@@ -34,6 +35,35 @@ bench = false
...
+cognitive_complexity = "allow"
+missing_const_for_fn = "allow"
```
EdJoPaTo, line 59: const functions might be neat but probably not very helpful here as they are likely still executed on runtime.

daemon/Cargo.toml
```diff
@@ -34,6 +35,35 @@ bench = false
...
+missing_errors_doc = "allow"
+missing_panics_doc = "allow"
```
EdJoPaTo, line 61: When the docs are a goal of this project then these should be fixed

daemon/Cargo.toml
```diff
@@ -34,6 +35,35 @@ bench = false
...
+must_use_candidate = "allow"
+should_panic_without_expect = "allow"
```
EdJoPaTo, line 63: This would make sure the tests are catching the expected panics

daemon/Cargo.toml
```diff
@@ -34,6 +35,35 @@ bench = false
 doc = false
 required-features = ["executable-deps"]
 
+[features]
+default = ["executable-deps"]
+executable-deps = ["dep:clap", "tokio/signal", "tokio/rt-multi-thread"]
```
EdJoPaTo, line 40: I moved the features up as they are most often one of the more interesting parts when looking into a `Cargo.toml`. Also, they are more alphabetically sorted with the rest of the keys now too.

Personally I also put the default feature first as it has a special meaning while the others don't. In this case it's also the alphabetical order.

daemon/src/config.rs
zormit: not sure either what was the intention, probably it was just coded to make the compiler happy 😅 i think the default is fine!

daemon/Cargo.toml
zormit: I don't think we have that goal right now :)

daemon/Cargo.toml
zormit: Sounds good, added a TODO.

PR #372, opened 2025-09-01
Allow configuring VS Code plugin

blinry:
Same goal as in #368. I'm basing this on #368 to benefit from the URI handling improvements.

This works, but is a bit rough:

- [x] Go through TODOs, especially those which throw errors!
- [ ] If multiple configs match, currently only one is started, I think
- [ ] Split into more files?
- [x] Re-think config format:
    - [ ] Maybe the names of the configurations could be keys in a dictionary
    - [ ] "activatePattern" is probably not a good name
    - [ ] "activatePattern" is less flexible than Neovim's root_dir…
    - [ ] "activatePattern": Document well what it matches on: The URI or the filename? (What should it do?)
    - [ ] "activatePattern" should probably need to match the *full* filename, instead of one part? How could it match directories like `~/http/…` (for Braid?)
- [x] We don't save anywhere which TextDocuments are attached to which Clients. Do we need to? In Neovim, we do, for closing purposes, I think. Is closing implemented properly here?
- [x] Document the functionality
- [x] Delete `enabled`

### DIFF
--- a/vscode-plugin/README.md
+++ b/vscode-plugin/README.md
@@ -5,8 +5,37 @@ SPDX-FileCopyrightText: 2024 zormit <nt4u@kpvn.de>
 SPDX-License-Identifier: CC-BY-SA-4.0
 -->
 
-# VS Code plugin for 🍃 [Ethersync](https://github.com/ethersync/ethersync)
+# VS Code plugin for 🍃 [Ethersync](https://github.com/ethersync/ethersync)-compatible projects
 
-Note: You will need to install and run the Ethersync daemon in addition to this plugin!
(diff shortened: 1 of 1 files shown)


nvim-plugin/lua/ethersync.lua
```diff
@@ -220,25 +220,17 @@ local function on_buffer_close()
     debug("on_buffer_close: " .. closed_file)
 
     -- Find the correct client, and remove this buffer from it.
-    local client = nil
```
zormit, line 223: We sneaked in some additions to #368 here :)

PR #378, opened 2025-09-06
vscode: improved strategy for distinguishing remote edits

dglittle:
Previously, remote edits were detected by adding them to attemptedRemoteEdits, and then, in processUserEdit, seeing if the edits matches one of those.. however, sometimes vscode will take an edit in applyEdit, and modify the patches a bit before emitting an event to processUserEdit. For example, if a document contains the text "a", and a remote edit replaces that with "AaA", then vscode will actually emit an event to processUserEdit which has two deltas: an "A" at the beginning, and an "A" after the first character.. which looks different to the isRemoteEdit function than "AaA".. so now, the strategy is to just increment a number of outstanding remote edits, and assume that those edits will arrive first to processUserEdit, and we just decrement the number when they do.. this has been tested in a circumstance that was having issues because of misidentified remote edits, and it now works in the cases where it didn't.. still, this new strategy may have some hole. Its not clear to me how vscode "wants" users to distinguish true user edits from their own applyEdit commands.

### DIFF
--- a/vscode-plugin/src/extension.ts
+++ b/vscode-plugin/src/extension.ts
@@ -123,7 +123,7 @@ let clients: Client[] = []
 // TODO: if we load from disk, this will also cause edits :-/
 let t0 = Date.now()
 const mutex = new Mutex()
-let attemptedRemoteEdits: Set<vscode.TextEdit[]> = new Set()
+let attemptedRemoteEdits: number = 0
 
 const openType = new rpc.RequestType<{uri: string; content: string}, string, void>("open")
 const closeType = new rpc.RequestType<{uri: string}, string, void>("close")
@@ -224,7 +224,7 @@ async function processEditFromDaemon(client: Client, edit: Edit) {
             const document = documentForUri(uri)
             if (document) {
                 let textEdit = ethersyncDeltasToVSCodeTextEdits(document, edit.delta)
-                attemptedRemoteEdits.add(textEdit)
+                attemptedRemoteEdits++
                 let worked = await applyEdit(client, document, edit)
                 if (worked) {
                     revision.daemon += 1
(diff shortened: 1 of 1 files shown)


blinry:
Good catch! That VS Code changes the text edit is pretty surprising. It seems to calculate its own diff?

We confirmed that this is happening, following the example you gave.

Unfortunately, we also confirmed that the approach of counting expected remote edits sometimes ignores the wrong changes: I'm guessing the `vscode.workspace.onDidChangeTextDocument` callback works so asynchronously that it's not guaranteed that the "next" instance of where it's called is always the remote edit we're applying. Especially when there are lots of local and remote edits at the same time, it can happen that this approach ignores the user's edits, and doesn't send them out to the daemon.

We're not sure how to solve this problem right now.

zormit:
Here are some further pointers into how VS Code seems to work and what we could do:
- it seems the code for changing edits that are being applied is in [`bulkTextEdits.ts`](https://github.com/microsoft/vscode/blob/233a3bd7365d09686fb7bc9463333181878c6a53/src/vs/workbench/contrib/bulkEdit/browser/bulkTextEdits.ts#L201) (via [this](https://github.com/microsoft/vscode/issues/129969#issuecomment-899386508) issue comment. The issue author has the [same problem](https://github.com/microsoft/vscode/issues/129969#issuecomment-899618354) as we do.). Wondering if we can somehow use this approach and make the edits minimal before putting them into the tracking set.
- we had thought about this before, but I don't think documented anywhere: The [`TextDocumentChangeEvent`](https://code.visualstudio.com/api/references/vscode-api#TextDocumentChangeEvent) has a `TextDocumentChangeReason` since https://github.com/microsoft/vscode/pull/124996, unfortunately it only covers undo/redo and nothing else (as proposed in the underlying issue https://github.com/microsoft/vscode/issues/120617)
- Can we exploit some underlying ordering guarantees? https://github.com/microsoft/vscode/issues/111331 Our tests breaking @dglittle's approach seem to imply that the way we coded it right now we don't have ordering guarantees. Maybe we are able to fix it somehow. [This comment](https://github.com/microsoft/vscode/issues/122883#issuecomment-833487051) also implies that there might be a way to "[wait] for `vscode.workspace.applyEdit` to finish before doing the next"

blinry:
My second attempt at tackling this problem in a more general way is on the [vscode-compare-results](https://github.com/ethersync/ethersync/tree/vscode-compare-results) branch. It's in a draft state, but I thought I'd link it here.

dglittle:
thanks for following up with it! excited to try to new solution (though I read it also has some edge cases to worry about.. I'll see if I see them in real life)

PR #382, opened 2025-09-09
fix: If a wormhole mailbox cannot be created, warn and proceed

zormit:
It's a simple fix that I'm not sure how to easily test.


daemon/src/wormhole.rs
```diff
@@ -15,7 +15,10 @@ pub async fn put_secret_address_into_wormhole(address: &str) {
 
     tokio::spawn(async move {
         loop {
-            let mailbox_connection = MailboxConnection::create(config.clone(), 2).await.unwrap();
+            let Ok(mailbox_connection) = MailboxConnection::create(config.clone(), 2).await else {
+                warn!("Creating a connection with magic wormhole mailbox failed. Skipping.");
```
zormit: Or maybe rather
```suggestion
                warn!("Creating a connection with magic wormhole mailbox failed. Retrying.");
```
as it's just coming back to this line. Now I'm not sure anymore if my fix is legti 🤔

blinry: An alternative could be to `return` from the attempt entirely. If the wormhole servers would ever go down, for example, right now we'd attempt to create a mailbox every 500 ms… so maybe we should "give up" after one try?

Issue #365, opened 2025-08-27
Add a `.ethersync/.gitignore` file


blinry:
When researching #364, I learned jujutsu's solution to have the .jj directory ignored by Git: They just add a .gitignore containing `/*` to their subdirectory!

We could do the same, instead of instructing users to add the .ethersync directory to ~/.config/git/ignore!

blinry:
Implemented in #384.

Issue #364, opened 2025-08-26
Safer operation on version-controlled projects


meejah:
It might make sense to ignore any files or directories that start with "." by default.

In our particular use-case, @shapr and I tried this out in a project where we have ".jj/*" directories (a version-control system) and the "host" person overwrote all the repository files in the "guest" system.

Probably causes problems for ".git" as well, although we haven't yet tested that.

(My bug title is "a possible solution" I suppose, the real potential bug here is something like "it is surprising if .git or .jj or .svn or other version-control files are overwritten from the host to guest systems).

On the plus side, it "just worked" as per the quickstart, and we are able to quickly see each other's edits. 🥳

blinry:
Hi! :)

Oh right, I definitely think ignoring VCS directories like `.jj`, `.hg` and others by default makes a lot of sense. We were already ignoring `.git`, but only that one.

I'm also using jj, and I was a bit surprised about why this had never been an issue for me – I think the reason is that I almost always have Git colocated, so there's also a `.git` directory. In that case, the ignore crate will obey the `.gitignore` inside `.jj`, so Ethersync will ignore the directory, as well. We changed the behavior to *always* obeying .gitignores in 69181bff, so there it should already work better for you.

(Side note: On current main, we now have support for binary files, so it's actually rather interesting to *not* ignore .git! You can type commit messages together, and after one person created a commit, the other will immediately have it. Reminded me of `pear-on`! It's very untested and probably unstable yet, but it might be fun to offer that as an option in the future!)

As for ignoring all dot files, I think that's not the proper solution. Files like `.prettierrc` or `.pre-commit-config.yaml` would be great to be shared and potentially collaborated on!

Summary: As a workaround, you could create `.git` (even an empty one) in the projects you're working on, I think! As of the main branch, it should already be ignored, because of the changed ignore crate setting. We might still look into ignoring other VCS directories. At least `.hg`, `.svn`, `.bzr`, `.pijul`, I think?

meejah:
Yeah, this stuff is tough.

If there's some sort of API for "add excluded files to the collaborative session" at least there would be a path forward if the defaults were wrong. This could then allow somewhat-conservative defaults.

What about this?

- _directories_ that start with "." are ignored (by default)
- _files_ that start with "." are ... included by default?
- anything in .gitignore is ignored by default
- if .git exists, un-ignore anything tracked by Git

[...]

PR #384, opened 2025-09-10
Ignore VCS directories, and new way of ignoring .ethersync

blinry:
See #364 and #365.

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -93,10 +93,7 @@ async fn main() -> Result<()> {
                 );
             }
 
-            // Silently make sure that .ethersync is ignored.
-            if config::ethersync_directory_should_be_ignored_but_isnt(&directory) {
-                config::add_ethersync_to_local_gitignore(&directory)?;
-            }
+            config::ensure_ethersync_is_ignored(&directory)?;
 
             let mut init_doc = false;
             let mut app_config;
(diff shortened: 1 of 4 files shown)


daemon/src/main.rs
```diff
@@ -93,13 +93,9 @@ async fn main() -> Result<()> {
                 );
             }
 
+            // Silently make sure that .ethersync is ignored.
             if config::ethersync_directory_should_be_ignored_but_isnt(&directory) {
-                if ask("Ethersync uses the directory .ethersync/ to store sensitive secrets. Add it to your global ~/.config/git/ignore?")? {
```
zormit, line 97: Very nice to get rid of a potentially confusing/tedious question to the user (and make it better at the same time =))

daemon/src/main.rs
```diff
@@ -93,13 +93,9 @@ async fn main() -> Result<()> {
...
-                    config::add_ethersync_to_global_gitignore()?;
-                    info!("Added! Resuming launch.");
-                } else {
-                    bail!("Aborting launch. We *really* don't want you to accidentally commit your secrets. Add .ethersync/ to a (global or local) .gitignore and try again.");
-                }
+                config::add_ethersync_to_local_gitignore(&directory)?;
```
zormit: [refactoring] I'm wondering whether this logic needs to be here. It was here because of the user prompt at startup. We could now refactor it to a `config::ensure_ethersync_is_ignored` or something like that and then just call that from here instead of having the conditional here.

PR #386, opened 2025-09-17
Versionining between Editor and Daemon

### LINKED ISSUE #278: Version & compatibility check between editor plugin(s) and ethersync client
In #276 we're on a shed, where there'll be a compatibility issue if users do not update both the daemon and the plugin. Also there might be (in fact, there [_are_](https://github.com/ethersync/ethersync-jetbrains)) plugins around, that are not necessarily developed in the same speed as the daemon.

So we can think of some basic compatibility & version check. How?
- the plugin could run a form of `ethersync --version` when starting, and have some pattern matching logic that detects, when it's not the expected version
- the daemon could refuse to start, when it's called from the correct plugin version. This might be more tricky, given that there are multiple version of plugins possible and the daemon might not even know about them
-in the case of #276, the `--directory` is a new parameter, which is optional. We could make it a strict requirement for `ethersync client` and thus make older plugins fail.
- the plugin could also run some `ethersync client --expected-version x.x.x` and refuse to start if it doesn't match. Seems verbose.

Two related (afaik currently unwritten) issues are compatibility between two daemons and a versioning of the daemon-editor protocol.

zormit:
closes #278

### DIFF
--- a/daemon/src/editor.rs
+++ b/daemon/src/editor.rs
@@ -136,28 +136,54 @@ async fn handle_editor_connection(
     if let Some(message) = reader.next().await {
         match message {
             Ok(incoming_message) => {
-                match incoming_message {
+                // TODO: refactor this into a helper
+                let response = match incoming_message {
                     IncomingMessage::Request {
                         id,
                         payload: EditorProtocolMessageFromEditor::Initialize { version },
                     } => {
-                        // TODO: check version here!
-                        let message = OutgoingMessage::Response(JSONRPCResponse::RequestSuccess {
-                            id,
-                            result: "success".to_string(),
-                        });
-                        writer.send(message).await.unwrap_or_else(|err| {
-                            error!("Failed to send message to editor: {err} Removing editor.");
-                        });
-                        info!("Editor #{editor_id} connected.");
+                        let expected_protocol_version = "0.9";
+                        if version == expected_protocol_version {
+                            info!("Editor #{editor_id} connected.");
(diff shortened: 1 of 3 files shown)


blinry:
I like "version 2" better – it mimicks LSP's approach, and would allow arbitrary changes in upcoming versions. For example, we might want to receive all cursors before we even send an open.

Some notes for what would be required to finish it up:

- In the plugins, the version dance would need to happen when first spawning the connection to the Collaboration Server – in Neovim, for example, in `connection.connect()`.
- Plugins still need to send an "initialized" message, and the daemon needs to wait for it. This requires three states, I think: "pre-initialized", "daemon-initialized" (after we replied to the "initialize"), "editor-initialized" (after we received an "initialized").
- When editor and daemon are incompatible, the daemon should probably close the connection? (*"Dave, this conversation can serve no purpose anymore. Goodbye."*)
- We need to think about the versioning scheme for the editor protocol.
	- Would it just be an integer? The past version could be (implicitly) 1, and the upcoming version could be 2?
	    - Alternatively, the upcoming version could be "0.8"! I think I quite like that. Once we go "out of beta", the version could be 1.
	- I don't see a need for minor or patch versions – it's not clear to me what they would mean.
- What should happen when old plugins connect, which don't support the version dance yet? They will send an "open" directly. The daemon should probably reject this and disconnect.
- In the daemon, all other message types (edit and close) also should require an initialized connection.

PR #387, opened 2025-09-22
vscode: Don't try to auto-save empty files

### LINKED ISSUE #358: VS Code: Error message when deleting entire text content
The error is:

> Failed to save 'file': The content of the file is newer. Please compare your version with the file contents or overwrite the content of the file with your changes.

When pressing "Compare", both sides are empty, however. Overwriting solves the issue without problems.

blinry:
For unknown reasons, when overwriting an empty file with another empty file, VS Code will complain: "The content of the file is newer."

Upon quick testing, I found no problem when closing the file in that state.

Fixes #358.

### DIFF
--- a/vscode-plugin/src/extension.ts
+++ b/vscode-plugin/src/extension.ts
@@ -442,7 +442,11 @@ function processUserEdit(event: vscode.TextDocumentChangeEvent) {
 
                 // Attempt auto-save to avoid the situation where one user closes a modified
                 // document without saving, which will cause VS Code to undo the dirty changes.
-                document.save()
+                // If the document is empty, don't save - this will trigger a
+                // "The content of the file is newer." warning.
+                if (document.getText() !== "") {
+                    document.save()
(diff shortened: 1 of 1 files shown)


vscode-plugin/src/extension.ts
```diff
@@ -442,7 +442,11 @@ function processUserEdit(event: vscode.TextDocumentChangeEvent) {
 
                 // Attempt auto-save to avoid the situation where one user closes a modified
                 // document without saving, which will cause VS Code to undo the dirty changes.
-                document.save()
+                // If the document is empty, don't save - this will trigger a
+                // "The content of the file is newer." warning.
+                if (document.getText() !== "") {
```
zormit, line 447: FWIW I have done a tiny research if it's better to compare for a non-zero-`.length` but I didn't find that that would be considerably faster or has other advantages. Comparing for an empty string is very explicit so let's keep it :)

PR #388, opened 2025-09-24
bump automerge to 1.0.0-beta.3

zormit:
The idea is that everything is a little bit more efficient with the new automerge version :-)


blinry:
We benchmarked a bit: Startup time went down from 8 s to 3 s in our shared wiki.
And memory usage went down from 1 GB to 160 MB.

PR #390, opened 2025-09-24
vscode: Instead of expecting certain remote edits, expect their result

blinry:
When a remote edit replaces "x" with "XxX", VS Code will see a
TextDocumentChangeEvent that inserts two X's around the x. Because of
that, it's not possible to directly compare the edits.

Here's a proposed solution: Store the expected resulting content of the
document after the remote edit, then see if a local edit results in that
content.

### DIFF
--- a/vscode-plugin/package.json
+++ b/vscode-plugin/package.json
@@ -44,7 +44,8 @@
                             }
                         },
                         "required": [
-                            "cmd", "rootMarkers"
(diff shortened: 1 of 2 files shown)


blinry:
Even though this PR doesn't solve all VS Code consistency issues, we decided that it's an improvement, and we can iterate more on it in the future.

PR #393, opened 2025-09-25
Extract cross-platform code into shared crate

winniehell:
helps with integration of ethersync code into https://github.com/ethersync/ethersync-web/

see also https://github.com/ethersync/ethersync-web/compare/main...ethersync-submodule

### DIFF
--- a/daemon/shared/src/platform.rs
+++ b/daemon/shared/src/platform.rs
@@ -4,13 +4,13 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
 #[cfg(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))]
-pub(crate) use dioxus_core::prelude::spawn;
-
-#[cfg(not(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none"))))]
-pub(crate) use tokio::spawn;
+pub mod webassembly;
 
 #[cfg(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))]
-pub(crate) use dioxus_hooks::UnboundedSender as Sender;
+pub use webassembly::*;
 
 #[cfg(not(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none"))))]
-pub(crate) use tokio::sync::mpsc::Sender;
+pub mod desktop;
+
+#[cfg(not(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none"))))]
+pub use desktop::*;
--- a/daemon/shared/src/platform/desktop.rs
+++ b/daemon/shared/src/platform/desktop.rs
@@ -1,0 +1,7 @@
+// SPDX-FileCopyrightText: 2025 blinry <mail@blinry.org>
+// SPDX-FileCopyrightText: 2025 zormit <nt4u@kpvn.de>
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+pub use tokio::spawn;
+pub use tokio::sync::mpsc::Sender;
--- a/daemon/shared/src/platform/webassembly.rs
+++ b/daemon/shared/src/platform/webassembly.rs
@@ -1,0 +1,7 @@
+// SPDX-FileCopyrightText: 2025 blinry <mail@blinry.org>
+// SPDX-FileCopyrightText: 2025 zormit <nt4u@kpvn.de>
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+pub use dioxus_core::prelude::spawn;
+pub use dioxus_hooks::UnboundedSender as Sender;


daemon/shared/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+# SPDX-License-Identifier: AGPL-3.0-or-later
+[package]
+name = "ethersync-shared"
+description = "Cross-platform portion of the code."
+version = "0.0.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+iroh = "0.35.0"
```
winniehell, line 20: I'm unsure if there is a way to reuse the top level version here. we could use a [workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) but I think that would require the `Cargo.toml` of the daemon package to be on a deeper level.

winniehell: in the long term, I think we can remove the iroh dependency from the daemon because everything that relies on it can be part of the shared crate.

daemon/src/editor_connection.rs
```diff
@@ -3,21 +3,59 @@
...
+impl FileUri {
+    pub fn from_absolute_path(path: &AbsolutePath) -> Self {
+        Self::try_from(format!("file://{}", path.display()))
+            .expect("Should be able to create File URI from absolute path")
+    }
+
+    pub fn to_absolute_path(&self) -> AbsolutePath {
+        let path_buf = Url::parse(self)
+            .expect("Should be able to parse file:// URL as Url")
+            .to_file_path()
```
winniehell, line 35: this cannot be used in WASM context, so I have moved the whole struct out of path

daemon/src/path.rs
```diff
@@ -95,38 +89,6 @@ impl From<&RelativePath> for Prop {
     }
 }
 
-// TODO: Wrap the newtype around url::Url instead?
-#[derive(Debug, Clone, PartialEq, Eq, Hash, Deref)]
-#[must_use]
-pub struct FileUri(String);
```
winniehell, line 101: see https://github.com/ethersync/ethersync/pull/393/files#r2386327276

daemon/src/daemon.rs
```diff
@@ -983,7 +986,12 @@ impl Daemon {
             );
         }
         if app_config.emit_join_code {
-            put_secret_address_into_wormhole(address).await;
+            let address = connection_manager.secret_address().clone();
+            tokio::spawn(async move {
```
winniehell, line 993: in WebAssembly we have to use `dioxus::spawn` instead of `tokio::spawn`. it would be helpful to make that distinction in one place and then simply import a general `spawn` from there.

winniehell: done

daemon/shared/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+description = "Cross-platform portion of the code."
+version = "0.0.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+anyhow = "1.0.81"
+automerge = "1.0.0-beta.3"
+derive_more = { version = "2.0.1", features = ["as_ref", "deref", "display"] }
+dioxus-core = "0.6.0"
```
winniehell, line 16: can this be included only if we build for WebAssembly?

daemon/shared/src/platform.rs
```diff
@@ -0,0 +1,16 @@
+// SPDX-FileCopyrightText: 2025 blinry <mail@blinry.org>
+// SPDX-FileCopyrightText: 2025 zormit <nt4u@kpvn.de>
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+#[cfg(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))]
+pub(crate) use dioxus_core::prelude::spawn;
+
+#[cfg(not(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none"))))]
```
winniehell: is there an easier way to do this?

daemon/shared/src/wormhole.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use anyhow::Result;
+use magic_wormhole::{transfer, AppID, Code, MailboxConnection, Wormhole};
+use std::{str::FromStr};
+use tracing::{error, warn};
+use crate::secret_address::SecretAddress;
+use crate::platform::{spawn, Sender};
+
+#[cfg(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))]
+use futures_util::sink::SinkExt;
```
winniehell, line 14: can this be moved to platform instead?

daemon/shared/src/wormhole.rs
```diff
@@ -0,0 +1,49 @@
...
+    spawn(async move {
+        let Ok(mailbox_connection) = MailboxConnection::create(config.clone(), 2).await else {
+            error!(
+            "Failed to share join code via magic wormhole. Restart Ethersync to try again."
+        );
+            return;
+        };
+
+        let code = mailbox_connection.code().clone();
+        if join_code_tx.to_owned().send(code).await.is_err() {
```
winniehell, line 29: without `.to_owned()` build for WebAssembly fails with

```
   2.872s  INFO  error[E0596]: cannot borrow `join_code_tx` as mutable, as it is not declared as mutable
  --> ethersync/daemon/shared/src/wormhole.rs:29:12
   |
29 |         if join_code_tx.send(code).await.is_err() {
   |            ^^^^^^^^^^^^ cannot borrow as mutable
```

PR #401, opened 2025-10-03
display tooltip on new cursor connection

MichaelBitard:
Hi,

I don't know if this could be interesting to merge, but on our project, it can take some time before the cursor is visible on the other machine. This print a little information as soon as the cursor is in. What way we don't have to spam the EthersyncInfo command anymore :D

I'm not sure I understand the multicursor code though, so maybe what i'm doing is not really what I want, but it seems to "do the job".

### DIFF
--- a/nvim-plugin/lua/ethersync/cursor.lua
+++ b/nvim-plugin/lua/ethersync/cursor.lua
@@ -62,7 +62,7 @@ end
 function M.set_cursor(uri, user_id, name, ranges)
     -- Find correct buffer to apply edits to.
     local bufnr = vim.uri_to_bufnr(uri)
-
+    local new_cursor = false
     if user_cursors[user_id] then
         for _, user_cursor in ipairs(user_cursors[user_id].cursors) do
             if user_cursor.extmark then
@@ -71,6 +71,8 @@ function M.set_cursor(uri, user_id, name, ranges)
                 vim.api.nvim_buf_del_extmark(old_bufnr, cursor_namespace, old_id)
             end
         end
+    else
+        new_cursor = true
     end
     user_cursors[user_id] = { name = name, cursors = {} }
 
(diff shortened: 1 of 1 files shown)


zormit (commented):
Hi @MichaelBitard,
thanks for taking action to attempt to improve the UI for user awareness! 👍🏼 

My thoughts on this after testing it quickly:
- it seems the message also appears for the actively joining peer (who opens a file). It feels a bit noisy to get a message for every incoming cursor while opening the file myself.
- Dependent on the nvim setup the `print` might need a confirmation (pressing return). So it's not always rendered as a "tooltip". If that happens, then it's a bit annoying to having to confirm it every time when opening a file.
    - Also it seems to be blocking: The own cursor is only sent out, when I confirm the incoming cursors. Not sure how to "detach" that from each other.

I didn't think a lot about alternatives yet, but this solution seems not fully mergeable yet. If you find it very very useful to have in the codebase we could think about putting it behind some kind of configuration flag so it's disabled by default and enabled for you.

The "dream" solution I'd imagine would be some sort of extra buffer showing a list of connected users that you _could_ keep open on the side similar to a filetree or something.

btw, I'm not even sure if the level of the editor is best for such information or if it would be worthwhile to have a view in the daemon for connected peers (which would be a bit less fine-grained than cursors, I guess). We recently tested [git withme](https://git.sr.ht/~meejah/git-withme), which has a nice TUI solution for something that would be on the daemon side, but now I'm side-tracking this PR a bit :D

MichaelBitard:
Hi @zormit,

Once again, thanks for the really detailed answer!
I agree with you, after working with it for a while, it can be quite annoying.

I think something like the https://github.com/nvim-mini/mini.map could be better as you said, I'll see if I can do it.

I agree with the daemon, it could be really neat to "talk" with it in a cli way. That way we could see what it is doing, which peers are connected, we could see the cursors too I guess.

I'm closing this for now.

Thanks again!

PR #407, opened 2025-10-10
Make the watcher "debounce" file events

blinry:
Motivation: On a Git checkout, a file gets deleted and then recreated. It would be great to "wait a bit" in the watcher, and only communicate a "change".

- [x] Check why the `change` test fails on macOS -> accept it for now/disable test
- [x] Make Clippy happy (and why didn't blinry see it locally?)
- [x] Indeed change the `WatcherEvent` type to contain `WatcherEventType`
- [x] Think about which pending-timeout would be good.
- [x] Try it – does this actually help with git checkouts?
- [x] The Beginning of `run` (the timer calculation) is pretty ugly...

Maybe these "next steps":

- Gesamt-Wartezeit in pending_events merken; wenn zu lang schon nix rauskommuniziert, dann sollten wir!
    - => das wird ggf. auch durch rescan erkannt, also nicht nötig?
    - oder: "Rescan pro File"?
- Rescan aus spawn_file_watcher raus und in Watcher rein
    - (Lots of work, needs cached for file existance (+ possibly content) in the watcher...)
- Macht es Sinn, das Created-statt-Changed-Problem im Watcher zu lösen?
    - [x] Erstmal im daemon gelöst
- Is it possible that for creations/modifications, we could immediately communicate them, while we want to be careful around communicating a deletion?
- Should the WatcherEventType-s `Created` and `Changed` be combined, if they are handled the same? Or just have one type, and the daemon checks if the file is there or not?

### DIFF
--- a/daemon/src/watcher.rs
+++ b/daemon/src/watcher.rs
@@ -81,22 +81,16 @@ impl Watcher {
 
     async fn run(&mut self) {
         loop {
-            let (duration, event) = self.upcoming_timeout();
+            let timeout_maybe = self.upcoming_timeout();
             tokio::select! {
-                () = sleep(duration) => {
-                    match event {
-                        Some(WatcherEvent{ file_path, event_type }) => {
-                            // Removed triggered pending event.
-                            self.pending_events.remove(&file_path);
-                            self.event_tx.send(WatcherEvent{
-                                file_path,
-                                event_type,
-                            }).await.expect("Channel closed");
-                        },
-                        None => {
-                            panic!("Watcher timed out without an event. This is a bug.");
-                        }
-                    }
+                WatcherEvent{ file_path, event_type } = Self::await_timeout(timeout_maybe) => {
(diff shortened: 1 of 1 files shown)


daemon/src/watcher.rs
```diff
@@ -213,6 +184,29 @@ impl Watcher {
...
+            .iter()
+            .min_by_key(|(_, pending_event)| pending_event.timestamp);
+        if let Some((next_file_path, next_pending_event)) = next_pending_maybe {
+            let pending_event = TimeoutEvent::PendingEvent {
+                file_path: next_file_path.clone(),
+                event_type: next_pending_event.event_type.clone(),
+            };
+
+            let now = SystemTime::now();
+            match next_pending_event.timestamp.duration_since(now) {
```
zormit, line 168: As we want to use `pending_event` in both arms, it's probably better to use match instead of `map_or_else`
https://stackoverflow.com/a/69174092

PR #425, opened 2025-10-16
daemon: When writing to a .git/objects/... file, don't overwrite

### LINKED ISSUE #415: daemon crashes when file is not writeable
This came up when sharing the `.git` directory:

```
    thread 'tokio-runtime-worker' panicked at src/daemon.rs:589:35:
    Failed to write to file '[...]test/.git/objects/58/ef3f6e3a9f0129393f3e1098a164d94b7be245': Permission denied (os error 13)
``` 

Either we show a warning and don't write the file or we force-write it (🤔), but at least it shouldn't crash.

Related to #400 in the sense that Ethersync is a bit ignorant about file permissions :)

blinry:
Instead, assume that the content will be okay.

This avoids a problem with file permissions if the content differs (which can likely happen because zlib is not reproducible). Fixes #415.

### DIFF
--- a/daemon/src/daemon.rs
+++ b/daemon/src/daemon.rs
@@ -573,6 +573,15 @@ impl DocumentActor {
         if sandbox::exists(&self.app_config.base_dir, &abs_path)
             .expect("Failed to check for file existence before writing to it")
         {
+            // Special case: If we want to write a .git/objects/... file, and there's one already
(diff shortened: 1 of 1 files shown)


zormit (approved):
I find it quite hacky to hardcode exceptions for `.git` stuff now and it will still panic on other read-only files 😅 I will accept and merge because it does the job and we want to user-test this feature :)

PR #426, opened 2025-10-17
nvim: automatically unfollow as soon as the user type something

### LINKED ISSUE #420: Neovim: Stop follow mode when moving/on any key press
## Problem

Sometimes, I follow someone, and then move around to check out something nearby. Then, I forget that I'm in follow mode, and am later surprised when the followed person moves around, and I jump to them.

## Possible solution

Maybe we should stop the follow mode as soon as the following person moves their cursor? Or even when the following person presses *any* key? Then the explicit <Esc> mapping wouldn't be necessary anymore.

## References

How does a plugin like Live Share solve this? I think they also have a follow mode?

/cc @MichaelBitard, I'd be curious to hear your thoughts on this!

MichaelBitard:
fixes #420 

I did not remove the EthersyncUnfollow (but I can if you want to), I thought we could keep it in case I missed something and we ended up in an un-unfollowable condition, but, as far as I tested, it does not happen.

### DIFF
--- a/nvim-plugin/lua/ethersync/cursor.lua
+++ b/nvim-plugin/lua/ethersync/cursor.lua
@@ -248,10 +248,12 @@ function M.track_cursor(bufnr, callback)
     vim.api.nvim_exec_autocmds("CursorMoved", {})
 end
 
-function M.unfollow_cursor()
+local on_key_nsid = nil
+local function unfollow_cursor()
     if following_user_id ~= nil then
         print("Unfollowed cursor")
         following_user_id = nil
+        vim.on_key(nil, on_key_nsid)
     end
 end
 
@@ -325,8 +327,16 @@ end
 function M.follow_cursor()
     local callback = function(user_id)
         print("Following cursor")
+        jump_to_user_id(user_id)
         following_user_id = user_id
-        jump_to_user_id(following_user_id)
+
+        on_key_nsid = vim.on_key(function(_, typed)
+            -- {typed} may be empty if the first argument is produced by non-typed key(s) or by the same typed key(s) that produced a previous {key}
+            -- on_key is triggered even when the followee moves/types, but {typed} is empty in this case
+            if following_user_id ~= nil and typed ~= nil and typed ~= "" then
+                unfollow_cursor()
+            end
+        end)
     end
     pick_cursor_menu(callback)
 end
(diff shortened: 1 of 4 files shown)


nvim-plugin/lua/ethersync/cursor.lua
```diff
@@ -325,8 +330,14 @@ end
 function M.follow_cursor()
     local callback = function(user_id)
         print("Following cursor")
+        jump_to_user_id(user_id)
         following_user_id = user_id
-        jump_to_user_id(following_user_id)
+
+        on_key_nsid = vim.on_key(function(_, typed)
+            if following_user_id ~= nil and typed ~= nil and typed ~= "" then
```
blinry, line 336: Could you maybe add a short comment here with your best understanding of when `typed` can be `nil`, and why it's important to filter those callbacks here? Might help someone reading this snippet in the future! :)

nvim-plugin/lua/ethersync/cursor.lua
```diff
@@ -248,12 +248,17 @@ function M.track_cursor(bufnr, callback)
...
-function M.unfollow_cursor()
+local on_key_nsid = nil
+local function unfollow_cursor()
     if following_user_id ~= nil then
         print("Unfollowed cursor")
         following_user_id = nil
+        vim.on_key(nil, on_key_nsid)
     end
 end
+function M.unfollow_cursor()
```
blinry: You're already anticipating removing the public function here, right? I like it!

blinry (dismissed):
Neat, it works for me, as well! `on_key` even triggers when scrolling or clicking with the mouse, which is great!

One thing that would still be nice is updating the docs: We can remove the `<esc>` mapping from the README, and remove the mention of the function in the Tips section. And maybe add a sentence to the help file in *docs/ethersync.txt*, explaining in the description of `:EthersyncFollow` that any input will stop following!

Thanks for putting this together! \o/

nvim-plugin/lua/ethersync/cursor.lua
blinry: (And, after we've tried it out a bit in practice, I would probably favor doing so!)

nvim-plugin/lua/ethersync/cursor.lua
MichaelBitard: Of course! Done!

nvim-plugin/lua/ethersync/cursor.lua
MichaelBitard: Fine by me! Done!

Issue #138, opened 2024-08-16
Improve the editor protocol


blinry:
This is a tracking issue for collecting things that are currently a bit weird in the daemon-editor protocol. We can add to it as we go, and eventually get to a "v2" of the protocol.

- [ ] To tell an editor that a peer disappeared, you *have* to provide a filename and an empty range. It would be preferable to directly express what we want. Maybe the message should contain a list of (filename + range) combinations after all?

PR #427, opened 2025-10-17
nvim: Implement an teamtype.status() function for status bars

### LINKED ISSUE #417: Neovim: Allow adding info to your status bar
It would be nice to show:

- The connection status
- The number of other cursors
- The follow state

This would be implemented in a way so that people can embed this info in their status bar configuration.

blinry:
This is supposed to show a short status like "Teamtyping with <name> and <name>" in a status bar plugin of your choice.

## Missing things

- [x] Actually check whether the client is still online (should also be added to `:TeamtypeInfo`)
- [x] Don't repeat user names if they have multiple files open
- [x] Maybe mention in which files people are?
- [x] Document

## How to try

This can be used with a lualine config like this (feel free to add the teamtype part in whichever section you want):

```lua
return {
    "nvim-lualine/lualine.nvim",
    dependencies = {
        {
            "nvim-tree/nvim-web-devicons",
        },
    },
    opts = {
        sections = {
            lualine_a = { "mode" },
            lualine_b = { "filename" },
            lualine_c = {
                {
                    function()
                        return require("teamtype").status()
                    end,
                },
            },
            lualine_x = {},
            lualine_y = { "location" },
            lualine_z = { "progress" },
        },
    },
}
```

### DIFF
--- a/nvim-plugin/lua/teamtype.lua
+++ b/nvim-plugin/lua/teamtype.lua
@@ -141,6 +141,16 @@ local function find_or_create_client(config_name, root_dir)
     }
     local the_connection = connection.connect(configurations[config_name].cfg.cmd, root_dir, function(m, p)
         process_operation_for_editor(client, m, p)
+    end,
+    function()
+        -- React to disconnect.
+        for _, c in ipairs(clients) do
+            if c == client then
+                for _, buf_nr in ipairs(c.buffers) do
+                    vim.bo[buf_nr].modifiable = false
+                end
+            end
+        end
     end)
     client.connection = the_connection
     table.insert(clients, client)
--- a/nvim-plugin/lua/teamtype.lua
+++ b/nvim-plugin/lua/teamtype.lua
@@ -141,14 +141,22 @@ local function find_or_create_client(config_name, root_dir)
     }
     local the_connection = connection.connect(configurations[config_name].cfg.cmd, root_dir, function(m, p)
         process_operation_for_editor(client, m, p)
-    end,
-    function()
+    end, function()
         -- React to disconnect.
-        for _, c in ipairs(clients) do
+        local to_remove = {}
+        for i, c in ipairs(clients) do
             if c == client then
                 for _, buf_nr in ipairs(c.buffers) do
                     vim.bo[buf_nr].modifiable = false
                 end
+            end
+            to_remove[i] = true
+        end
+
+        -- Remove from clients list.
+        for i = #clients, 1, -1 do
+            if to_remove[i] then
+                table.remove(clients, i)
             end
         end
     end)
--- a/nvim-plugin/lua/teamtype.lua
+++ b/nvim-plugin/lua/teamtype.lua
@@ -171,7 +171,6 @@ local function find_or_create_client(config_name, root_dir)
                     vim.api.nvim_err_writeln(
                         "Connection to Teamtype daemon lost. Probably it crashed or was stopped. Please restart the daemon, then Neovim."
                     )
-                    -- TODO: Enable writing here again, so that user can make backup of file?
                 end)
             else
                 print(
@@ -186,11 +185,18 @@ local function find_or_create_client(config_name, root_dir)
                 local to_remove = {}
                 for i, c in ipairs(clients) do
                     if c == client then
-                        for _, buf_nr in ipairs(c.buffers) do
-                            vim.bo[buf_nr].modifiable = false
+                        to_remove[i] = true
+
+                        -- If the daemon crashed, prevent modifications to the buffers to save users from
+                        -- accidental data losss.
+                        if code == 0 then
+                            for _, buf_nr in ipairs(c.buffers) do
+                                vim.bo[buf_nr].modifiable = false
+
+                                -- TODO: Enable writing here again, so that user can make backup of file?
+                            end
                         end
                     end
-                    to_remove[i] = true
                 end
 
                 -- Remove from clients list.
--- a/nvim-plugin/lua/teamtype.lua
+++ b/nvim-plugin/lua/teamtype.lua
@@ -156,27 +156,54 @@ local function find_or_create_client(config_name, root_dir)
         buffers = {},
         connection = nil,
     }
-    local the_connection = connection.connect(configurations[config_name].cfg.cmd, root_dir, function(m, p)
-        process_operation_for_editor(client, m, p)
-    end, function()
-        -- React to disconnect.
-        local to_remove = {}
-        for i, c in ipairs(clients) do
-            if c == client then
-                for _, buf_nr in ipairs(c.buffers) do
-                    vim.bo[buf_nr].modifiable = false
-                end
-            end
-            to_remove[i] = true
-        end
 
-        -- Remove from clients list.
-        for i = #clients, 1, -1 do
-            if to_remove[i] then
-                table.remove(clients, i)
+    -- TODO: Encapsulate these dispatchers in the connection module better?
+    local dispatchers = {
+        notification = function(m, p)
+            process_operation_for_editor(client, m, p)
+        end,
+        on_error = function(code, ...)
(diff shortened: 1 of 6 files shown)


alerque (approved):
Force push to

* rebase on current *main* (which is irrelevant to the scope of this PR, just easier to test since it has the convenience job to launch the development plugin version),
* rewrite the commit messages with conventional-commit scopes, and
* expand the docs a tiny bit.

LGTM.

nvim-plugin/lua/teamtype.lua
```diff
@@ -45,6 +45,23 @@ function M.enable(name, enable)
...
+        return ""
+    end
+
+    local result = "Teamtype: "
+
+    local n = cursor.number_of_cursors()
+    if n > 0 then
+        result = result .. cursor.short_cursor_description()
+    else
+        result = result .. "connected"
```
zormit: I find it a bit surprising/confusing to have the string `connected` here, when no one else is connected.

Ideas that could be less confusing: "running" or "waiting for connections" or "no one connected"?

nvim-plugin/lua/teamtype/cursor.lua
```diff
@@ -357,6 +357,48 @@ function M.jump_to_cursor()
...
+                users[data.name] = {}
+            end
+            for _, c in ipairs(data.cursors) do
+                users[data.name][vim.fs.basename(c.uri)] = true
+            end
+        end
+    end
+
+    -- Now, build a string that we can return.
+    local userList = {}
```
zormit: I can never remember, but from what I see around, I think we're using snake_case in Lua?
```suggestion
    local user_list = {}
```

nvim-plugin/lua/teamtype/cursor.lua
```diff
@@ -357,6 +357,48 @@ function M.jump_to_cursor()
...
+
+    for name, files in pairs(users) do
+        local fileList = {}
```
zormit: similarly
```suggestion
        local file_list = {}
```

nvim-plugin/lua/teamtype/cursor.lua
zormit, line 64: This is an _existing_ bug/weird behavior, but the statusline amplified it:

I think we should filter here for our own cursors. If I'm in multiple files with multiple vim instances, I should know where I am and I don't want to be informed about it in the other instances.

nvim-plugin/lua/teamtype.lua
```diff
@@ -139,9 +156,59 @@ local function find_or_create_client(config_name, root_dir)
         buffers = {},
         connection = nil,
     }
-    local the_connection = connection.connect(configurations[config_name].cfg.cmd, root_dir, function(m, p)
-        process_operation_for_editor(client, m, p)
-    end)
+
+    local dispatchers = {
+        notification = function(m, p)
```
zormit, line 162: [nitpick] From a code quality / code abstraction point of view I'm not super happy that the more internal parameter table (`{notification= ..., on_error=..., on_exit=...}`) to `lsp.rpc.start` is now _more exposed_ in a rather generic module. But it's a nitpick and I currently don't have the capacity to think about a better way...

So this is definitely not blocking, but wanted to mention it in case someone stumbles over this during a refactoring session.

nvim-plugin/lua/teamtype/cursor.lua
```diff
@@ -357,6 +357,48 @@ function M.jump_to_cursor()
...
+
+        for file, _ in pairs(files) do
+            table.insert(fileList, file)
+        end
+
+        table.insert(userList, name .. " (" .. table.concat(fileList, ", ") .. ")")
+    end
+
+    return table.concat(userList, ", ")
```
zormit: How does this string behave with a _larger_ number of users or files, does it get shortened by the statusline plugin or is this our job?

zormit (approved):
Thanks, works like a charm. I've left some questions and small ideas for improving the code.

nvim-plugin/lua/teamtype.lua
alerque: Good point. I actually bumped into this in another context too. It will show up in other editors and interfaces as well. We should probably standardize on what this state is called. Some brainstorm ideas:

* Active
* Enabled
* Listening
* Running
* Shared
* Waiting

nvim-plugin/lua/teamtype/cursor.lua
alerque: The results of `rg -t lua '\p{Ll}\p{Lu}'` suggest camelCase appears in a handful of other places but that *most* of the Lua code uses sake_case. I say most because there are a handful of other cases as well including smashcase. I don't see any coherent scheme such as different variable types being styled differently; it seems like just a few camelCase things slipped through.

nvim-plugin/lua/teamtype.lua
blinry: I picked "active" here for now, I feel that expresses the situation pretty well.

nvim-plugin/lua/teamtype/cursor.lua
blinry: But fixed these two (userList and fileList) here.

nvim-plugin/lua/teamtype/cursor.lua
blinry: Interestingly, I think *we don't know who we are*! :D

Meaning that the plugin always sends an "anonymous" cursor to the daemon, and the daemon has the knowledge about the user name.

In the context of Teamtype, this makes the plugin need to do "less work"/be "non-configurable", and I think that's acceptable. When using the plugin for other collaborative projects, maybe it would be nice to be able to configure it in the plugin; I'm not sure.

Added to #138:

- [ ] When sending a username, would it be nice to provide a name along with it? That would move the configurability from the "collaboration backend" (in Teamtype: the daemon) to the plugin.

nvim-plugin/lua/teamtype.lua
blinry: I added a TODO for our future selves. It's noticable that the connection module now doesn't "do very much" anymore, after all.

nvim-plugin/lua/teamtype/cursor.lua
blinry: Tried it (and found an unrelated bug #562 while doing so), and at least lualine abbreviates it:

image

The result is not super helpful, though. Ideas for how to shorten the output:

- Shorten long usernames to a maximum
- Leave out filename if it's the current one
- Cluster users by filename?
- If it's more than three users, maybe just say "n users connected"?

nvim-plugin/lua/teamtype/cursor.lua
alerque: Follow up to catch the other instances (and start a space vs. tabs war :stuck_out_tongue:) here: https://github.com/teamtype/teamtype/pull/563

nvim-plugin/lua/teamtype/cursor.lua
zormit: Ok in that case maybe the daemon can just not send the cursors "locally"? I haven't looked into it what that entails, might be out of scope.

nvim-plugin/lua/teamtype/cursor.lua
zormit: I think all of these are good ideas. Could also be a _feature_ issue and we decide to merge this one without that improvement.

nvim-plugin/lua/teamtype/cursor.lua
alerque: Something to look into fixing for sure, but out of scope and will affect other editors too.

nvim-plugin/lua/teamtype/cursor.lua
blinry: Opened one as #566!

blinry:
Thanks for reviewing, @zormit and @alerque! :) Merged this as a first iteration.

PR #431, opened 2025-10-17
feat(cli): create completions and man page from clap

### LINKED ISSUE #429: Build shell completion files and manpage
See https://github.com/EdJoPaTo/rust-binary-metafile-template for inspiration.

EdJoPaTo:
closes #429

### DIFF
--- a/daemon/integration-tests/Cargo.lock
+++ b/daemon/integration-tests/Cargo.lock
@@ -96,6 +96,56 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.6.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "is_terminal_polyfill",
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle"
+version = "1.0.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


EdJoPaTo:
> I have a few naive questions:

Naive questions are often the best for learning and understanding :innocent: 

> * how to use these files? where would you document this? (README vs [book](https://ethersync.github.io/ethersync/))

I think most people do not care about them, and they are most likely packaged and placed at the right position by maintainers. Then most people can just use completions without caring about where to put the files as installing the package is enough.

> * do we now need to ping package maintainers that CLI completion and man pages are available?

So this is likely what should go into release notes to notify packagers. They know where to put these files correctly depending on the distro.

> * are they usually part of the [release](https://github.com/ethersync/ethersync/blob/main/.github/workflows/release.yml#L43)?

@blinry and I thought about that earlier too. Personally I put them together with the release, but I'm not sure how many people use them. But it's definitely helpful for packaging bin packages (like in the AUR) as the completions don't need to be built then.

Given that the integration tests use a feature to not build everything and this kinda goes around to, @blinry and I thought earlier about maybe putting it somewhere else than a build script. An example does not feel right. Some executables are also able to print their own completions, but I'm not sure whether I like that in comparison to providing that at compile time. The build scripts are also a bit annoying as they need to be executed before the rest of the build (but `cargo:rerun-if-changed` helps with that a bit). So placing that somewhere else would be neat, just not sure where. (Have not figured that out yet myself.) As far as I remember there is an open issue somewhere for this kind of use-case, where it doesn't need to be part of the `build.rs`. Maybe cargo script is gonna be a built-in thing in the future so maybe its better placed there in the future.

zormit:
@alerque I'd be curious what for example your needs would be (thanks, at this occasion, for maintaining the Arch Linux package! =))

blinry:
Thanks again for contributing this, @EdJoPaTo!

PR #441, opened 2025-11-03
README: Add new Arch package and Homebrew package

blinry:


zormit (approved):
double checked => works!

PR #442, opened 2025-11-06
README: Mention nixpkgs teamtype package

blinry:


zormit (approved):
Good catch that it's available now!

PR #443, opened 2025-11-06
Create issue templates

blinry:
Main motivation: We'd often like to know which version of the various software components involved people are using!

PR #445, opened 2025-11-06
daemon: bump version to a development version to differentiate from release

zormit:


daemon/src/cli.rs
```diff
@@ -7,9 +7,25 @@ use std::path::PathBuf;
...
+    let is_dirty = option_env!("GIT_DIRTY").unwrap_or("false") == "true";
+
+    let version = match (git_hash, commit_date) {
+        (Some(hash), Some(date)) if hash != "unknown" && date != "unknown" => {
+            let dirty_suffix = if is_dirty { "-modified" } else { "" };
+            format!("{version} ({hash}{dirty_suffix} {date})")
+        }
+        _ => version.to_string(),
+    };
+    Box::leak(version.into_boxed_str())
```
zormit, line 23: Introducing our first memory leak 😱 (tiny and non-threatening). 

`Box::leak` is _one_ way to get to a `'static str`, but it seems to be a hack...?

Alternatives might be to use [`once_cell`](https://github.com/matklad/once_cell) or [`LazyLock`](https://doc.rust-lang.org/nightly/std/sync/struct.LazyLock.html), I'm not sure yet what I (should) prefer.

daemon/src/daemon.rs
```diff
@@ -975,7 +975,7 @@ impl Daemon {
         }
         if let Some(config::Peer::SecretAddress(ref secret_address)) = app_config.peer {
             connection_manager
-                .connect(secret_address.to_string())
+                .connect(secret_address.clone())
```
zormit, line 978: Unrelated but necessary clippy fix due to [Rust Version update ](https://blog.rust-lang.org/2025/10/30/Rust-1.91.0/)

daemon/Cargo.toml
```diff
@@ -7,7 +7,7 @@
 [package]
 name = "teamtype"
 description = "Enables real-time co-editing of local text files."
-version = "0.9.0"
+version = "0.9.1-dev"
```
zormit, line 10: Motivated by #443, this is the core of this PR :))

daemon/src/cli.rs
blinry: I think it's okay like this; it really seems non-threatening.

daemon/Cargo.toml
blinry: To me, it seems a bit strange (& possibly annoying) having to bump this after each release – especially when we don't know yet whether the next version will be 0.9.1 or 0.10! :O

Could we maybe print the output in a different way?

```
$ teamtype --version
teamtype (b48e6fffa 2025-11-07), last release: 0.9.0
```

Or even dropping the version completely in this case?

How do other programs do that?

daemon/build.rs
```diff
@@ -6,7 +6,42 @@ include!("src/cli.rs");
...
+            .and_then(|output| String::from_utf8(output.stdout).ok())
+            .map_or_else(|| "unknown".to_string(), |s| s.trim().to_string());
+
+        // Check if working directory is dirty
+        let is_dirty = Command::new("git")
+            .args(["status", "--porcelain"])
+            .output()
+            .ok()
+            .filter(|output| output.status.success())
+            .is_some_and(|output| !output.stdout.is_empty());
```
blinry, line 39: What do you think about getting this information via a crate like [shadow-rs](https://github.com/baoyachi/shadow-rs)? I might slightly prefer having "less code", against the downside "more dependencies". But if you have a strong opinion, you can probably convince me otherwise :)

PR #446, opened 2025-11-12
Add optional rendezvous url for magic wormhole

martengooz:
### Summary
This PR enables an optional configuration and command line argument to specify a custom magic wormhole rendezvous / mailbox server. If not configured, teamtype will use the default community rendezvous server. 

To configure a custom mailbox server you can either use the command line argument:
```
teamtype share --rendezvous-url ws://example.com:4000/v1
```
or by adding it to the .teamtype/config file
```
peer=6d02...2f4587
rendezvous_url=ws://example.com:4000/v1
```

### Why this change
This would enable workflows involving sensitive documents / information or with high usage (traffic) to use teamtype without relying on the default community servers - where security or performance considerations might be an issue. 
This is a first step towards being able to self host the whole infra, and looking at adding similar support for Iroh as well. 

### Feedback
I'm open to any feedback you see fit. I'm not a Rust developer and have only been testing out Teamtype for a short while.

### DIFF
--- a/docs/teamtype-cli-ui.md
+++ b/docs/teamtype-cli-ui.md
@@ -16,7 +16,7 @@ This document contains our current idea of how the Commandline User Interface lo
 
 ## Overview
 
-A main inspiration for our UI is a tool called [`Magic Wormhole`](https://magic-wormhole.readthedocs.io/en/latest/welcome.html#example),
+A main inspiration for our UI is a tool called [Magic Wormhole](https://magic-wormhole.readthedocs.io/en/latest/welcome.html#example),
 which uses a random but easily communicable keyword (number-word-word, e.g. 7-crossover-clockwork) to identify a certain transfer.
 
 In our case the keyword identifies a shared project.


daemon/src/cli.rs
```diff
@@ -39,6 +39,9 @@ pub enum Commands {
         /// keep Magic Wormhole out of the loop.
         #[arg(long)]
         no_join_code: bool,
+        /// Use an alternative magic wormhole rendezvous url
```
blinry: For consistency, could we write "Magic Wormhole" capitalized here?

daemon/src/cli.rs
```diff
@@ -39,6 +39,9 @@ pub enum Commands {
...
+        #[arg(long)]
+        rendezvous_url: Option<String>,
```
blinry: My feeling is to clearly state in the name of this flag that this is about Magic Wormhole. Especially because we'll need a similar thing for Iroh later, as you mentioned!

It's a bit long, but what do you think of `magic_wormhole_rendezvous_url`?

daemon/src/cli.rs
```diff
@@ -49,6 +52,9 @@ pub enum Commands {
     Join {
         /// Specify to connect to a new peer. Otherwise, try to connect to the most recent peer.
         join_code: Option<String>,
+        /// Use an alternative magic wormhole rendezvous url
+        #[arg(long)]
+        rendezvous_url: Option<String>,
```
blinry: To avoid duplicating this here, it'd be nice to put the `sync_vcs` flag and this one into a struct, which is then *flattened* into both the share and join subcommands. To do this, we could rename the `SyncVcsFlag` into `SharedFlags`, for example.

daemon/src/wormhole.rs
```diff
@@ -5,12 +5,21 @@
...
 use magic_wormhole::{transfer, AppID, Code, MailboxConnection, Wormhole};
-use std::{str::FromStr, time::Duration};
+use std::{borrow::Cow, str::FromStr, time::Duration};
 use tokio::time::sleep;
 use tracing::{error, info, warn};
 
-pub async fn put_secret_address_into_wormhole(address: &str) {
-    let config = transfer::APP_CONFIG.id(AppID::new("teamtype"));
+pub async fn put_secret_address_into_wormhole(address: &str, rendezvous_url: Option<String>) {
+    let config: magic_wormhole::AppConfig<transfer::AppVersion> = if let Some(url) = rendezvous_url
```
blinry: Instead of letting the config value come out of the `if`, I'd prefer it here to make the config `mut`, and calling the `rendezvous_url` on it in case we need to. This avoids the duplicated construction of the `AppId`.

daemon/src/wormhole.rs
```diff
@@ -41,8 +50,19 @@ pub async fn put_secret_address_into_wormhole(address: &str) {
...
+) -> Result<String> {
+    let config: magic_wormhole::AppConfig<transfer::AppVersion> = if let Some(url) = rendezvous_url
+    {
+        info!("Using magic wormhole rendezvous url {}", url);
+        transfer::APP_CONFIG
+            .id(AppID::new("teamtype"))
+            .rendezvous_url(Cow::Owned(url))
+    } else {
+        transfer::APP_CONFIG.id(AppID::new("teamtype"))
+    };
```
blinry: This code is duplicated from the `put_into_...` function. Could we put it into a shared function, like `build_magic_wormhole_config(rencezvous_url)`?

blinry (changes requested):
Hey @martengooz! Thanks a lot for this contribution! I completely agree that this seems like a required feature for more privacy-sensitive scenarios! And implementing custom servers one by one seems like a good idea.

Props to you to build this without a lot of Rust experience! It looks pretty good, but I left a couple of ideas as review comments.

We run clippy in our CI to keep the code quality up – you can check its suggestions yourself by running `cargo clippy --all-targets`.

Also, it'd be great to check if this actually works with self-hosted rendezvous servers, have you done that already, by any chance?

martengooz:
> Hey @martengooz! Thanks a lot for this contribution! I completely agree that this seems like a required feature for more privacy-sensitive scenarios! And implementing custom servers one by one seems like a good idea.
> 
> Props to you to build this without a lot of Rust experience! It looks pretty good, but I left a couple of ideas as review comments.
> 
> We run clippy in our CI to keep the code quality up – you can check its suggestions yourself by running `cargo clippy --all-targets`.
> 
> Also, it'd be great to check if this actually works with self-hosted rendezvous servers, have you done that already, by any chance?

Thanks! I noticed the failing builds as well, I’ll check on the comments and apply the suggestions from clippy. Good to know that I can run it myself before submitting :)

And yes, I have tested with a self hosted server from the latest build, and I could see the connection coming in and working fine.

daemon/src/wormhole.rs
martengooz: This is now refactored to use a shared private function `build_magic_wormhole_config(rencezvous_url)`

daemon/src/wormhole.rs
martengooz: Handled in the new `build_magic_wormhole_config(rencezvous_url)`

daemon/src/cli.rs
martengooz: I tried this out but felt like the name `SharedFlags` was a bit vague and could cause issues if more commands are added. Will all (new) commands have the same shared flags - maybe, maybe not? Also, if more configuration is added for magic wormhole, this might not be used in both `Join` and `Share` commands. Ex. if we add the option to choose the length of the generated code by providing `code_length` in  `MailboxConnection::create(config, code_length)`. 

I added commit [8affc68](https://github.com/teamtype/teamtype/pull/446/commits/8affc68d46754a5baa39d0234325022161ecdf74) to address this. Similar struct could be done for Iroh config later if you think it's a good idea. 

Let me know your thoughts on this, I'll rebase the commits based on your opinion.

daemon/src/cli.rs
martengooz: Renamed it to your suggestion

zormit: I also have some opinions on the naming :) (especially since @blinry asked me what I think about the long name). I do think `magic_wormhole_rendezvous_url` is a bit unwieldy as a CLI parameter. Also I guess a lot of people that are not speaking French would trip over the spelling of `rendezvous`.

Magic Wormhole itself [calls the parameter](https://magic-wormhole.readthedocs.io/en/latest/welcome.html#relays) `--relay-url`. I think we could take that, shorten "Magic Wormhole" to "wormhole", then we would have `--wormhole-relay-url`.
- it'll be different from a potential `--iroh-relay-url`
- relay is easier than rendezvous
- it's consistent, for whatever that's worth, with MW itself. (edit: Although I realized that in the _codebase_ they're [using "rendezvous"](https://docs.rs/magic-wormhole/latest/magic_wormhole/struct.AppConfig.html#method.rendezvous_url), which is probably what you got inspired by ¯\_(ツ)_/¯)

Different options might be:
- `--magic-wormhole-relay`, we could drop URL and add the Magic
- `--wormhole-relay`, or drop URL without Magic to shorten even more
- `--join-code-relay[-url]`, or go away from Magic Wormhole and towards our own naming of what goes over that relay. I just got that idea, but I suppose people who care to use a different relay will have an easier time remembering the fact that it's a Magic Wormhole relay.

daemon/src/cli.rs
zormit: Ok, while we're here, and this is not on you, @martengooz, at all: We could make it consistent all the way through. I just searched for a lowercase "magic wormhole" in the codebase:
- https://github.com/teamtype/teamtype/blob/20270b5aee5565c3b9283fafc1bdca1181c60181/docs/teamtype-cli-ui.md?plain=1#L19 this is even completely inconsistent, because it looks like a command but is missing the `-` in it. Let's just put "Magic Wormhole" there instead of "`magic wormhole`".
- https://github.com/teamtype/teamtype/blob/20270b5aee5565c3b9283fafc1bdca1181c60181/daemon/src/wormhole.rs#L20

I thought it's tedious that you have to fix our legacy so I just pushed it myself in [3996298](https://github.com/teamtype/teamtype/pull/446/commits/3996298b072c561630cee12673dadef09443fadb) :)

daemon/src/daemon.rs
```diff
@@ -971,11 +971,15 @@ impl Daemon {
...
+            put_secret_address_into_wormhole(
+                address,
+                app_config.magic_wormhole_rendezvous_url.clone(),
+            )
+            .await;
         }
         if let Some(config::Peer::SecretAddress(ref secret_address)) = app_config.peer {
             connection_manager
-                .connect(secret_address.to_string())
+                .connect(secret_address.clone())
```
zormit: This change will go away if you rebase :) I had already fixed it in b48e6fffae. Just dropping the commit in case you were wondering why you had to fix it: Due to a new Rust stable release.

daemon/src/wormhole.rs
```diff
@@ -41,12 +41,25 @@ pub async fn put_secret_address_into_wormhole(address: &str) {
...
             .await?;
     let bytes = wormhole.receive().await?;
     Ok(String::from_utf8(bytes)?)
 }
+
+fn build_magic_wormhole_config(rendezvous_url: Option<String>) -> AppConfig<transfer::AppVersion> {
+    let mut config = transfer::APP_CONFIG.id(AppID::new("teamtype"));
+
+    if let Some(url) = rendezvous_url {
+        info!("Using rendezvous url {}", url);
```
zormit: This info could benefit from some "Magic Wormhole" contextualization. Like 
```suggestion
        info!("Using non-default Magic Wormhole rendezvous url {}", url);
```

And maybe replacing "rendezvous" by "relay" as discussed in the other thread.

zormit (commented):
Looks good to me :) I found one place worth changing.

daemon/src/cli.rs
martengooz: Of the alternatives I personally like `--magic-wormhole-relay` best
* keeping the name "Magic Wormhole" allows for easy googling if needed instead of just "wormhole"
* skipping the "url" part could be done, since it's implicit from the usage. 

Then it would be `--iroh-relay` 

But maybe we should specify that it's the mailbox-server and not the [transit-relay](https://github.com/magic-wormhole/magic-wormhole-transit-relay). Would it suffice to do that in the help text related to the flag?

But it's up to you guys, either way works for me :)

martengooz:
Added additional docs on how to use and configure the feature

book/src/connection-making.md
```diff
@@ -11,6 +11,11 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 
 When you run `teamtype share`, you will get a short "join code" like `3-exhausted-bananas`. Another person can use it to connect to you! The code only works once. You can learn about the security properties in the [Magic Wormhole documentation](https://magic-wormhole.readthedocs.io/en/latest/welcome.html#safely).
 
+You can make Teamtype use a custom [relay server](https://github.com/magic-wormhole/magic-wormhole-mailbox-server) with the optional `--magic-wormhole-relay` parameter or by adding it to your (configuration file)[configuration.md].
```
zormit:
```suggestion
You can make Teamtype use a custom [relay server](https://github.com/magic-wormhole/magic-wormhole-mailbox-server) with the optional `--magic-wormhole-relay` parameter or by adding it to your [configuration file](configuration.md).
```
typo link syntax

docs/teamtype-cli-ui.md
```diff
@@ -16,7 +16,7 @@ This document contains our current idea of how the Commandline User Interface lo
 
 ## Overview
 
-A main inspiration for our UI is a tool called [`magic wormhole`](https://magic-wormhole.readthedocs.io/en/latest/welcome.html#example),
+A main inspiration for our UI is a tool called [`Magic Wormhole`](https://magic-wormhole.readthedocs.io/en/latest/welcome.html#example),
```
zormit:
```suggestion
A main inspiration for our UI is a tool called [Magic Wormhole](https://magic-wormhole.readthedocs.io/en/latest/welcome.html#example),
```
another tiny fix (I had _intended_ to fix this earlier, lol)

daemon/src/cli.rs
zormit: I don't have a strong opinion for myself, but I can imagine `--magic-wormhole-relay` seems good enough for users :) "handing" this PR back to @blinry to finish it =)

zormit:
btw, for testing this, one can use either the default relay, but pass it with the flag:

`teamtype share --wormhole-relay-url ws://relay.magic-wormhole.io:4000/v1`

An alternative is using this one by Least Authority (found [here](https://github.com/magic-wormhole/magic-wormhole/issues/243#issuecomment-1483201907)):
`teamtype share --wormhole-relay-url ws://mailbox.mw.leastauthority.com/v1`

daemon/src/cli.rs
blinry: I tried to find more info about the two types of relays. The terminology is a bit confusing:

1. https://github.com/magic-wormhole/magic-wormhole-mailbox-server
    - This is what we're setting here
    - Used like this: `wormhole send --relay-url=ws://localhost:4000/v1`
    - Set with [`rendezvous_url()`](https://docs.rs/magic-wormhole/latest/magic_wormhole/struct.AppConfig.html#method.rendezvous_url) in Rust
	- Documented as ["Mailbox Server" (mostly) or "Rendezvous Server"](https://magic-wormhole.readthedocs.io/en/latest/welcome.html#relays)
	- The public one is `ws://relay.magic-wormhole.io:4000/v1`
2. https://github.com/magic-wormhole/magic-wormhole-transit-relay
    - Has "relay" in its name…
    - Used like this: `wormhole send --transit-helper=tcp:[server ip here]:4001`
    - Set via a [`relays` parameter](https://docs.rs/magic-wormhole/latest/magic_wormhole/transit/struct.Hints.html#impl-Hints) in Rust
    - Documented as ["Transit Relay"](https://magic-wormhole.readthedocs.io/en/latest/transit.html#relay)
    - The default one is `tcp://transit.magic-wormhole.io:4001`
    - Seems only be used for "large-volume data transfer", according to [the docs](https://magic-wormhole.readthedocs.io/en/latest/file-transfer-protocol.html#transit)?

The last point seems the most important to me: Because we only use Magic Wormhole to exchange short text messages, we probably will never need the second kind of server.

So I'm also happy with `--magic-wormhole-relay` :) The documentation of the parameter (and in the "book") makes it clear which kind of relay it is!

PR #458, opened 2025-11-27
nvim: *Really* ensure that a file exists before we "open" it

### LINKED ISSUE #92: neovim complains about "file has been changed on disk" for newly created one
bug probably introduced in #85

### LINKED ISSUE #457: nvim: "Warning: File "..." has been created after editing started"
**Describe the bug**
When I create a new file with Neovim, the file will be created by the daemon too.

**To Reproduce**
I think this should do:
- `:e new-file` while Neovim is connected to daemon.
- move away focus from and come back to Neovim

**Expected behavior**
It should not happen

**Versions:**
latest `main` (7a8935b0e)

**Additional context**

Image

blinry:
Turns out that 0 is truthy in Lua...

Fixes #457, and *really* fixes #92.

### DIFF
--- a/nvim-plugin/lua/teamtype.lua
+++ b/nvim-plugin/lua/teamtype.lua
@@ -167,7 +167,7 @@ local function activate_config_for_buffer(config_name, buf_nr, root_dir)
     -- Ensure that the file exists on disk before we "open" it in the daemon,
     -- to prevent a warning that the file has been created externally (W13).
     -- This resolves issue #92.
-    if string.find(uri, "file://") == 1 and not vim.fn.filereadable(filename) then
+    if string.find(uri, "file://") == 1 and vim.fn.filereadable(filename) == 0 then
         vim.cmd("silent write")
     end
 


m10090:
So, I think this will work in most cases. But in one scenario — and correct me if I’m wrong — this could lead to a race condition. If I open the file in Neovim, edit it, and save, it’s possible that Neovim could save the file before the daemon does because the daemon was slower. I had a similar issue where it took around 30 seconds to load a file in TeamType (debug build).

I’ll do some feature testing to confirm, but one option is: if the file doesn’t exist in the directory and needs to be created, let Neovim create it and don’t send the open request to the daemon. Alternatively, force all editors to create the file and remove the creation step from the daemon entirely.
And again than you for letting me in.

blinry:
We don't quite understand the race condition you're describing – can you provide a step-by-step instruction to reproduce it?

We'll merge this PR for now, as we think it's an improvement, but we can iterate more on it later!

PR #466, opened 2025-12-10
Allow configuring username show at cursor

zormit:
Introduces a --username command-line flag and the option to configure
the username via .teamtype/config, in order to have more flexibility on
a per project level without having to rely on Git.

Also comes with some small refactors of the code touched, e.g. adding a
consolidated ShareJoinFlags struct in the CLI code.

- [x] the fallback to `"Anonymous"` doesn't work yet, because `get_git_username` will produce an empty string instead of `None`
- [x] update documentation
- [x] add to changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ SPDX-FileCopyrightText: 2024 zormit <nt4u@kpvn.de>
 SPDX-License-Identifier: CC-BY-SA-4.0
 -->
 
+# 0.9.1 (unreleased)
+
+## Make the username configurable
+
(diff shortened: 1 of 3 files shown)


zormit:
@cosmicexplorer probably out of scope of this PR, but related: I'd be curious what your thoughts are how to approach configuration stuff in general. I recently saw a longer [mastodon thread of yours](https://circumstances.run/@hipsterelectron/115668914443407176). I bookmarked it but haven't fully processed/understood the implications yet 😅. Still, I thought you might have opinions/tipps on how to approach this.

It's a bit quirky in the code how we have a `.teamtype/config` but then we have to get overwriting with command-line-flags correct while providing some sane defaults for each value.

PR #467, opened 2025-12-19
Refactored structure to use more dynamic types

d0mmi:
(WIP)

### DIFF
--- a/daemon/Cargo.lock
+++ b/daemon/Cargo.lock
@@ -4354,6 +4354,7 @@ dependencies = [
  "constant_time_eq",
  "derive_more 2.1.1",
  "dissimilar",
+ "dyn-clone",
  "futures",
  "git2",
  "ignore",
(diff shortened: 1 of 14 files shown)


alerque:
If this is a WIP, marking it as a draft will help the GitHub UI surface the right information for reviewers.

Also two other minor notes:

* Unless the dependencies are changing on purpose, I highly recommend not mixing lockfile updates with other refactoring work. If they do need to get bumped for some reason, do it in a separate commit prior to the actual refactoring so they can be evaluated on their own merits and cherry picked without conflicts.
* Is this all hand editing or is there some LLM output mixed in?

zormit:
I was in touch / in a call with @d0mmi about this. If I remember correctly it isn't generated code, but LLMs might have been involved in an advisory way. The idea was that we integrate the changes step by step, where #490 was the first one that we paired on.

This one is kind of open as a reminder and a source for inspiration, so maybe it could even be closed as there's currently no intention to merge it by itself... Not sure. Turning into a draft one makes sense for sure.

(FWIW, I have some non-public notes that I could post here from when @blinry and I talked about this, but I'd have to translate & extend them to make sense, so I won't do that now)

d0mmi:
Yes the code is not ai generated. I only used ai to ask questions about Rust because this is the first Rust project i worked on.

PR #470, opened 2026-01-13
Add option to use temporary directory as shared directory

### LINKED ISSUE #463: Join automatic tmp folder
**Is your feature request related to a problem? Please describe.**
When joining someone (or some other device of mine) I want to join in a temporary folder which is only relevant for the share. The original stays with the sharing device anyway.

For this I basically used this: `cd ~/tmp && mkdir something && cd something && teamtype join 1-2-3`

This feels like a common enough workflow to have it built-in?

**Describe the solution you'd like**

Maybe a command like `teamtype join --tmpdir 1-2-3` could be neat. This might automatically get the name of the share from the host and create a folder in some (configurable?) place like ~/.cache/teamtype/workspaces/$sharename` and enters it. That way I can quickly join in and team-type on the given share.

Maybe `--tmpdir` without argument uses some default, maybe `--tmpdir=/tmp/foo/bar` accepts a path? Not sure whether that might work well with clap env. Setting `TEAMTYPE_TMP_WORKSPACES=$HOME/.cache/teamtype/workspaces` could be an alternative to a config file or something like that?

axelmartensson:
Add global `--use-temporary-directory` command-line option, which allows you to create a new temporary directory and use it as the shared directory. The directory is removed on exit.

Fixes #463

### DIFF
--- a/daemon/src/cli.rs
+++ b/daemon/src/cli.rs
@@ -1,5 +1,6 @@
 // SPDX-FileCopyrightText: 2025 blinry <mail@blinry.org>
 // SPDX-FileCopyrightText: 2025 zormit <nt4u@kpvn.de>
+// SPDX-FileCopyrightText: 2026 axelmartensson <axel.martensson@hotmail.com>
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -1,5 +1,6 @@
 // SPDX-FileCopyrightText: 2024 blinry <mail@blinry.org>
 // SPDX-FileCopyrightText: 2024 zormit <nt4u@kpvn.de>
+// SPDX-FileCopyrightText: 2026 axelmartensson <axel.martensson@hotmail.com>
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 
 Add global `--temporary-directory` command-line option, which allows you to create a new temporary directory and use it as the shared directory. The temporary directory is removed on exit. The temporary directory is created in `$XDG_CACHE_DIR/teamtype/`, or if `XDG_CACHE_DIR` is not set in `$HOME/.cache/teamtype/`, or if `HOME` is not set in `/home/$USER/.cache/teamtype/`.
 
+Thanks to @axelmartensson for taking this on.
+
 ## Make the username configurable
 
 Teamtype uses the configured Git username by default, to display next to the cursors other people see.
--- a/daemon/src/sandbox.rs
+++ b/daemon/src/sandbox.rs
@@ -1,5 +1,6 @@
 // SPDX-FileCopyrightText: 2024 blinry <mail@blinry.org>
 // SPDX-FileCopyrightText: 2024 zormit <nt4u@kpvn.de>
+// SPDX-FileCopyrightText: 2026 axelmartensson <axel.martensson@hotmail.com>
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 


daemon/src/cli.rs
```diff
@@ -33,6 +33,11 @@ pub struct Cli {
     /// The shared directory. Defaults to current directory.
     #[arg(long, global = true)]
     pub directory: Option<PathBuf>,
+    /// Create a new temporary directory  and use it as the shared directory.
```
zormit:
```suggestion
    /// Create a new temporary directory and use it as the shared directory.
```

daemon/src/cli.rs
```diff
@@ -33,6 +33,11 @@ pub struct Cli {
...
+    /// The temporary directory is removed on exit.
+    /// The temporary directory is created in `$XDG_CACHE_DIR/teamtype/`, or if `XDG_CACHE_DIR` is not set in `$HOME/.cache/teamtype/`, or if `HOME` is not set in `/home/$USER/.cache/teamtype/`.
+    #[arg(long)]
```
zormit: It would be nice to have a `-t` for the "power user" as well as it's quite long :)

daemon/src/cli.rs
```diff
@@ -33,6 +33,11 @@ pub struct Cli {
...
+    pub use_temporary_directory: bool,
```
zormit: As this feature is only used in `teamtype join` or `teamtype share`, it should rather be part of `ShareJoinFlags`

daemon/src/main.rs
```diff
@@ -49,7 +51,35 @@ async fn main() -> Result<()> {
...
+                "Failed to get parent directory of the directory {}",
+                app_cache_dir.display()
+            )
+        })?;
+        let has_app_cache_dir = sandbox::exists(app_cache_dir_parent, &app_cache_dir)
+            .expect("Failed to check")
+            && app_cache_dir.is_dir();
+        if !has_app_cache_dir {
+            sandbox::create_dir(app_cache_dir_parent, &app_cache_dir)?;
+        }
```
zormit: We would prefer the functionality to ensure the existence of the directory (L58-71) is integrated in `sandbox::create_dir`, s.t. that method does not fail if a directory already exists.

(this is non-blocking and we can also refactor it after this PR is finished)

daemon/src/main.rs
```diff
@@ -49,7 +51,35 @@ async fn main() -> Result<()> {
...
+            format!(
+                "Failed to create a temporary directory in the directory {}",
+                app_cache_dir.display()
+            )
+        })?;
+        Some(temporary_directory.path().to_path_buf())
+    } else {
+        cli.directory
+    };
+    let directory = get_directory(directory).context("Failed to find .teamtype/ directory")?;
```
zormit: Maybe it'll be possible to move all the new code of this block _into_ `get_directory`, s.t. the main doesn't grow (even though admittedly `get_directory` is already long as well, but we'd be curious what you think is best in terms of clean code)

daemon/src/main.rs
zormit: Our preferred user-flow would be that this question doesn't get asked in the case of a temporary directory, as it would rather be surprising if `.teamtype` existed.

Note: It'll then be necessary to print out the path to the temporary directory, so the user knows on `teamtype join / share --use-temporary-directory` where to open the editor.

daemon/src/cli.rs
axelmartensson: Agree, fixed in latest update

daemon/src/cli.rs
axelmartensson: fixed in latest update

daemon/src/main.rs
axelmartensson: fixed in latest update

daemon/src/main.rs
axelmartensson: Fixed in latest update.

I see what you mean :slightly_smiling_face: I moved most of the new code into a new method `get_temporary_directory()` as we still need to keep the reference to `temporary_directory` in the main method, because the temporary directory on disk gets deleted when there are no more references pointing to it.

daemon/src/main.rs
axelmartensson: fixed in the latest update

daemon/src/cli.rs
axelmartensson: fixed in the latest update

daemon/src/cli.rs
axelmartensson: I also updated the long form name of the flag from `--use-temporary-directory` to `--temporary-directory` so that it starts with the same letter as the short form of the flag `-t`

axelmartensson:
:+1: looks good to me

axelmartensson:
I am totally OK with not being mentioned in the copyright text. For reference, this is how the GNOME project handles copyright notices: https://handbook.gnome.org/development/legal.html

PR #471, opened 2026-01-16
Add log message in daemon to inform user which username others will see

### LINKED ISSUE #322: Inform user in the daemon which name others will see
In #319 we implemented that the git username will be picked up. This is probably fine but still somewhat surprising, we should let the user know, what name we picked up and how to change it.

### LINKED ISSUE #469: Git username is not detected, if config file is missing
**Describe the bug**
When config file is missing, the git fallback won't work.

**To Reproduce**
Start the daemon in a project without config. Username will be `None`.

**Expected behavior**
Username should fall back to git.

**Versions:**
fc6d8e7286962

**Additional context**
This bug was detected via #468 and was most probably introduced in #459.

axelmartensson:
This displays a log message to inform the user about which username others will see displayed next to the cursor, for example if my Git username is `John Doe`, then the log message is:
```
Using the Git username 'John Doe' as username, to display next to the cursors other people see.
Teamtype uses the Git username as username by default.
You can set the configuration value `username` in your `.teamtype/config` to override this username.
You can also use the flag `--username` when using the `share`/`join` subcommands.
```

fixes #322 , fixes #469

### DIFF
--- a/daemon/src/config.rs
+++ b/daemon/src/config.rs
@@ -46,7 +46,6 @@ pub struct AppConfig {
 }
 
 impl AppConfig {
-    #[must_use = "need to use function from_config_file_and_cli()"]
     // Merges the app config from file with the CLI app config by taking the "superset" of them.
     //
     // It depends on the attribute how we're merging it:


daemon/src/config.rs
```diff
@@ -132,27 +142,6 @@ impl AppConfig {
     pub const fn is_host(&self) -> bool {
         self.peer.is_none()
     }
-
```
zormit, line 135: So it seems you decided to get rid of merge - I don't fully understand everything yet, but I would assume it's because of #469?

daemon/src/config.rs
```diff
@@ -197,6 +186,67 @@ pub fn has_local_user_config(base_dir: &Path) -> Result<bool> {
     Ok(false)
 }
 
+fn get_username(
+    app_config_cli_username: Option<String>,
+    base_dir: &Path,
+    general_section: &Properties,
+) -> Option<String> {
+    app_config_cli_username
```
zormit, line 194: First impression: This chain is a bit long and not so easily readable. I think it should be split into mulitple steps/lines. But maybe we don't need it anymore if we use a library as mentioned above.

zormit (commented):
Thanks for tackling this! I just had a quick look at it, no full review yet.

We were actually thinking of changing this configuration code to maybe use a configuration reader library, currently my favorite is https://lib.rs/crates/config, but there are [many options](https://lib.rs/crates/config) and I didn't thoroughly evaluate their pros and cons yet. Would you be willing to give it a shot to rewrite it? It's probably a larger issue than you signed up for so no worries if not.

daemon/src/config.rs
axelmartensson: Yes, this MR fixes that issue too, I added that issue with a "fixes" comment to the MR description above so that it is linked to this MR.

daemon/src/config.rs
axelmartensson: Yes, I see what you mean, the chain is not easy to read because it is so long. I refactored it a bit so that it is a bit easier to overview. I factored out the code inside the chain into `get_username_from_*` functions (I also removed the superfluous `\n` at the end of the last line of the log messages, because the `info!()` macro already adds a newline at the end of the log message). Let me know what you think.

I looked into using the configuration reader library https://lib.rs/crates/config . It looks very cool, but I am not sure if the juice is (currently) worth the squeeze :slightly_smiling_face:.

Assuming that we want to provide different log messages depending on where the `username` config value comes from, the problem I ran into was how to trace the origin of where the `username` config value comes from.

The `username` config value can come from the `.teamtype/config` file, which is supported out of the box,  but I am not sure how to handle when the `username` config value comes from CLI or from Git. One could maybe use a "custom file format" (example: https://github.com/rust-cli/config-rs/tree/main/examples/custom_file_format ) or construct a "fake" file with `File::from_str`(https://docs.rs/config/latest/src/config/file/mod.rs.html#45-51)

daemon/src/config.rs
```diff
@@ -197,6 +186,80 @@ pub fn has_local_user_config(base_dir: &Path) -> Result<bool> {
...
+        .or_else(|| get_username_from_git(base_dir))
+        .or_else(get_username_from_fallback_value)
+}
+
+fn get_username_from_cli(username: String) -> String {
+    info!(
+        concat!(
+            "Using the username '{}' from the flag `--username` ",
+            "from when using the `share`/`join` subcommands as username, ",
+            "to display next to the cursors other people see."
```
blinry: We think in this case we could shorten the message to:

```suggestion
            "Using the username '{}' to display next to the cursors other people see."
```

daemon/src/config.rs
```diff
@@ -197,6 +186,80 @@ pub fn has_local_user_config(base_dir: &Path) -> Result<bool> {
...
+    if let Some(ref username) = username {
+        info!(concat!(
+            "Using the Git username '{}' as username, ",
+            "to display next to the cursors other people see.\n{}"),
+            username,
+            concat!(
+                "Teamtype uses the Git username as username by default.\n",
+                "You can set the configuration value `username` in your `.teamtype/config` to override this username.\n",
+                "You can also use the flag `--username` when using the `share`/`join` subcommands."),
+        );
```
blinry: Style nit: What do you think about avoiding `concat!` here by using `info!` multiple times. Possibly in some of the other functions, as well.

```suggestion
        info!("Using the Git username '{username}' as username, to display next to the cursors other people see.");
        info!("Teamtype uses the Git username as username by default.");
        info!("You can set the configuration value `username` in your `.teamtype/config` to override this username.");
        info!("You can also use the flag `--username` when using the `share`/`join` subcommands.");
```

daemon/src/config.rs
```diff
@@ -197,6 +186,80 @@ pub fn has_local_user_config(base_dir: &Path) -> Result<bool> {
...
+    }
+    username
+}
+
+fn get_username_from_fallback_value() -> Option<String> {
+    let username = Some(USERNAME_FALLBACK.to_string());
+    if let Some(ref username) = username {
```
blinry: Let's directly output our text here, and then return `Some(username)`, instead of having the `if let`.

daemon/src/config.rs
```diff
@@ -197,6 +186,80 @@ pub fn has_local_user_config(base_dir: &Path) -> Result<bool> {
...
+        info!(concat!(
+            "Using the fallback value for username '{}' as username, ",
+            "to display next to the cursors other people see.\n{}"),
+            username,
+            concat!(
+                "Teamtype uses the Git username as username by default.\n",
+                "Using the fallback value instead can happen if Git is not installed,\n",
+                "or if Git is installed, but no username is set on any level of Git configuration.\n",
```
blinry: Our wish would be to leave out this information about Git entirely here. People who don't have Git or a configured Git username will probably not care about this, and it makes the output pretty long.

daemon/src/config.rs
```diff
@@ -45,51 +45,61 @@ pub struct AppConfig {
 }
 
 impl AppConfig {
-    #[must_use]
-    // Note that this reads the config and has some fallbacks, if the config value doesn't exist.
-    // These fallbacks only get used by the calling main, if there's no value with higher
-    // precedence (e.g. CLI-provided).
-    pub fn from_config_file(base_dir: &Path) -> Option<Self> {
+    #[must_use = "need to use function from_config_file()"]
```
blinry: This note doesn't seem necessary, let's leave it out.

daemon/src/config.rs
```diff
@@ -45,51 +45,61 @@ pub struct AppConfig {
...
+    // Merges the app config from file with the CLI app config by taking the "superset" of them.
+    //
+    // It depends on the attribute how we're merging it:
+    // - For strings, the CLI app config attribute has precedence.
+    // - For booleans, if a value deviates from the default, it "wins".
+    // - The `base_dir` will be taken from the CLI app config.
+    pub fn from_config_file(app_config_cli: Self) -> Self {
```
blinry: Name suggestion: `from_config_file_and_cli`, to clarify what it's doing.

daemon/src/config.rs
axelmartensson: fixed

daemon/src/config.rs
axelmartensson: fixed

daemon/src/config.rs
axelmartensson: fixed (clippy gives a warning about "unneccessary wraps" if I return `Some(username)`, so I return `String` instead of `Option<String>` from this function, and put the wrap of `Some()` around the function call instead)

daemon/src/config.rs
axelmartensson: fixed

daemon/src/config.rs
axelmartensson: I tried removing the note, but clippy gives an error (<https://github.com/teamtype/teamtype/actions/runs/21381598059/job/61549409014?pr=471#step:5:92>):

```
error: this function has a `#[must_use]` attribute with no message, but returns a type already marked as `#[must_use]`
  --> src/config.rs:56:5
   |
56 |     pub fn from_config_file_and_cli(app_config_cli: Self) -> Self {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: either add some descriptive message or remove the attribute
   = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.93.0/index.html#double_must_use
   = note: `-D clippy::double-must-use` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::double_must_use)]`
```

So, I put the note back in, for now

daemon/src/config.rs
axelmartensson: fixed

daemon/src/config.rs
blinry: Good catch! We refactored it a tiny bit more, so that `get_username_from_fallback_value()` always returns a `String`.

daemon/src/config.rs
blinry: We noticed that `AppConfig` was already `#[must_use]` itself, so it's unnecessary nere. We removed it.

Issue #280, opened 2025-06-20
tracking issue: dependency upgrades that are held back


zormit:
This issue is a reminder which crates are not fully upgraded and what blocks us from doing so
- [x] Keeping `constant_time_eq` at version `0.3.1` as described in f01ed49.
  - Once https://github.com/n0-computer/iroh/issues/3353 is resolved, we can try upgrading it.
- [ ] Keeping `rand` pinned to version `0.8.5`, as `automerge` version 1.0.0.beta-3 requires that.

zormit:
Update on the first one: As described in the linked issue I can actually produce working release builds with the Rust Beta or Nightly channels. That should be good enough. So I've opened #283 where I'm bumping `constant_time_eq`, as well as Automerge (let's see if we can get that one working).

zormit:
The former checkbox was/will be resolved in #476, for the latter we are now stuck with `automerge` instead of `iroh`.

alerque:
The info in this issue is obsolete, and notes about the current state of held back crates is in the Cargo manifest.

Suggest closing.

PR #476, opened 2026-01-30
update rust version and package dependencies

zormit:
Outcome of all this: We're more up to date than before. We're still waiting for automerge to take on the newest `rand` as described in #280.
 Not sure how to measure the number of dependencies correctly, but here's a heuristics: The tree shrunk! 🎉 

```
mn@u daemon on branch dependency-updates
% cargo tree | wc -l
    1304
mn@u daemon on branch main
% cargo tree | wc -l
    1334
```

The versions we're still "stuck" with:
```
% cargo update --verbose
    Updating crates.io index
     Locking 0 packages to latest Rust 1.91.0 compatible versions
   Unchanged generic-array v0.14.7 (available: v0.14.9)
   Unchanged hexane v0.1.6 (available: v0.1.7)
   Unchanged iroh v0.92.0 (available: v0.96.0)
   Unchanged rand v0.8.5 (available: v0.9.2)
```
The only one I'm not sure why it's not fully upgraded is `generic-array`, all the other ones are pinned (edit: it seems that 0.14.8 and .9 are deprecated. Not sure why we don't go automatically to 1.x.x then, but I guess it's alright).

### DIFF
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -80,7 +80,7 @@ dissimilar = "1.0.9"
 futures = { version = "0.3.30", default-features = false }
 git2 = { version = "0.20.2", default-features = false }
 ignore = "0.4.22"
-iroh = "0.92.0"
+iroh = "0.96.0"
 magic-wormhole = "0.7.6"
 microxdg = "0.2.0"
 notify = { version = "8.0.0", default-features = false, features = [
@@ -89,7 +89,7 @@ notify = { version = "8.0.0", default-features = false, features = [
 operational-transform = "0.6.1"
 path-clean = "1.0.1"
 postcard = { version = "1.1.1", features = ["alloc"] }
-rand = "0.8.5"
+rand = "0.9.2"
 ropey = "1.6.1"
 rust-ini = "0.21.1"
 serde = { version = "1.0.198", features = ["derive"] }
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -74,12 +74,13 @@ async-trait = "0.1.88"
 automerge = "1.0.0-beta.3"
 clap = { version = "4.5.3", features = ["derive", "env"], optional = true }
 constant_time_eq = "0.4.2"
+data-encoding = "2.10.0"
 derive_more = { version = "2.0.1", features = ["as_ref", "deref", "display"] }
 dissimilar = "1.0.9"
 futures = { version = "0.3.30", default-features = false }
 git2 = { version = "0.20.2", default-features = false }
 ignore = "0.4.22"
-iroh = "0.35.0"
+iroh = "0.90.0"
 magic-wormhole = "0.7.6"
 microxdg = "0.2.0"
 notify = { version = "8.0.0", default-features = false, features = [
@@ -94,7 +95,7 @@ rust-ini = "0.21.1"
 serde = { version = "1.0.198", features = ["derive"] }
 serde_json = "1"
 tempfile = "3.24.0"
-time = { version = "0.3.36", features = ["formatting"] }
+time = { version = "0.3.36", features = ["formatting", "macros"] }
 tokio-util = { version = "0.7.11", features = ["codec"] }
 tracing = "0.1.40"
 tracing-subscriber = { version = "0.3.18", default-features = false, features = [
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -79,6 +79,7 @@ derive_more = { version = "2.0.1", features = ["as_ref", "deref", "display"] }
 dissimilar = "1.0.9"
 futures = { version = "0.3.30", default-features = false }
 git2 = { version = "0.20.2", default-features = false }
+hexane = "=0.1.6"
 ignore = "0.4.22"
 iroh = "0.92.0"
 magic-wormhole = "0.7.6"
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -80,7 +80,7 @@ dissimilar = "1.0.9"
 futures = { version = "0.3.30", default-features = false }
 git2 = { version = "0.20.2", default-features = false }
 ignore = "0.4.22"
-iroh = "0.90.0"
+iroh = "0.92.0"
 magic-wormhole = "0.7.6"
 microxdg = "0.2.0"
 notify = { version = "8.0.0", default-features = false, features = [
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -71,7 +71,7 @@ clap_mangen = "0.2"
 [dependencies]
 anyhow = "1.0.81"
 async-trait = "0.1.88"
-automerge = "1.0.0-beta.3"
+automerge = "0.7.3"
 clap = { version = "4.5.3", features = ["derive", "env"], optional = true }
 constant_time_eq = "0.4.2"
 data-encoding = "2.10.0"
@@ -79,7 +79,6 @@ derive_more = { version = "2.0.1", features = ["as_ref", "deref", "display"] }
 dissimilar = "1.0.9"
 futures = { version = "0.3.30", default-features = false }
 git2 = { version = "0.20.2", default-features = false }
-hexane = "=0.1.6"
 ignore = "0.4.22"
 iroh = "0.92.0"
 magic-wormhole = "0.7.6"
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -21,8 +21,8 @@ keywords = [
     "peer-to-peer",
 ]
 categories = ["command-line-utilities", "text-editors"]
-edition = "2021"
-rust-version = "1.81.0"
+edition = "2024"
+rust-version = "1.85.0"
 default-run = "teamtype"
 
 [lib]
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -73,7 +73,7 @@ anyhow = "1.0.81"
 async-trait = "0.1.88"
 automerge = "1.0.0-beta.3"
 clap = { version = "4.5.3", features = ["derive", "env"], optional = true }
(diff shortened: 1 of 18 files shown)


daemon/Cargo.toml
```diff
@@ -21,8 +21,8 @@ keywords = [
     "peer-to-peer",
 ]
 categories = ["command-line-utilities", "text-editors"]
-edition = "2021"
-rust-version = "1.81.0"
+edition = "2024"
+rust-version = "1.91.0"
```
blinry: This seems to indicate the minimal supported Rust version. How did you determine this?

I like updating the edition! :) Upping the Rust version to one from October last year should not result in any packaging problems, right?

daemon/src/cli_ask.rs
```diff
@@ -3,7 +3,7 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-use anyhow::{bail, Result};
+use anyhow::{Result, bail};
```
blinry, line 6: Just for my understanding: Why does `cargo fmt` produce different results after upgrading the edition?

Would it be useful/possible to denote the version of `cargo fmt` that the CI uses to lint somewhere?

daemon/Cargo.toml
```diff
@@ -74,12 +74,13 @@ async-trait = "0.1.88"
...
 clap = { version = "4.5.3", features = ["derive", "env"], optional = true }
 constant_time_eq = "0.4.2"
+data-encoding = "2.10.0"
 derive_more = { version = "2.0.1", features = ["as_ref", "deref", "display"] }
 dissimilar = "1.0.9"
 futures = { version = "0.3.30", default-features = false }
 git2 = { version = "0.20.2", default-features = false }
 ignore = "0.4.22"
-iroh = "0.35.0"
+iroh = "0.90.0"
```
blinry: Note: This means jumping ahead into the [canary series](https://www.iroh.computer/blog/iroh-0-90-the-canary-series) of iroh releases.

My feeling was always to wait for iroh 1.0, to avoid increased maintenance effort. But if it turns out that it's not a lot of work for us, I'd gladly take the performance improvements promised in that blog post!

daemon/src/peer.rs
```diff
@@ -65,7 +65,8 @@ impl ConnectionManager {
 
         let (endpoint, my_passphrase) = Self::build_endpoint(base_dir).await?;
 
-        let secret_address = format!("{}#{}", endpoint.node_id(), my_passphrase);
+        let encoded_passphrase = data_encoding::HEXLOWER.encode(&my_passphrase.to_bytes());
```
blinry: Where in the changelog in iroh 0.90 is this recommended? I see that it's required, but does it bring any advantages?

daemon/Cargo.toml
```diff
@@ -94,7 +95,7 @@ rust-ini = "0.21.1"
 serde = { version = "1.0.198", features = ["derive"] }
 serde_json = "1"
 tempfile = "3.24.0"
-time = { version = "0.3.36", features = ["formatting"] }
+time = { version = "0.3.36", features = ["formatting", "macros"] }
```
blinry, line 98: Which change requires enabling this feature?

daemon/vim
blinry, line 1: Seems like this file was added by accident.

blinry (changes requested):
Thanks for these updates! I left a comment on the iroh update below.

I noticed that the automerge crate now started a 0.7 series, that seems to have replaced the 1.0-beta approach! It updates rand to 0.9, so we could update to that, and be very very up-to-date! \o/

daemon/Cargo.toml
zormit: tbh I updated it to `1.91` because that was the one I had installed and on the train I didn't have the internet bandwidth to update my local installation further. But we can definitely go to stable `1.93`

daemon/src/cli_ask.rs
zormit: There's a section on rustfmt [here](https://blog.rust-lang.org/2025/02/20/Rust-1.85.0/) which probably explains this? It seems we _could_ also use a [style edition](https://doc.rust-lang.org/edition-guide/rust-2024/rustfmt-style-edition.html), not sure if that's what we want.

daemon/Cargo.toml
zormit: I'd be happy to be a canary 🐣 maybe we find some cool bugs and learn something =) (and if it's too tedious, we'll go back)

daemon/vim
zormit: dang, that would probably not have happened in my manual `git add -p` process 😬

daemon/Cargo.toml
blinry: I've used https://github.com/foresterre/cargo-msrv in the past to figure that stuff out. I think going as low as possible is the way to go here.

daemon/Cargo.toml
zormit: I was wondering about this as well how the feature requirements change when you don't change the version. Basically it popped up in `cargo check` once you change Iroh from `0.35` to `0.90` and then I fixed it to make it compile again.

The error that comes up
```
error[E0432]: unresolved import `time::macros`
  --> src/logging.rs:7:11
   |
 7 | use time::macros::format_description;
   |           ^^^^^^ could not find `macros` in `time`
   |
note: found an item that was configured out
  --> /Users/mn/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/time-0.3.36/src/lib.rs:95:9
   |
94 | #[cfg(feature = "macros")]
   |       ------------------ the item is gated behind the `macros` feature
95 | pub mod macros;
   |         ^^^^^^
```

=> https://github.com/teamtype/teamtype/blob/5114ef7b6d604f8cd3bd192acec5b44527fd0e4e/daemon/src/logging.rs#L7

Do you have an idea why it could now be required? I tried to dig through [iroh's changes](https://github.com/n0-computer/iroh/compare/v0.35.0...v0.90.0) but didn't see anything that touches `time::macros`.

blinry: Now I suspect that the newer iroh forced a newer version of the time crate, which added this feature. As we're using `time::macros` in our code, it seems fine to enable that feature. :)

daemon/Cargo.toml
zormit:
```
% cargo msrv find
  [Meta]   cargo-msrv 0.18.4

Compatibility Check #1: Rust 1.89.0
  [OK]     Is compatible

Compatibility Check #2: Rust 1.87.0
  [OK]     Is compatible

Compatibility Check #3: Rust 1.86.0
  [OK]     Is compatible

Compatibility Check #4: Rust 1.85.1
  [OK]     Is compatible

Result:
   Considered (min … max):   Rust 1.85.1 … Rust 1.93.0
   Search method:            bisect
   MSRV:                     1.85.1
   Target:                   x86_64-apple-darwin
 ```
If we want to go to edition 2024, we can't go older than that.

File: daemon/Cargo.toml (hunk 5 above)
Comment (zormit) [reply daemon/Cargo.toml:98]: I don't get it, but I'm happy to accept it as well :)

File: daemon/src/cli_ask.rs (hunk 2 above)
Comment (blinry) [reply daemon/src/cli_ask.rs:6]: This I don't really understand, but I'm happy with newer cargo fmt in any case.

**Review by blinry — APPROVED** [review]

File: daemon/Cargo.toml [hunk 6]
```diff
@@ -71,15 +71,15 @@ clap_mangen = "0.2"
...
+automerge = "0.7.3"
 clap = { version = "4.5.3", features = ["derive", "env"], optional = true }
 constant_time_eq = "0.4.2"
 data-encoding = "2.10.0"
 derive_more = { version = "2.0.1", features = ["as_ref", "deref", "display"] }
 dissimilar = "1.0.9"
 futures = { version = "0.3.30", default-features = false }
 git2 = { version = "0.20.2", default-features = false }
-hexane = "=0.1.6"
+hexane = "0.1.6"
```
zormit: I think we then can completely drop `hexane` here, as we're not ourselves depending on it.

daemon/src/peer.rs
```diff
@@ -154,8 +153,8 @@ impl ConnectionManager {
             )
         } else {
             debug!("Generating new keypair.");
-            let secret_key = SecretKey::generate(rand::rngs::OsRng);
-            let passphrase = SecretKey::generate(rand::rngs::OsRng);
+            let secret_key = SecretKey::generate(&mut rand::rng());
+            let passphrase = SecretKey::generate(&mut rand::rng());
```
zormit, line 157: This is a tangent and might have to be moved to a new issue. Two questions:

1. Is it as safe as before to use `rng()` instead of `OsRng`?
2. Is it even safe to use `rand` for our use-case?

So, I'm a bit unsure about this change, I remember seeing somewhere yesterday (and I don't find it anymore) that using `OsRng` is "more secure" than the default. But from [what I find now](https://docs.rs/rand/latest/rand/rngs/struct.ThreadRng.html) it seems that `rng` uses `OsRng`, as `rng()` creates a `ThreadRng` which does "automatic seeding via [OsRng](https://docs.rs/rand/latest/rand/rngs/struct.OsRng.html) and periodically thereafter"

Then the [update guide](https://rust-random.github.io/book/update-0.9.html) says

> It was determined in [#1514](https://github.com/rust-random/rand/pull/1514) that “rand is not a crypto library”.
```

So should we even use it? As we need a random number for generating a security key... does that require a crypto lib?

blinry: As iroh uses `rand::rng()` [in their examples](https://docs.iroh.computer/examples/chat#basic-endpoint-configuration), I'm not concerned. And as you say, this construct still seems to use `OsRng`.

The [Rust Rand Book](https://rust-random.github.io/book/guide-rngs.html#state-and-seeding) states:

> `rand::rng` [...] is fast, good quality, and (to the best of our knowledge) cryptographically secure.

PR #479, opened 2026-02-05
back to iroh 0.35

zormit:
It seems [iroh canary](https://www.iroh.computer/blog/iroh-0-90-the-canary-series) is too much bleeding edge for us, even though we're only a small project. Unless we have clear reasons to upgrade, we'll stay with 0.35 and wait for the stable release.

See also [Zulip: #daemon > iroh canary series implications](https://teamtype.zulipchat.com/#narrow/channel/562985-daemon/topic/iroh.20canary.20series.20implications/with/572146869)

### DIFF
--- a/daemon/Cargo.lock
+++ b/daemon/Cargo.lock
@@ -410,7 +410,7 @@ dependencies = [
  "rand 0.9.2",
  "rustc-hash",
  "serde",
- "sha2 0.11.0-rc.4",
+ "sha2 0.11.0-rc.5",
  "smol_str",
  "thiserror 2.0.18",
  "tinyvec",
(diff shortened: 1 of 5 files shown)


zormit:
FWIW, at least we'll stay with MSRV 1.89 (:

```
Compatibility Check #3: Rust 1.88.0
  [FAIL]   Is incompatible

╭────────────────────────────────────────────────────────────────────────╮
│ error: rustc 1.88.0 is not supported by the following packages:        │
│   automerge@0.7.3 requires rustc 1.89.0                                │
│   smol_str@0.3.5 requires rustc 1.89                                   │
│ Either upgrade rustc or select compatible dependency versions with     │
│ `cargo update <name>@<current-ver> --precise <compatible-ver>`         │
│ where `<compatible-ver>` is the latest version supporting rustc 1.88.0 │
╰────────────────────────────────────────────────────────────────────────╯
```

PR #484, opened 2026-02-12
Refactoring JSON-RPC Forwarder in order to prepare for multiplatform

zormit:
This is a partial extract from #467, to break it down into smaller chunks.

The idea is to refactor the existing code in order to make room for Windows support later - an example how that could look like can be seen in #260.

### DIFF
--- a/daemon/src/jsonrpc_forwarder.rs
+++ b/daemon/src/jsonrpc_forwarder.rs
@@ -32,32 +32,10 @@ pub trait JSONRPCForwarder<
     W: AsyncWrite + Unpin + Send + 'static,
 >
 {
-    async fn connection(&self, directory: &Path) -> anyhow::Result<()>;
     async fn connect_stream(
         &self,
         directory: &Path,
     ) -> anyhow::Result<(FramedRead<R, LinesCodec>, FramedWrite<W, LinesCodec>)>;
-}
-
-pub struct UnixJSONRPCForwarder {}
-
-#[async_trait(?Send)]
-impl JSONRPCForwarder<OwnedReadHalf, OwnedWriteHalf> for UnixJSONRPCForwarder {
-    async fn connect_stream(
-        &self,
-        directory: &Path,
-    ) -> anyhow::Result<(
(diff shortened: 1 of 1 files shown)


daemon/src/jsonrpc_forwarder.rs
```diff
@@ -13,40 +13,75 @@
...
-                .await
-                .expect("Failed to write to stdout");
+// TODO: Put these defaults to a module accessible by config.rs as well.
+pub const DEFAULT_SOCKET_NAME: &str = "socket";
+pub const CONFIG_DIR: &str = ".teamtype";
+
+#[async_trait(?Send)]
+pub trait JSONRPCForwarder<
+    R: AsyncRead + Unpin + Send + 'static,
+    W: AsyncWrite + Unpin + Send + 'static,
```
zormit, line 32: @d0mmi and I were thinking whether these Generics maybe could be also be "summarized" into some `JSONRPCWriter` and `JSONRPCReader` trait / newtype.

daemon/src/jsonrpc_forwarder.rs
```diff
@@ -13,40 +13,75 @@
...
+>
+{
+    async fn connect_stream(
+        &self,
+        directory: &Path,
+    ) -> anyhow::Result<(FramedRead<R, LinesCodec>, FramedWrite<W, LinesCodec>)>;
```
blinry, line 38: No matter of the implementation, we'll always want to frame the bytes using `LinesCodec`. So I think that this should not be part of the trait, to make the implementation simpler and requiring less code duplication.

Could we return a `Result<impl AsyncRead, impl AsyncRead>` here?

(Maybe this also would need `Unpin` and `Send`? I'm not sure.)

daemon/src/jsonrpc_forwarder.rs
```diff
@@ -13,40 +13,75 @@
...
+
+    async fn connection(&self, base_dir: &Path) -> anyhow::Result<()> {
```
blinry, line 40: The code in this function will be the same for all implementations, I think. So we could pull this out of the trait, and put it a function instead.

PR #490, opened 2026-02-19
Small preparations towards Windows support

zormit:
This basically reverts 1aa6439, which introduced the Unix-specific
SIGTERM. In our observation this didn't even work (tested with `kill -9`).

Removing SIGTERM support enables compiling this code on Windows.

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -287,14 +287,6 @@ fn get_current_directory() -> Result<PathBuf> {
 }
 
 async fn wait_for_shutdown() {
-    let mut signal_terminate = signal::unix::signal(signal::unix::SignalKind::terminate())
-        .expect("Should have been able to create terminate signal stream");
-    tokio::select! {
-        _ = signal::ctrl_c() => {
-            debug!("Got SIGINT (Ctrl+C), shutting down");
-        }
-        _ = signal_terminate.recv() => {
-            debug!("Got SIGTERM, shutting down");
-        }
-    }
+    signal::ctrl_c().await.expect("failed to listen for event");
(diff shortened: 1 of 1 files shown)


alerque:
There are several things wrong here. First of all `kill -9` will not send a `SIGTERM`, it will send `SIGKILL`. Hence it is unsurprising that didn't work. Typically you'd use a normal `kill <process>` to send a `SIGTERM` which allows processes to easily trap it and possibly shut down gracefully. Only if that fails and something egregiously bad is going on (e.g. the process is not listening to signals) would you send a `SIGKILL` (by adding the `-9` or whatever) and home the process handler forcefully kills it off for you.

Additionally, NONE of this code would ever work because it's all in the wrong place. You're spawning and running a main loop (daemon or client) then later after that has completed starting a new loop to listen for possible interruptions. That layout won't work, you need to get both loops running at once so that either one can terminate first, then use that info to terminate the others.

This means the CTRL-C processing code here is not actually working either. You can prove this because the `debug!` logging output is never reached, the CTRL-C just gets caught by the tokio process and the main loop gets killed off, then you start listening for shutdown signals after the process has already ended, so that silently terminates too.

Lastly the Posix specific dependency stuff should be gated behind a platform flag. Removing it doesn't seem like the right option here. I'd vote for fixing it, then to enable Windows add the requisite platform specific handling or just skip it and at least hide the Unix specific bits behind a compile time flag.

I'll submit a fix for the main loop issue to at least get the existing code working. I abandoning this PR direction.

zormit:
😅 oops, sounds good, looking forward for your fix. Let me know if I can support / if you want to pair on it.

zormit:
Closing in favor of #533

PR #493, opened 2026-02-24
book: add first-steps page to summary to force rendering

deanrobertcook:
The "first steps" link in the [editor-plugin development page](https://teamtype.github.io/teamtype/editor-plugin-dev-guide.html#sending-an-example-message-to-the-daemon) is broken. This PR just adds the first steps page to an appropriate section in the book to force the page's creation.

Alternatively, I could remove the link and page if it's no longer desired :)


blinry:
Hey Dean! :) Good find! The page used to exist, but we decided to move those "first steps" to the project's main README instead (it's now the "Basic usage" section there). So I think we could slightly rephrase that paragraph ("on a playground" -> "on a testing directory"?), and link to the README instead.

PR #494, opened 2026-02-24
nvim: only send cursor updates on visual mode changes

deanrobertcook:
I noticed that changing from normal to insert or to command mode triggered two cursor updates. The reason being that the ModeChanged event also gets fired. 

If I've understood it correctly, we only care about the "ModeChanged" when leaving visual mode so that we tell peers to clear the visual selection for our cursor. 

It feels a little dirty just to add a flag here. Perhaps there's a tidier trick we could use here?

Hope I'm not spamming with minor fixes here, I'm just trying to familiarize myself with the project. No hard feelings if this fix feels risky or has to be tested more thoroughly. 

Speaking of which, what is the testing landscape for neovim plugins like?

### DIFF
--- a/nvim-plugin/lua/teamtype/cursor.lua
+++ b/nvim-plugin/lua/teamtype/cursor.lua
@@ -20,6 +20,8 @@ local cursor_namespace = vim.api.nvim_create_namespace("Teamtype")
 local offset_encoding = "utf-32"
 local cursor_timeout_ms = 300 * 1000
 local following_user_id = nil
+-- keeps track across ModeChanged events whether the last mode was visual
+local was_visual_selection = nil
 
 local function show_cursor_information(name, cursor)
     return (name or "Unknown user") .. " @ " .. vim.uri_to_fname(cursor.uri) .. ":" .. cursor.range.start.line + 1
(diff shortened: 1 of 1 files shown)


nvim-plugin/lua/teamtype/cursor.lua
```diff
@@ -241,6 +248,7 @@ function M.track_cursor(bufnr, callback)
                 ranges = { range }
             end
 
+            was_visual_selection = visualSelection
```
blinry: Could we move this line up, directly after the if that tests it? To me, that feels like keeping related code together better.

nvim-plugin/lua/teamtype/cursor.lua
```diff
@@ -20,6 +20,7 @@ local cursor_namespace = vim.api.nvim_create_namespace("Teamtype")
 local offset_encoding = "utf-32"
 local cursor_timeout_ms = 300 * 1000
 local following_user_id = nil
+local was_visual_selection = nil
```
blinry, line 24: Ideally, we could add a brief comment here explaining what this variable means/is used for. (Even if we didn't do it for the others ^^')

blinry (approved):
Ha, it's great that you find and fix these "minor" issues!

This PR seems like a good idea, and I confirmed in a manual test that sending a de-selection still works. I left two nitpick comments in the code.

You asked about testing: We have two tests that involve the Neovim plugin [here](https://github.com/teamtype/teamtype/tree/main/daemon/integration-tests/tests). One is a fuzzer that tests correct sync in a session with two Neovim instances and two daemons, and the other one test whether the plugin sends and processes deltas correctly. However, these don't test anything about cursors or selections at all. Thoughts on how to improve/extend these would be very welcome. We don't really have tests for other editor plugins, for example...

PR #498, opened 2026-03-13
fix: support string JSON-RPC ids in editor protocol

### LINKED ISSUE #497: JSON-RPC Protocol Must Handle String IDS
**Describe the bug**
According to the [JSON-RPC spec](https://www.jsonrpc.org/specification) IDs can be strings or numbers:

> An identifier established by the Client that MUST contain a String, Number, or NULL value if included. If it is not included it is assumed to be a notification. The value SHOULD normally not be Null [[1]](https://www.jsonrpc.org/specification#id1) and Numbers SHOULD NOT contain fractional parts [[2]](https://www.jsonrpc.org/specification#id2)

While working on https://github.com/teamtype/teamtype-jetbrains/pull/19 I noticed that the `JsonRequest` methods never finish and block some threads. With some trace logging I see that Eclipse's lsp4j generates string IDs:

```json
{
  "jsonrpc" : "2.0",
  "id" : "1",
  "method" : "open",
  "params" : {
    "uri" : "file:///tmp/untitled/test.txt",
    "content" : ""
  }
}
```

**To Reproduce**

1. Start Teamtype Jetbrains in debug mode via Intellij (based on the aforementioned PR)
2. Put a breakpoint [here](https://github.com/teamtype/teamtype-jetbrains/pull/19/changes#diff-ff7cfeea366f005e637b036972906d4f2a0c7e73e7360e09552602e394ba9722R291) and [here](https://github.com/teamtype/teamtype-jetbrains/pull/19/changes#diff-ff7cfeea366f005e637b036972906d4f2a0c7e73e7360e09552602e394ba9722R351)
3. Launch the Teamtype daemon in the UI
4. Notice that the debugger hits the first breakpoint but not the second.
**Expected behavior**
Should hit both breakpoints.

You can also see that the debug output in the Intellij UI lists a notification instead of a request.

```
21:57:50Z DEBUG ThreadId(08) teamtype::daemon: Handling doc message: FromEditor(0, Notification { payload: Open { uri: "file:///tmp/untitled/test.txt", content: "" } })
```

**Versions:**
- Teamtype `0.9.1`:
- Jetbrains, this PR: https://github.com/teamtype/teamtype-jetbrains/pull/19

saschabuehrle:
Fixes #497.

JSON-RPC allows request ids to be either numbers or strings, but the daemon previously only accepted numeric ids. This caused parsing failures with clients that send string ids.

This PR adds support for both id types in request parsing and response serialization, and includes regression tests for string-id cases.

**Self-review checklist**

- [x] I have tested, and self-reviewed my code.
- [ ] I have added myself to the REUSE headers in the files where I made significant changes.
- [x] I have manually tested my changes, if applicable.

Greetings, saschabuehrle

### DIFF
--- a/daemon/src/editor_protocol.rs
+++ b/daemon/src/editor_protocol.rs
@@ -31,6 +31,8 @@ impl OutgoingMessage {
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(untagged)]
+/// JSON-RPC 2.0 allows request ids to be either Number or String.
+/// See: https://www.jsonrpc.org/specification#request_object
 pub enum JsonRpcId {
     Number(usize),
     String(String),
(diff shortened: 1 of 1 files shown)


saschabuehrle:
Withdrawing this PR. Thank you for your time reviewing.

Greetings, saschabuehrle

PR #500, opened 2026-03-19
Add JSON RPC ID String Handling

### LINKED ISSUE #497: JSON-RPC Protocol Must Handle String IDS

schrieveslaach:
This change ensures that Teamtype supports JSON RPC IDs as usize and strings.

Fixes #497 

**Self-review checklist**

- [ ] I have manually tested and self-reviewed my changes (not yet, planning to test with https://github.com/teamtype/teamtype-jetbrains/pull/19).
- [ ] I have added myself to the REUSE headers in the files where I made significant changes.
    - [ ] I have reviewed the output carefully, and I understand the resulting changes in detail. I am responsible for the changes, and I made sure that my PR represents excellent work.

### DIFF
--- a/daemon/src/editor_protocol.rs
+++ b/daemon/src/editor_protocol.rs
@@ -29,11 +29,36 @@ impl OutgoingMessage {
     }
 }
 
+/// An identifier based on the [JSON-RPC spec](https://www.jsonrpc.org/specification):
+///
+/// > An identifier established by the Client that MUST contain a String, Number, or NULL value if
+/// > included. If it is not included it is assumed to be a notification. The value SHOULD normally
+/// > not be Null [1](https://www.jsonrpc.org/specification#id1) and Numbers SHOULD NOT contain
+/// > fractional parts [2](https://www.jsonrpc.org/specification#id2)
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum JSONRPCId {
+    Usize(usize),
+    String(String),
+}
+
+impl From<usize> for JSONRPCId {
+    fn from(value: usize) -> Self {
+        Self::Usize(value)
+    }
+}
+
+impl From<String> for JSONRPCId {
+    fn from(value: String) -> Self {
+        Self::String(value)
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum IncomingMessage {
     Request {
-        id: usize,
+        id: JSONRPCId,
         #[serde(flatten)]
         payload: EditorProtocolMessageFromEditor,
     },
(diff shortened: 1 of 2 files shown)


daemon/src/editor_protocol.rs
```diff
@@ -29,11 +29,36 @@ impl OutgoingMessage {
...
 
+/// An identifier based on the [JSON-RPC spec](https://www.jsonrpc.org/specification):
+///
+/// > An identifier established by the Client that MUST contain a String, Number, or NULL value if
+/// > included. If it is not included it is assumed to be a notification. The value SHOULD normally
+/// > not be Null [1](https://www.jsonrpc.org/specification#id1) and Numbers SHOULD NOT contain
+/// > fractional parts [2](https://www.jsonrpc.org/specification#id2)
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum JSONRPCId {
```
schrieveslaach, line 40: I used `JSONRPCResponse ` as naming inspiration to blend in. :wink:

PR #504, opened 2026-03-28
doc: Build Instructions added to README

christf:
this closes #503

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,15 @@ Connected to peer: 5e6b787fff79074735eb9b56939269100de1e37bc7f7a4d29c277cc24f7ee
 
 The directories are now connected, and changes will be synced instantly. You can open text files (using editors with a Teamtype plugin), and start collaborating in real time! :sparkles:
 
+## 🚜 Compiling
+
+Navigate to daemon and execute
+
+```
+$ cargo build --release
+
+This will create a binary called target/release/teamtype
+
 ## 🎓 Learn more
 
 - Learn more about Teamtype in [the documentation](https://teamtype.github.io/teamtype).


README.md
```diff
@@ -119,6 +119,15 @@ Connected to peer: 5e6b787fff79074735eb9b56939269100de1e37bc7f7a4d29c277cc24f7ee
 
 The directories are now connected, and changes will be synced instantly. You can open text files (using editors with a Teamtype plugin), and start collaborating in real time! :sparkles:
 
+## 🚜 Compiling
```
alerque, line 122: If added at all, this should go at the end of the installation section as a "build from source code" alternative, not as a new section in an unrelated location.

README.md
```diff
@@ -119,6 +119,15 @@ Connected to peer: 5e6b787fff79074735eb9b56939269100de1e37bc7f7a4d29c277cc24f7ee
...
+
+Navigate to daemon and execute
```
alerque, line 124:
```suggestion
Navigate to the `daemon/` subdirectory of the source, then execute
```

README.md
```diff
@@ -119,6 +119,15 @@ Connected to peer: 5e6b787fff79074735eb9b56939269100de1e37bc7f7a4d29c277cc24f7ee
...
+
+```
+$ cargo build --release
```
alerque, line 127:
```suggestion
$ cargo build --locked --release
```

README.md
```diff
@@ -119,6 +119,15 @@ Connected to peer: 5e6b787fff79074735eb9b56939269100de1e37bc7f7a4d29c277cc24f7ee
...
+
+This will create a binary called target/release/teamtype
```
alerque, line 129:
```suggestion
This will create a binary called `target/release/teamtype` that needs to be moved to your shell's `$PATH`, similar to other manual installation methods above.
```

alerque (changes requested):
I'm still a bit confused why this is needed at all, it still seems to be the recommendation in the README to use `cargo install` is already the best choice for people that don't really know what they are doing with Rust builds and just want to get it running, while anybody that wants something more flexible than that probably won't benefit from these details. I don't think we want to give people the impression from the README that downloading the monorepo sources and building bits of it here and there is the way to get going.

If this does get added to the readme, at least here are some edits and comments on where it should go.

zormit:
Yeah I agree that the README isn't really the right place for this. We could put something similar in CONTRIBUTING.md and maybe link there?

alerque:
That sounds better; instructions for people messing around with building from sources would be a better fit in CONTRIBUTING.md than the README.md since it's really only an approach to getting running if you are planning to modify something along the way.

blinry:
I gave that attempt a go in #518.

zormit:
Closing in favor of #518, thanks @christf for bringing up the discussion!

PR #506, opened 2026-03-29
introduce the ability to specify an .teamtypeignore file

thyseus:
Added support for customer `.teamtypeignore` file in addition to .gitignore file.

This is also possible in syncthing for example with .stignore for files which should not be synced but also should not appear in .gitignore.

The syntax is similar to `.gitignore`, supporting file names, directories, and glob patterns.

- [?] I have added myself to the REUSE headers in the files where I made significant changes.
- [x] I have reviewed the output carefully, and I understand the resulting changes in detail. I am responsible for the changes, and I made sure that my PR represents excellent work.

### DIFF
--- a/daemon/src/sandbox.rs
+++ b/daemon/src/sandbox.rs
@@ -248,6 +248,19 @@ mod tests {
         fs::write(project_dir.join("dir").join("b"), b"This is b file")
             .expect("Failed to write file");
         fs::write(dir.path().join("secret"), b"This is a secret").expect("Failed to write file");
+        fs::write(project_dir.join("a.txt"), b"visible file").expect("Failed to write file");
+        fs::write(project_dir.join("b.txt"), b"another visible file")
+            .expect("Failed to write file");
+        fs::write(project_dir.join("secret.txt"), b"should be ignored")
+            .expect("Failed to write file");
+        fs::create_dir(project_dir.join("ignored_dir")).expect("Failed to create directory");
+        fs::write(
+            project_dir.join("ignored_dir").join("c.txt"),
+            b"should be ignored",
+        )
+        .expect("Failed to write file");
+        fs::write(project_dir.join("debug.log"), b"should be ignored")
+            .expect("Failed to write file");
 
         dir
     }
@@ -346,21 +359,6 @@ mod tests {
     fn obeys_teamtypeignore() {
         let dir = temp_dir_setup();
         let project_dir = dir.path().join("project");
-
-        fs::write(project_dir.join("a.txt"), b"visible file").expect("Failed to write file");
-        fs::write(project_dir.join("b.txt"), b"another visible file")
-            .expect("Failed to write file");
-        fs::write(project_dir.join("secret.txt"), b"should be ignored")
-            .expect("Failed to write file");
-
-        fs::create_dir(project_dir.join("ignored_dir")).expect("Failed to create directory");
-        fs::write(
-            project_dir.join("ignored_dir").join("c.txt"),
-            b"should be ignored",
-        )
-        .expect("Failed to write file");
-        fs::write(project_dir.join("debug.log"), b"should be ignored")
-            .expect("Failed to write file");
 
         fs::write(
             project_dir.join(".teamtypeignore"),
--- a/daemon/src/sandbox.rs
+++ b/daemon/src/sandbox.rs
@@ -247,20 +247,11 @@ mod tests {
         fs::create_dir(project_dir.join("dir")).expect("Failed to create directory");
         fs::write(project_dir.join("dir").join("b"), b"This is b file")
             .expect("Failed to write file");
-        fs::write(dir.path().join("secret"), b"This is a secret").expect("Failed to write file");
-        fs::write(project_dir.join("a.txt"), b"visible file").expect("Failed to write file");
-        fs::write(project_dir.join("b.txt"), b"another visible file")
-            .expect("Failed to write file");
-        fs::write(project_dir.join("secret.txt"), b"should be ignored")
-            .expect("Failed to write file");
-        fs::create_dir(project_dir.join("ignored_dir")).expect("Failed to create directory");
         fs::write(
-            project_dir.join("ignored_dir").join("c.txt"),
-            b"should be ignored",
+            dir.path().join("file-outside-of-project"),
+            b"This is a file outside of project",
         )
         .expect("Failed to write file");
-        fs::write(project_dir.join("debug.log"), b"should be ignored")
-            .expect("Failed to write file");
 
         dir
     }
@@ -359,44 +350,29 @@ mod tests {
     fn obeys_teamtypeignore() {
         let dir = temp_dir_setup();
         let project_dir = dir.path().join("project");
+        let teamtypeignore = project_dir.join(".teamtypeignore");
 
-        fs::write(
-            project_dir.join(".teamtypeignore"),
-            b"secret.txt\nignored_dir/\n*.log\n",
-        )
-        .expect("Failed to write .teamtypeignore");
+        fs::write(&teamtypeignore, b"a\n").expect("Failed to write .teamtypeignore");
 
         let app_config = AppConfig {
-            base_dir: project_dir,
+            base_dir: project_dir.clone(),
             ..Default::default()
         };
 
-        let files = enumerate_non_ignored_files(&app_config);
-        let file_names: Vec<&str> = files
-            .iter()
-            .filter_map(|p| p.file_name())
-            .filter_map(|n| n.to_str())
-            .collect();
+        assert!(
+            ignored(&app_config, &project_dir.join("a")).unwrap(),
+            "a should be ignored"
+        );
+        assert!(
+            !ignored(&app_config, &project_dir.join("dir/b")).unwrap(),
+            "b should be not ignored"
+        );
+
+        fs::write(&teamtypeignore, b"a\ndir\n").expect("Failed to write .teamtypeignore");
 
         assert!(
-            file_names.contains(&"a.txt"),
-            "a.txt should be in the list of files"
-        );
-        assert!(
-            file_names.contains(&"b.txt"),
-            "b.txt should be in the list of files"
-        );
-        assert!(
-            !file_names.contains(&"secret.txt"),
-            "secret.txt should be ignored by .teamtypeignore"
-        );
-        assert!(
-            !file_names.contains(&"ignored_dir"),
-            "ignored_dir should be ignored by .teamtypeignore"
-        );
-        assert!(
-            !file_names.contains(&"debug.log"),
-            "debug.log should be ignored by .teamtypeignore"
+            ignored(&app_config, &project_dir.join("dir/b")).unwrap(),
+            "now b should be ignored"
         );
     }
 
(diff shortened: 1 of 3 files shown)


thyseus:
First use case: collaborating on projects that does not have a .gitignore or git at all. 
Syncthing also has a isolated .stignore file.

Thinking about it: if i include the .teamtypeignore in .gitignore, each collaborator could have his own customized secrets to stay out of sync to other peers. Would not do that by default.

Yes, `.git/info/exclude` would have a related effect (local-only ignore) but still slighly different.

What is a REUSE header ?

alerque (approved):

daemon/src/sandbox.rs
```diff
@@ -341,6 +342,69 @@ mod tests {
         assert!(read_file(&project_dir, &project_dir.join("nonexistant")).is_err());
     }
 
+    #[test]
+    fn obeys_teamtypeignore() {
+        // set up test scenario:
```
zormit: I would be better to reuse `temp_dir_setup` here, _maybe_ if it's really necessary, add a second method for `add_ignored_files`, otherwise extend `temp_dir_setup` for some more files that could be ignored.

thyseus: You are right - will fix it...

blinry:
@thyseus, in the self-review checklist, you declared that you used generative AI to create this PR. Could you please disclose after that bullet point which tool you used, and to what extent? This is important to us for transparency and trust reasons.

thyseus:
I have unchecked the checkbox, i have found the `add_custom_ignore_filename` by the LSP auto complete actually, thats not AI in the modern AI sense :)

blinry:
Ah, makes sense. Thanks for clarifying! :)

I'm still a bit hesitant whether the additional complexity of having a `.teamtypeignore` would suit the project.

## Use case

I'd be curious about very concrete use-cases, if you can think of some! So far, I always thought that "sync the entire Git repo" was a pretty useful way to look at it. If collaborating on a directory/project that doesn't use Git, it seems rather natural for me to still add a `.gitignore` for that purpose. For example, the jujutsu VCS simply uses `.gitignore`. ripgrep uses `.gitignore` and `.ignore`, I think.

Why would people want to exclude synchronization of certain files from a project, but still track these files using Git?

- For performance reasons? (In that case, could we make Teamtype more performant for those cases?)
- For privacy reasons? (Where might that apply, if the files are tracked in Git.)
- For other practical reasons? (Like maybe, only wanting to "quickly collaborate" on a single text file?)

## Tracking `.teamtypeignore` with Git?

It seems projects would have a choice between tracking the `.teamtypeignore` in Git or not. Trying to get clarity of the properties of both options:

Advantages of tracking the file in Git:

- No work required to set up the set of ignored files on different peers.
- Set of ignored files is always the same for all peers, which avoids strange states (see below).

Disadvantages of tracking the file in Git:

- Ignored files are not customizable per person, without introducing a "dirty Git state". If, in a big project, different pairs of people would like to use Teamtype to collaborate on different sets of files, they have to deal with that, maybe reset the file after their pairing session and make sure not to commit the change.
- The `.teamtypeignore` would randomly show up in projects, even though it's not relevant for people not using Teamtype?

## Options

Trying to summarizing the options we'd have to allow users to sync different files than are tracked by Git:

1. Introduce a `.teamtypeignore`.
    - Good, because it's an straightforward & sharable way of denoting this kind of information.
    - Good, because it's a well-established pattern (`.stignore`, `.npmignore`, `.dockerignore`, ...)
	- Bad, because projects have the choice between tracking it with Git or not, which both have downsides, see above.
2. Add an option like [`rg`'s `--glob` option](https://manpages.debian.org/testing/ripgrep/rg.1.en.html), which would allow you to specify which files should be synced. Maybe also as an option in `.teamtype/config`?
    - Similar properties to a non-checked-in ignore file, I think.
    - Bad, because users have to remember to always include the correct glob options.
3. Don't add a new functionality, and write documentation detailing how people could work around the limitation. For example, using `.git/info/exclude`.
    - Similar properties to a non-checked-in ignore file?
    - Good, because it doesn't increase complexity.
    - Bad, because now users definitely *need* to track the repository using Git.

(Haven't thought about this super deeply, and made no decision, open to change my mind on the "ratings"! :)

## Open questions

- If a project contains both a `.teamtypeignore`, as well as a `.gitignore`, should both be respected? I think yes?
- If the `.teamtypeignore` is not synchronized by Teamtype (because all peers Git-ignore it, or because the file ignores itself), what would be the result of peers having different `.teamtypeignore`s? I'd imagine if a file is ignored by A, but not by B, the sync would only happen in the direction of B to A?

alerque:
> So far, I always thought that "sync the entire Git repo" was a pretty useful way to look at it.

This is a useful bootstrap/default, but I believe it would be terrible for it to always function this way.

> If collaborating on a directory/project that doesn't use Git, it seems rather natural for me to still add a `.gitignore` for that purpose.

Uncouth :-P

Most apps that start out trying to glob-on to another apps configuration files usually end up making a mess and eventually implementing their own. This has happened to many apps that started with `.gitignore` and ended up with their own. Interpreting it to play nice with existing things if it happens to exist is one thing, creating it with the contents tailored for a different purpose that what it is documented as being for where it may mess up something else down the road is different.

> For example, the jujutsu VCS simply uses `.gitignore`.

Jujutsu is a wrapper around Git, it uses Git object storage and directly interops with Git.

> ripgrep uses `.gitignore` and `.ignore`, I think.

Ripgrep optionally respects gitignore as a starting point, but also has an elaborate system of glob matching and excluding—because exactly matching what Git tracks is a useful place to start but NOT the most efficient/effectual for many use cases.

> Why would people want to exclude synchronization of certain files from a project, but still track these files using Git?
> 
>     * For performance reasons? (In that case, could we make Teamtype more performant for those cases?)

Yes.

Lets say your team is working on documenting a mono repo. The repo might have tens of thousands of code files tracked in Git. For your team collaborating, you may only want to sync edits on the Markdown files or only on the docs directory or something like that. Trying to sync the whole workspace would be absurd network overhead to bring in a new client.

Git reops can also have all sorts of non text assets including LFS or GitAnnex tracked ones that should be excluded from teamtype sessions but tracked in VCS.

>     * For privacy reasons? (Where might that apply, if the files are tracked in Git.)

Yes.

One way to handle secrets is to track them in a repo but use clean/smudge filters so the shared VCS repo technically contains the data but only team members that need to have the secrets can decrypt them. For example I use `git crypt` for this purpose.

>     * For other practical reasons? (Like maybe, only wanting to "quickly collaborate" on a single text file?)

Yes.

Another example might be a large project where somebody wants to invite collaboration on a small part of it without having to break up a larger repo that might even have unpublished work without having to rethink access control to the whole thing.

> It seems projects would have a choice between tracking the `.teamtypeignore` in Git or not. Trying to get clarity of the properties of both options:

The typical default would be to track files like this unless they were ephemeral or only used by a small subset of a team. This would be case by case with a suggestion for projects that use it regularly to track it.

>     * Set of ignored files is always the same for all peers, which avoids strange states (see below).

This is an orthogonal issue that will have to be dealt with both ways anyway.

>     * The `.teamtypeignore` would randomly show up in projects, even though it's not relevant for people not using Teamtype?

In that case the people using it could add it to their USER git ignore.

> Trying to summarizing the options we'd have to allow users to sync different files than are tracked by Git:

Missing option:

>     1. Introduce a `.teamtypeignore`.
>        
>        * Good, because it's an straightforward & sharable way of denoting this kind of information.
>        * Good, because it's a well-established pattern (`.stignore`, `.npmignore`, `.dockerignore`, ...)
>        * Bad, because projects have the choice between tracking it with Git or not, which both have downsides, see above.

I don't think this 'bad' is even bad at all.

>     2. Add an option like [`rg`'s `--glob` option](https://manpages.debian.org/testing/ripgrep/rg.1.en.html), which would allow you to specify which files should be synced. Maybe also as an option in `.teamtype/config`?
>        
>        * Similar properties to a non-checked-in ignore file, I think.
>        * Bad, because users have to remember to always include the correct glob options.

Globbing and excluding are complicated, implementing this manually and using it would be complicated. For one off use cases it might be nice to have a "launch with just this file set" option, but for regular use excluding files it would be too confusing to use.

>     3. Don't add a new functionality, and write documentation detailing how people could work around the limitation. For example, using `.git/info/exclude`.
>        
>        * Similar properties to a non-checked-in ignore file?
>        * Good, because it doesn't increase complexity.
>        * Bad, because now users definitely _need_ to track the repository using Git.

Don't send people into `.git` for anything. There are dragons there and sending people there to touch it manually for any reason is not a good plan.

>     * If a project contains both a `.teamtypeignore`, as well as a `.gitignore`, should both be respected? I think yes?

This is a harder question. I would tend to default to using .gitignore only if .teamtypeignore does not exist, and if it does using it directly or having some explicit switch for using it as an additional filter.

>     * If the `.teamtypeignore` is not synchronized by Teamtype (because all peers Git-ignore it, or because the file ignores itself), what would be the result of peers having different `.teamtypeignore`s? I'd imagine if a file is ignored by A, but not by B, the sync would only happen in the direction of B to A?

Yes. I think than possible foot gun is much better than not having the ability to build a filter list at all.

blinry:
Thanks for sharing your thoughts and reasonings, @alerque! :)

> Lets say your team is working on documenting a mono repo. The repo might have tens of thousands of code files tracked in Git. For your team collaborating, you may only want to sync edits on the Markdown files or only on the docs directory or something like that.

This would already be possible by starting the daemon in the docs directory. But, granted, it only works if the part you want to collaborate on is a subdirectory.

I still feel a bit hesitant (for complexity/maintenance burden reasons), but seeing that we have 2-3 people in this discussion who really seem to want this feature, I'm open to us trying this!

## Interplay of .gitignore and .teamtypeignore

Intuitively, I think that Teamtype should ignore the *union* of both types of files. I think this is what the PR as it currently is would do.

Reasoning: Imagine a monorepo that has a `.gitignore` in its root, which ignores build artifacts of some kind. Now users want to collaborate on a subdirectory, so they create the following `.teamtypeignore`:

```gitignore
*
!/subdir/
```

If this would now "replace" the `.gitignore`, the users would need to make sure to also find the correct lines that ignore the build artifacts, and put them in the `.teamtypeignore` additionally.

On the other hand, I guess `.teamtypeignore` should be able to "override" `.gitignore`, so *if* the users want to share the build artifacts like object files for whatever reason, they could put `!*.o` in the `.teamtypeignore`, I guess?

blinry:
## Ignoring something means never deleting/editing it?

Another thing we should be careful about: If A ignores "/dir/", but B doesn't, and then B deletes "/dir/file", probably that file shouldn't be deleted on A's side, right? I'd expect the daemon to "constrain" which files are touched to the set of non-ignored files for that peer.

That also would mean that in the situation where A ignores "file", but B doesn't, and now B sends an edit, A should probably ignore that edit, and leave "file" untouched.

thyseus:
Imho we should rely on the 'ignore' crate that the union of multiple `.ignore` files is calculated correctly. I have screened their code, they do some crazy stuff and have a lot of thought in there already.

For having different behaviors depending on which peer has which .ignore setting, we should discuss that as a separate feature in a separate MR. Maybe a `.teamtype.toml` or something would be a solution where we could configure even more necessary stuff.

Will resolve the changelog conflict now.

alerque:
Yes to both points. I've used the `ignore` crate for some other projects and it's pretty robost and does about what people expect.

blinry:
@zormit and I made the tests for this mechanism a bit more minimalist, and added a brief documentation.

Thanks for bringing this up, @thyseus, and thanks everyone for the discussion! Please give it a try and see how it feels, and we can iterate more on it in the future!

PR #507, opened 2026-03-30
chore(release): Linux-archive renamed to better reflect content

### LINKED ISSUE #502: binary for glibc?
thank you for building teamtype. It would be great if there was a binary that would run on glibc-based linuxes in addition to musl. :-)

christf:
This removes the musl-suffix from the linux archive which has led to confusion as per #502.

This fixes #502

- [ ] I have manually tested and self-reviewed my changes.
    - [ ] I have reviewed the output carefully, and I understand the resulting changes in detail. I am responsible for the changes, and I made sure that my PR represents excellent work.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,8 +29,10 @@ jobs:
         include:
           - target: x86_64-unknown-linux-musl
             os: ubuntu-latest
+            archive: $bin-x86_64-linux-static
           - target: aarch64-unknown-linux-musl
             os: ubuntu-latest
+            archive: $bin-aarch64-linux-static
           - target: universal-apple-darwin
             os: macos-latest
     runs-on: ${{ matrix.os }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 
 Add support for JSON-RPC IDs in string to be compliant with the [JSON-RPC specification](https://www.jsonrpc.org/specification).
 
+## Note for package maintainers: Renamed the Linux binaries
+
+We've renamed the Linux binaries
+- `teamtype-x86_64-unknown-linux-musl` to `teamtype-x86_64-linux-static`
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -29,6 +29,7 @@ jobs:
         include:
           - target: x86_64-unknown-linux-musl
             os: ubuntu-latest
+            archive: teamtype-x86-linux-static
```
zormit: We could try the `$bin-$tag-` prefix, so add the version number while we're here. I'd also like to keep the 64 part, and maybe even musl, so I'd suggest
```suggestion
            archive: $bin-$tag-x86_64-linux-musl-static
```

Also we need a similar one for `aarch64` and Apple (as I added the `$tag`)

alerque: So a couple pros and cons here:

* re `x86_64` — keeping the \_64 makes 100% sense, it's the defacto name of the architecture and easiest to grab from various sources.
* re having the `$tag` — that is helpful when having several versions downloaded locally, but it does limit one frequent use case ... some people like to write pipelines that fetch a binary via curl and directly use it. It's easy to compose a link to a GitHub hosted release asset using their `latest` keyword is place of the tag name to always get the assets from the latest release. If you add the tag name to the asset name itself you force downstream consumers to first look up what the latest tag is, then compose a URL based on it. I'd say all other things being equal just forgo adding `$tag` to the filename, at least for *this* purpose. Attaching other assets like source archives, cargo vendor dumps, etc. probably are **not** constrained by the same use case requirements.
* re keeping `-musl`: I'd say the information is largely redundant. It isn't 100% overlap with static: one is more the 'how' and one is the 'what', but for all effective purposes a static binary is a static binary and nobody cares which C library it was linked with. Since there will not be a glibc-static or other alternative forthcoming I'd just call it 'static' myself.

And yes having the same treatment for the other platforms makes sense too.

blinry: I agree with all three points! So my preferred name, if we change it, would be `teamtype-x86_64-linux-static`.

blinry (approved):

christf:
thank you :-)

PR #511, opened 2026-04-08
Reformat related project table, sort by feature and optimize readability

alerque:
This should make both the Markdown and the rendered table easier to read and edit. The list is sorted by prioritizing the status of each column starting with the leftmost feature column working right, and using alphabetic to resolve the very few that are ambiguous on feature sets.

Addition ally this formats the Markdown table so the column status can be seen, moving the links out of line to save line width, etc.

A few compromises were made to shorted names so that in most cases the lines of the rendered table will stay as single-lines instead of awkward wraps.

### DIFF
--- a/book/src/related-projects.md
+++ b/book/src/related-projects.md
@@ -10,42 +10,45 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 There have been a number of attempts of enabling collaborative text editing!
 If you think a project is missing, feel free to [submit an issue or a PR][teamtype] to add them to the list!
 
-|                                                           | FOSS        | Active[^a] | P2P | CRDT   | Editor-agnostic                                        |
-|-----------------------------------------------------------|-------------|------------|-----|--------|--------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


alerque:
Before and after rendered via `mdbook`:

image

zormit (approved):
Thank you for cleaning it up, looks pretty neat!

We had also though about sorting it in a different way than by "an unexplicable metric based on gut feeling" :D

alerque:
I just force pushed to correct the sort of OpenColabTools which should have been above not below Collectives based on column checks.

blinry:
Thanks for cleaning this up! :) Polished the result a bit more in a follow-up commit.

To make the table more accessible to non-experts (like, for example, grant-givers :eyes:), I'd prefer to avoid abbreviations in the headers. One exception is "CRDT", which is probably more precise than the previous "Local-first". I added a footnote spelling it out, though. This is probably our "weakest" column. Maybe we should rename it to "offline support", and re-evaluate the projects?

Issue #344, opened 2025-08-04
Allow specifying custom Iroh + Wormhole relay servers


blinry:
We've got that feature request a couple of times:

- https://www.reddit.com/r/neovim/comments/1mcapt8/comment/n63y4r4/
- https://lobste.rs/s/wd6fhe/ethersync_peer_peer_collaborative#c_tmnqjv

alerque:
Also check `iroh-dns-server`.

alerque:
Closed by #523, we just didn't get the PR wired up to do so automatically.

PR #512, opened 2026-04-10
[WIP] Add optional custom iroh relay server url

martengooz:
## Summary
Added optional functionality to specify your own custom iroh server. 
This PR enables an optional configuration and command line argument to specify a custom iroh relay server. If not configured, teamtype will use the default iroh n0 servers.

To configure a custom relay you can either use the command line argument:

teamtype share --iroh-relay http://localhost:3340/

or by adding it to the .teamtype/config file

peer=6d02...2f4587
iroh_relay=http://localhost:3340/

## Why this change
Same reasons as [https://github.com/teamtype/teamtype/pull/446](https://github.com/teamtype/teamtype/pull/446)

## **Self-review checklist**

- [ x ] I described the changes of the PR in the [CHANGELOG](https://github.com/teamtype/teamtype/blob/main/CHANGELOG.md).
- [ x ] I have manually tested and self-reviewed my changes.
- [ x ] I have added myself to the REUSE headers in the files where I made significant changes.
- [ x ] I have used generative AI/LLMs for producing this PR. Here's which tool I used, and to which extent: Claude Code, 
    - [ x ] I have reviewed the output carefully, and I understand the resulting changes in detail. I am responsible for the changes, and I made sure that my PR represents excellent work.

### DIFF
--- a/daemon/src/peer.rs
+++ b/daemon/src/peer.rs
@@ -60,10 +60,14 @@ pub struct ConnectionManager {
 }
 
 impl ConnectionManager {
-    pub async fn new(document_handle: DocumentActorHandle, base_dir: &Path) -> Result<Self> {
+    pub async fn new(
+        document_handle: DocumentActorHandle,
+        base_dir: &Path,
+        iroh_relay: Option<String>,
+    ) -> Result<Self> {
         let (message_tx, message_rx) = mpsc::channel(1);
 
-        let (endpoint, my_passphrase) = Self::build_endpoint(base_dir).await?;
+        let (endpoint, my_passphrase) = Self::build_endpoint(base_dir, iroh_relay).await?;
 
         let secret_address = format!("{}#{}", endpoint.node_id(), my_passphrase);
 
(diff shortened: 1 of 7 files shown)


daemon/src/peer.rs
```diff
@@ -105,15 +109,24 @@ impl ConnectionManager {
...
         let (secret_key, my_passphrase) = Self::get_keypair(base_dir);
 
-        let endpoint = iroh::Endpoint::builder()
+        let mut builder = iroh::Endpoint::builder()
             .secret_key(secret_key)
             .alpns(vec![ALPN.to_vec()])
-            .discovery_n0()
-            .bind()
-            .await?;
+            .discovery_n0();
```
martengooz, line 121: Thinking about a mode where this is removed as well, for networks without internet access. Will need some more testing to see what happens.

martengooz, line 121: I believe this also introduces a fallback to n0 servers when the custom url cannot be reached. Seeing some logs about that, but needs more investigation.

zormit:
Thanks for this PR! And thanks for being transparent about LLM usage, we appreciate it. For you information, we are currently discussing whether we should make our LLM policy even stricter than just transparency. The discussion is [happening on our Zulip](https://teamtype.zulipchat.com/#narrow/channel/557642-general/topic/LLM.20usage.20policy/with/585592943) and we would leave your PR open for now until we have decided on a way forward.

It's a bit of a strange position to be in and feels slightly unfair to stall a contribution, as we appreciate the time and effort you put into this, so I hope you're not too bummed about our stalling. We also want a configurable relay and I don't really know we would do, if we say no to this contribution. We might try to code it "by hand" without looking at the PR 🙈 ¯\_(ツ)_/¯

If you would like, please share your thoughts, probably best on Zulip in the linked topic :)

(this was written while in a call with @blinry)

alerque:
Orthogonal to the LLM issue, I do agree that being able to use a relay other than Iroh hosted ones is something that should definitely be included.

blinry:
We decided to adopt an updated LLM policy, which says that we do not accept any contributions that contain content generated by LLMs: #519

We still want this feature, though! The tracking issue is this one: #344

@martengooz, we hope you understand. Thanks for your work on this so far!

PR #514, opened 2026-04-15
Add cargo-deny tooling and address security issues it highlights

alerque:
Iroh is badly out of date and needs addressing but it is too late tonight...

**Self-review checklist**

- [-] I have used generative AI/LLMs for producing this PR. Here's which tool I used, and to which extent: None

### DIFF
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -72,7 +72,7 @@ clap_mangen = "0.3"
 [dependencies]
 anyhow = "1"
 async-trait = "0.1"
-automerge = "0.7"
+automerge = "0.8"
 clap = { version = "4.5", features = ["derive", "env"], optional = true }
 constant_time_eq = "0.4"
 derive_more = { version = "2", features = ["as_ref", "deref", "display"] }
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -67,7 +67,7 @@ too_many_lines = "allow"
 [build-dependencies]
 clap = { version = "4", features = ["derive", "env"] }
 clap_complete = "4"
-clap_mangen = "0.2"
+clap_mangen = "0.3"
 
 [dependencies]
 anyhow = "1"
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: 2024 Danny McClanahan <dmcC2@hypnicjerk.ai>
 # SPDX-FileCopyrightText: 2024 blinry <mail@blinry.org>
 # SPDX-FileCopyrightText: 2024 zormit <nt4u@kpvn.de>
+# SPDX-FileCopyrightText: 2026 Caleb Maclennan <caleb@alerque.com>
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
@@ -69,49 +70,49 @@ clap_complete = "4"
 clap_mangen = "0.2"
 
 [dependencies]
-anyhow = "1.0.81"
-async-trait = "0.1.88"
-automerge = "0.7.3"
-clap = { version = "4.5.3", features = ["derive", "env"], optional = true }
-constant_time_eq = "0.4.2"
-derive_more = { version = "2.0.1", features = ["as_ref", "deref", "display"] }
-dissimilar = "1.0.9"
-futures = { version = "0.3.30", default-features = false }
-git2 = { version = "0.20.2", default-features = false }
-ignore = "0.4.22"
-iroh = "0.35.0"
-magic-wormhole = "0.7.6"
-microxdg = "0.2.0"
-notify = { version = "8.0.0", default-features = false, features = [
+anyhow = "1"
+async-trait = "0.1"
+automerge = "0.7"
+clap = { version = "4.5", features = ["derive", "env"], optional = true }
+constant_time_eq = "0.4"
+derive_more = { version = "2", features = ["as_ref", "deref", "display"] }
+dissimilar = "1"
+futures = { version = "0.3", default-features = false }
+git2 = { version = "0.20", default-features = false }
+ignore = "0.4"
+iroh = "0.35"
+magic-wormhole = "0.7"
+microxdg = "0.2"
+notify = { version = "8", default-features = false, features = [
     "macos_fsevent",
 ] }
-operational-transform = "0.6.1"
-path-clean = "1.0.1"
-postcard = { version = "1.1.1", features = ["alloc"] }
-rand = "0.8.5"
-ropey = "1.6.1"
-rust-ini = "0.21.1"
-serde = { version = "1.0.198", features = ["derive"] }
+operational-transform = "0.6"
+path-clean = "1"
+postcard = { version = "1", features = ["alloc"] }
+rand = "0.8"
+ropey = "1"
+rust-ini = "0.21"
+serde = { version = "1", features = ["derive"] }
 serde_json = "1"
-tempfile = "3.24.0"
-time = { version = "0.3.36", features = ["formatting", "macros"] }
-tokio-util = { version = "0.7.11", features = ["codec"] }
-tracing = "0.1.40"
-tracing-subscriber = { version = "0.3.18", default-features = false, features = [
+tempfile = "3"
+time = { version = "0.3", features = ["formatting", "macros"] }
+tokio-util = { version = "0.7", features = ["codec"] }
+tracing = "0.1"
+tracing-subscriber = { version = "0.3", default-features = false, features = [
     "ansi",
     "fmt",
     "time",
     "env-filter",
 ] }
-url = "2.5.7"
+url = "2"
 
 [dependencies.tokio]
 version = "1"
 features = ["rt", "io-std", "io-util", "macros", "net", "time"]
 
 [dev-dependencies]
-pretty_assertions = "1.4.0"
-tracing-test = "0.2.4"
+pretty_assertions = "1"
+tracing-test = "0.2"
 
 # Stripping symbols (not debuginfo) reduces binary size by ~1.5x.
 # This is the default profile for `cargo install`:
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -80,8 +80,8 @@ dissimilar = "1"
 futures = { version = "0.3", default-features = false }
 git2 = { version = "0.20", default-features = false }
 ignore = "0.4"
-iroh = "0.35"
-magic-wormhole = "0.7"
+iroh = "0.35" # https://www.iroh.computer/blog/iroh-0-90-the-canary-series
+magic-wormhole = "0.7" # 0.8 fails on macos due to a bad interaction with iroh's netwatcher usage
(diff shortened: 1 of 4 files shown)


blinry:
Just quickly regarding Iroh, we hoped to be able to stay on the 0.35 series until 1.0, see https://www.iroh.computer/blog/iroh-0-90-the-canary-series

Would be a breaking change for Teamtype otherwise, as the Iroh versions are not compatible with each other.

alerque:
Fair enough. I kind of figured that might be the case after looking at the release history. I guess whether it makes sense to keep to that plan or not is largely dependent on where they actually end up in their release cycle. They are already 6 months behind where they projected and the known security issues with the 0.35 release and it's dependencies are piling up. At this point the three known vulnerabilities in the dependency tree are all TLS related and at least a cursory review of them suggests they are unlikely to be relevant to TeamType's usage — it just makes it a bit awkward reviewing when such a large tree of dependencies is so badly out of date. Note it is also bloating the surface area quite a bit because there are dozens of duplicate crates with major version differences where we're compiling in both old and new versions.

alerque:
Are you sure about your assertion that bumping iroh to the 0.90 would be a breaking change for Teamtype? I haven't made it though the whole code base yet but I haven't seen anywhere iroh types or functions are directly exposed or re-exported. That's good, and also presumably not what you mean. The way I read your note mean the communication between TT clients would no longer be compatible so everybody on TT would be required to upgrade to matching versions, but I don't see where that is currently being checked/enforced on our side. Is that something iroh itself is catching in negotiation and would raise an error for? Do we know what it would actually look like in practice right now to and end user if we did do this? Would it panic or fail to connect with a cryptic error or would be actually get something sensible about incompatible protocols versions? I see one of the major changes to iroh is that everything now returns concrete error types. Does that mean we'd be able to catch this and show a proper error on the new client side if we release something based on 0.9x, but the old version compiled with 0.35 would get some cryptic crash or failure?

alerque:
The `iroh` issue is clearly out of scope for this PR (and the discussion I started should probably be moved elsewhere).

I believe the rest of this is ready to go. Each commit is atomic and passes tests and lints separately. Especially bumping the lock file first, then bumping dependencies to see the changes in the deeper dependency tree and effect on advisory status is useful history to have, so please don't squash this ;-)

alerque:
Oh fun! It looks like some crate has broken semver compatibility, but only for some platforms. I'll look into why macos is failing.

alerque:
The macos compile issue bisects to what is currently afc6dc5 in this PR (demonstrated by test CI run [here](https://github.com/alerque/teamtype/actions/runs/24510170066/job/71638870750)). The compilation failure is in netwatch, but the bump of magic-wormhole did *not* change the version of netwatch being used in the lock file. The rabbit hole goes deeper...

alerque:
I'm dropping the problematic bump of magic-wormhole for another adventure since it has its own set of problems. There are some upstream bugs — that may or may not have been worked out already in later versions of deep dependencies that are held back due to our iroh issue.

alerque:
Removing the commits I broke out into #516 from this PR revealed what might be a cross talk problem between PRs. It is possible the [timeout on this CI run](https://github.com/teamtype/teamtype/actions/runs/24562986177/job/71816013501?pr=514) is just an ephemeral GHA issue, but I've seen it happen several times now when I have recently pushed multiple branches –and hence have multiple runners going– and my spidey-senses are going up.

Is it possible the test suite somewhere/somehow spawns some client acting on a hard coded share address and we're getting cross talk between runners?

daemon/deny.toml
```diff
@@ -0,0 +1,46 @@
...
+[advisories]
+ignore = [
+	"RUSTSEC-2023-0089", # unmaintained: atomic-polyfill (via iroh and postcard)
+	"RUSTSEC-2024-0384", # unmaintained: instant (via iroh)
+	"RUSTSEC-2024-0436", # unmaintained: paste (via iroh)
+	"RUSTSEC-2025-0052", # unmaintained: async-std (via magic-wormhole)
+	"RUSTSEC-2026-0049", # vulnerable: rustls-webpki (via iroh)
+	"RUSTSEC-2026-0067", # vulnerable: tar (via magic-wormhole)
+	"RUSTSEC-2026-0068", # vulnerable: tar (via magic-wormhole)
+	"RUSTSEC-2026-0097", # unsound: rand (via iroh)
```
blinry: When I run `cargo deny check`, I get:

```
warning[advisory-not-detected]: advisory was not encountered
   ┌─ /home/blinry/wip/teamtype/daemon/deny.toml:19:3
   │
19 │     "RUSTSEC-2026-0097", # unsound: rand (via iroh)
   │      ━━━━━━━━━━━━━━━━━ no crate matched advisory criteria
```

Is it still necessary to include it? Can advisories different per platform? I also think the advisory has been updated recently.

daemon/deny.toml
```diff
@@ -0,0 +1,46 @@
...
+[graph]
+all-features = true
+exclude-dev = true
+
+[bans]
+multiple-versions = "allow"
+
+[advisories]
+ignore = [
+	"RUSTSEC-2023-0089", # unmaintained: atomic-polyfill (via iroh and postcard)
```
blinry: Just a note on consistency: All other files in the daemon directory use 4 spaces for indentation. I'd slightly prefer to do that here, as well.

daemon/Cargo.toml
```diff
@@ -67,7 +67,7 @@ too_many_lines = "allow"
 [build-dependencies]
 clap = { version = "4", features = ["derive", "env"] }
 clap_complete = "4"
-clap_mangen = "0.2"
+clap_mangen = "0.3"
```
blinry, line 70: Nitpick: The commit message says it's being updated to 0.4.

blinry (approved):
Thanks for working on this! These updates all seem safe to me. Good idea to do the more complex ones in separate PRs.

Regarding reducing the version detail in Cargo.toml – I wonder whether there's a way to tell cargo to do that automatically, when we add a new dependency in the future…

Left a number of other comments, but nothing important.

> Is it possible the test suite somewhere/somehow spawns some client acting on a hard coded share address and we're getting cross talk between runners?

I don't think we do any hardcoding of iroh addresses. The tests that time out don't even involve the daemon, I think, they test the Neovim plugin! So that seems unrelated/a bit flaky.

blinry:
Oh, and another thought: Do we maybe want to run `cargo deny` automatically in the CI? :thinking:

daemon/deny.toml
alerque: I suppose they might be different per platform (although I would not have expected it), but the reported library here is used on all platforms. I definitely still get this advisory flagged, and removing it here is an error.

You can see in e.g. my iroh PR draft (#515) that I've been removing advisory exceptions in the commit that addresses them.

A more likely candidate might be your Rust tool chain version or `cargo-deny` version. I'm on 1.95.0 and 0.19.4 respectively, you?

daemon/deny.toml
alerque: Gah, I thought I fixed this already but something (TOML LSP maybe?) keeps deciding not to follow the style. I'll amend to fix this.

alerque:
> Oh, and another thought: Do we maybe want to run `cargo deny` automatically in the CI? 🤔

Yes, but I have some other changes to CI I thought that would fit better with in another PR. E.g. at least some of the jobs need to use `--locked`. I don't like making CI changes inside a PR that makes a bunch of other changes that existing CI should be passing on because it makes it hard to debug/review if jobs fail (or pass for that matter).

daemon/Cargo.toml
alerque: :facepalm:

alerque:
> Regarding reducing the version detail in Cargo.toml – I wonder whether there's a way to tell cargo to do that automatically, when we add a new dependency in the future…

Personally I just add dependencies by editing the manifest manually, but if you use `cargo add` it will respect whatever version spec format you give it, e.g. `cargo add rand@0.8`. I don't know of a way to configure a project (without external tooling) such cargo itself enforces the version specs not include segments that will be effectively dropped anyway.

By the way the detail I dropped isn't 100% meaningless, technically this format does specify a *minimum* supported version. It is **possible** to have a scenario in the case of known bug fixes in dependencies that a project might rely on a bug fix and hence *want* that detail in the manifest. Many/most projects just add whatever minor/patch version is current at the time they add the dependency and then move on. The issue I have with that is it makes it *seem* like they are specifying a minimum version for a dependency when in reality they have no idea if that is actually the case or not. Many/most times if semver is actually being followed they would actually be fine with lower versions of dependencies too. This becomes an issue with distro packaging because MSRV bumps are not considered breaking in semver and hence ofter end up going through in patch versions. This quickly trickles down to projects that are not paying attention and just take whatever is current as their minimum supported version, thereby bumping their MSRV for no real reason.

In my book it's better to only specify the extra minimum version detail if you have verified it is necessary for some reason. For the general use case of "this is what we've tested with", the lock file covers that anyway. The lock file has what we have tested and recommend. If people want to build with *newer* deps than the lock file they are free to do so, but also if they need to downgrade for some reason (e.g. to use an older Rust toolchain, to resolve a linking issue with shared libraries, etc.) they also can.

daemon/deny.toml
alerque: The hot path for use of `iroh`'s use of `rand` is *not* the same across platforms so it is actually conceivable that the advisory does not apply equally across platforms. I'm still surprised the tooling doesn't just lump them together since they clearly still apply to the project.

alerque: ...and yes cargo deny does use the current host tuple for dependency graph resolution. It can also be configured to check for specific platforms, so I'll add that to the config so we can all get consistent results.

alerque: On further review, maybe not touching it is better.

So the deal is by default cargo-deny checks *ALL* the targets known to the toolchain and resolution system ([source](https://embarkstudios.github.io/cargo-deny/checks/cfg.html#the-targets-field-optional)). The likely difference between our systems is I have toolchains installed for several target platforms including WASM and RiscV that you likely don't have, as well as a fresh Rust toolchain that includes knowledge of new platforms your (possibly older) toolchain doesn't even know about yet.

The obvious first thing to do would be to specify a few platforms we care about and limit Cargo deny to those. I tried adding the whole Tier-1 list and immediately ran into two advisories that were then unused. But they *are* relevant to Teamtype because they affect WASM, RiscV, and ARMv6 targets. There is no reason we shouldn't get warned about such things even if they are low usage at this time.

I think the best thing to do here is *nothing*. Checking the whole dependency tree seems like the best thing to do by default. I can't think of a good reason to arbitrarily limit the checks except for developer convenience—and that doesn't seem like a good enough reason.

If we have problems with this in CI later we can tell the CI command (using args, not the config) about specific unused advisories it is allowed to ignore entirely based on whatever targets it isn't able to review in the CI environment.

alerque: I marked the other review comments as resolved because they were trivial and obvious fixes. I'll leave this one for you to dismiss in case you want to pursue idea further. As far as I'm concerned I would resolve it by leaving it alone for now.

blinry (approved):

zormit (approved):

PR #515, opened 2026-04-16
Upgrade to Iroh 1.0 (now that the time has come)

alerque:
This revives some previous work to bring iroh up to speed, rebased on other current work. See discussion starting in #514. This PR is just to have a place to keep the branch, it isn't ready for merge yet, only for experiments and keeping an eye on how iroh is shaping up.

**Self-review checklist**

### DIFF
--- a/crates/teamtype/Cargo.toml
+++ b/crates/teamtype/Cargo.toml
@@ -42,7 +42,7 @@ cli = ["dep:clap", "tokio/signal", "tokio/rt-multi-thread"]
 
 [build-dependencies]
 anyhow.workspace = true
-automerge = "0.8"
+automerge = "0.10"
 clap_complete = "4"
 clap_mangen = "0.3"
 regex = "1"
@@ -58,7 +58,7 @@ features = ["emit_and_set"]
 [dependencies]
 anyhow.workspace = true
 async-trait.workspace = true
-automerge = "0.8"
+automerge = "0.10"
 constant_time_eq = "0.4"
 dissimilar = "1"
 futures.workspace = true
--- a/crates/teamtype/Cargo.toml
+++ b/crates/teamtype/Cargo.toml
@@ -67,7 +67,7 @@ ignore = "0.4"
 # https://www.iroh.computer/blog/iroh-0-90-the-canary-series
 iroh = "0.35"
 # We cannot bump to 0.8 because Iroh 0.35's netwatcher usage causes mismatched types on macOS.
-magic-wormhole = "0.7"
+magic-wormhole = "0.8"
 microxdg = "0.2"
 operational-transform = "0.6"
 path-clean = "1"
--- a/crates/teamtype/Cargo.toml
+++ b/crates/teamtype/Cargo.toml
@@ -65,19 +65,20 @@ futures.workspace = true
 ignore = "0.4"
 # Iroh is currently held back to the 0.3x series while 0.9x are canary releases:
 # https://www.iroh.computer/blog/iroh-0-90-the-canary-series
-iroh = "0.35"
+iroh = "0.98"
 # We cannot bump to 0.8 because Iroh 0.35's netwatcher usage causes mismatched types on macOS.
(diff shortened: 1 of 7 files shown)


alerque:
I just rebased this and added the latest Iroh release just to keep an eye on where things are at, and a couple notes:

* iroh 0.98 is working to stop exposing vendored types (e.g. exposing types from the rand crate) and this is good.
* magic-wormhole 0.8 brought an MSRV bump to 1.92. I've applied that here but we should probably discuss MSRV policy.
* constant_time_eq 0.5 went a little crazy with a 1.95 MSRV. We can probably forgo bumping this dep until we don't care about that, but I did [ask about that upstream too](https://github.com/cesarb/constant_time_eq/issues/14).

alerque:
Rebased to get our new custom relay/dns/pkarr url options ported to the new Iroh releases. The *documentation* for Iroh's API changes is lacking, but the new API design in 0.90 is definitely better than 0.35! As yet untested but it builds.

alerque:
It seems obvious (at least from my perspective) that #543 or some iteration thereof will land long before we're ready for the Iroh bump, so I'm going to go ahead and stack this on that just for one less rebase to baby sit.

alerque:
Unless we're [thinking of another patch release](https://teamtype.zulipchat.com/#narrow/channel/557642-general/topic/Next.20release.20target.20level.2C.20major.20or.20another.20quick.20minor.3F/with/611255509) I think this is ready for review.

crates/teamtype/Cargo.toml
```diff
@@ -17,11 +17,11 @@ documentation = "https://teamtype.github.io"
...
-    "local-first",
-    "synchronization",
-    "crdt",
-    "peer-to-peer",
+	"collaboration",
+	"local-first",
+	"synchronization",
+	"crdt",
+	"peer-to-peer",
 ]
```
blinry, line 25: There was an accidental indentation change here. Should we set up a TOML formatter? :P

crates/teamtype/Cargo.toml
```diff
@@ -65,19 +65,20 @@ futures.workspace = true
 ignore = "0.4"
 # Iroh is currently held back to the 0.3x series while 0.9x are canary releases:
 # https://www.iroh.computer/blog/iroh-0-90-the-canary-series
-iroh = "0.35"
+iroh = "0.98"
```
blinry: Nit: The commit description says it's updating to 0.90.

blinry (approved):
Looking very good to me. Besides reviewing the code, I tried using this Teamtype version manually, and it all seemed to work as before.

One thing that would be nice is letting users know when they're trying to connect to an incompatible Iroh node. The current behavior when trying to connect between 0.9.2 and main is that all attempted connections time out. One "lightweight" approach to this would be to extend our "Failed to connect to specified peer" error message in `daemon.rs` to say "Does the Teamtype you're trying to connect to use the same 0.x version?" This would only work on the *connecting* side, though. But we can do this in another PR.

Another follow-up thing: We need to update and re-test our instructions on [running custom relays](https://teamtype.github.io/teamtype/relays.html) to Iroh 1.0.

crates/teamtype/Cargo.toml
alerque: Bah. This is my editor config getting in the way.

I'm not sure TOML formatting is a win right now. The ones I use are a bit inconsistent and I don't know of anything that consistently matches Cargo's own outputs (or the ones from `cargo edit` etc.)

I should probably setup my Neovim config with per-project disable for the TOML LSP because this isn't the only project I've had trouble with it being too aggressive.

crates/teamtype/Cargo.toml
alerque: Oops. This was a squash of several commits working through the 0.9x series bumps that I eventually had to merge because the lints and advisories introduced at intermediate bumps during the testing lifecycle on this were a constant pain point when rebasing. I'll fix the message.

PR #516, opened 2026-04-17
Cleanup integration test crate following daemon changes

alerque:
This is another follow up on #514 (to be merged afterwards) that I split out from there because it has a different set of issues with CI.

**Self-review checklist**

- [-] I have used generative AI/LLMs for producing this PR. Here's which tool I used, and to which extent: None

### DIFF
--- a/daemon/integration-tests/Cargo.lock
+++ b/daemon/integration-tests/Cargo.lock
@@ -27,7 +27,7 @@ checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
- "cpufeatures",
+ "cpufeatures 0.2.17",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


zormit:
So should it be *based on* the other PR? I didn't look very closely, but some of the PRs seem to be overlapping?

alerque:
Yes, I'm doing some other work deliberately based on the branch in #514 knowing that needs to be dealt with first and then I'll clean up/finish up these.

For things that don't happen to touch the same parts I'm branching from *main*, but some of the things I'm working on do overlap a lot, hence it is easier to branch from the other work and eventually rebase on it after merge.

alerque:
Still working on CI stuff elsewhere, but the issues I'd previously seen with this in CI were, I now believe, just the fuzzer stability issue.

This doesn't do much other than cleanup the manifest and review what can and can't be bumped where, but it should be safe to proceed with as a base for further work on the test runner.

blinry:
All reasonable changes, thank you! :)

PR #522, opened 2026-04-22
Move hard coded initial document generation to build process

alerque:
~~I can't find any call sites or documentation for this boiler plate. Was this an experiment?~~

This gets rid of the extra unpublished utility crate used for generating an initial document state (which was then copy/pasted into code elsewhere) in favor of an initial document tracked in VCS and an optional build-time regeneration function that rebuilds it. 

**Self-review checklist**


zormit:
It's indeed not very well documented, thanks for uncovering tacit knowledge. Happy to change that and make it explicit, but I wouldn't delete this code.

Compare the output of the automerge-baker with

https://github.com/teamtype/teamtype/blob/e159d4f4f139faf14e4bf89ce69a57b174d75b16/daemon/src/document.rs#L46-L54

=> We have been using it to pre-generate an initial "shared truth" document that contains both of the currently used keys.

When we were writing this I / we didn't know about build scripts... Maybe there's a way to integrate this code into https://github.com/teamtype/teamtype/blob/main/daemon/build.rs and then load the pre-baked document from the rustc-env during compilation?

alerque:
I grepped for that string, but apparently I used too long a segment and the wrapping is different. My bad.

But yes that can and should go in the build script ... or a proc macro that generates a constant or something like that. I can look into that.

alerque:
This isn't right yet. It shouldn't just regenerate if it doesn't exist because part of the generation process is a random seed. Since we don't have the initial seed we can't regenate this even to test this code.

1. I need to use the previously hard coded vec as the current initial doc, not generate a new one.
2. Look into whether we can stash and use our own random seed instead of a new one on each generation so that potentially in the future we could verify the automerge document structure hasn't changed between versions.

alerque:
Okay I've used the original byte-for-byte blob for the initial document, so this shouldn't break anything. I also added some validation so that we assert during the build that the initial state is actually a valid document with a single change root. It doesn't currently assert the hash of that root, which we could do as an extra way to prevent the root document changing without deliberately updating the hash. Currently one would have to notice that the bin file has been modified in the Git working dir and commit the change. It is possible to regenerate the root document from the build process (remove the bin file or set the env var, then build) but in normal development cycles this shouldn't be used until such a time as there is a good reason too.

Also I don't see a way to use our own random seed to actually regenerate a document and see if the structure of the empty document changes between Automerge versions, but I guess following their breaking/non-breaking versions and release documentation should serve our use case.

daemon/src/initial_automerge_doc.bin
blinry, line 1: Should we call this `initial_automerge_doc.bin`, to make clearer what it is to people who see it for the first time?

(I'm aware that there's `.teamtype/doc`, but nothing stops us from making better naming choices now. :P)

daemon/build.rs
```diff
@@ -1,15 +1,83 @@
...
+    let docfile = "initial_doc.bin";
+    let src_dir = Path::new("src");
+    let initial_doc_path = src_dir.join(docfile);
+    println!("cargo:rerun-if-changed={}", &initial_doc_path.display());
+    let force_regenerate = var_os("TEAMTYPE_REGENERATE_AUTOMERGE_DOC").is_some();
+    // The initial Automerge document state is hard coded and changing it would imply a breaking
+    // Teamtype release incompatible with old clients. Hence, the document generated here is tracked
+    // in our VCS so we can see if it ever needs to be bumped, e.g. to use a new breaking version of
+    // Automerge. Most people will not need to regenerate it, but removing the file or setting an
+    // env var can be used to force a regeneration to confirm that it has not changed.
```
blinry: Just commenting to say that I like the very explicit documentation here. It feels a bit bad to have a binary blob in a project, so trying to be clear about it seems good.

daemon/build.rs
```diff
@@ -1,15 +1,83 @@
...
+    if !initial_doc_path.exists() || force_regenerate {
+        let mut doc = AutoCommit::new();
+        doc.put_object(automerge::ROOT, "files", ObjType::Map)?;
+        doc.put_object(automerge::ROOT, "states", ObjType::Map)?;
```
blinry: Reading this, it occurred to me that we don't need the "states" map anymore. (We used to store cursor positions in it before we had "ephemeral" messages between peers.) See #291.

In the source code, there's no reference to that map anymore. Compatibility with really old versions seems irrelevant, as we just need to care about the 0.9.x series.

So we could remove it here. The compatibility of independently created Automerge documents is per-key, I think. We could confirm this by removing "states" here, then start two new Teamtype directories, each with a different file, running `teamtype share` on them for the initial loading, then syncing them together. As a result, both peers should have both files.

blinry (approved):
Neat cleanup of build.rs!

To protect against a Jia Tan-style binary takeover (😛), I confirmed that the bytes in the current `initial_doc.bin` are indeed the same as they were in the `initial_doc` array. For future reference: Its sha256 is `fc57f4cd3e2509a73bb5f4f29691118c86af521ddefe98036beca9fafdc6db93`.

Couple of comments, but nothing blocking. Feel free to not address them, depending on your available time/enjoyment of that task.

alerque:
>  Jia Tan-style binary takeover

Yup yup. Because with the package singing keys to instantly publish modified versions of any system package to unknown millions of Arch Linux users and downstream distros –not to mention the access to push Harfbuzz releases that go to *billions* of devices– I'd really be out here trying to take over the world by sneaking a different document ID into an Automerge base document. :grinning:

daemon/src/initial_automerge_doc.bin
alerque: We can. I didn't like the existing mix of verbose vs. abbreviated names in the code related to this myself, so maybe I'll go expand them all.

daemon/build.rs
alerque: Yes. I considered serializing it to a text representation in a file, but there really doesn't seem to be a good reason to add the complexity. Of course having binary blobs around isn't something you just want to do for convenience, but in this case all of the generation method, the ingest method, the eventual over-the-wire representation (sans encryption), and internal parsing operations are all done on this binary format, so I think it makes sense.

alerque: Also this comment is inaccurate, you can't confirm by regeneration it that it has not been changed since it uses a new random seed each time.

daemon/build.rs
alerque: I think we need to add that to a list of things for our next breaking release. Since we don't have the initial random seed available I don't think we can generate a new document to make that change without breaking compatibility with existing clients, including 0.9.x ones.

When we do regenerate this to remove the states object or whatever else we should save the initial document in the bin blob instead of the modified one with added roots. That would give us the option to change those later without breaking compatibility again.

blinry: Oh right! Interesting. I quickly added this to our internal release instructions document. Maybe we can make that more public in the future. But this also seems like a very unimportant cleanup. At some point, we'll probably want to make larger changes to the doc structure, for example to track the executable bit of files and symlinks.

alerque:
Renamed a few things to be a little more perspicuous and fixed the inaccurate bit of code comment. Should be go for merge.

zormit (approved):
I read through this, didn't double check the results, but I don't think that's necessary twice. Thanks for the cleanup, looks neat!

PR #517, opened 2026-04-17
Make the iroh-relay and the iroh-dns-server configurable.

ChristophNiehoff:
We would like to be able to teamtype completely self-hosted (i.e. without using magic wormhole or iroh servers).
This PR makes the iroh relay and the iroh DNS discovery server configurable, both, via CLI arguments and in the config file

**Self-review checklist**

- [X ] I described the changes of the PR in the [CHANGELOG](https://github.com/teamtype/teamtype/blob/main/CHANGELOG.md).
- [ X] I have manually tested and self-reviewed my changes.
- [ X] I have added myself to the REUSE headers in the files where I made significant changes.
- [ X] I have used generative AI/LLMs for producing this PR. Here's which tool I used, and to which extent:
  - OpenCode with Anthropic Claude Sonnet 4.6 and Claude Opus 4.6  
  - [X ] I have reviewed the output carefully, and I understand the resulting changes in detail. I am responsible for the changes, and I made sure that my PR represents excellent work.

PR #523, opened 2026-04-23
Make the Iroh relay configurable

blinry:
This addresses #344, and is an alternative to #512 and #517.

## Testing

To try this:

- clone https://github.com/n0-computer/iroh
- check out the `v0.35.0` tag
- go to `iroh-relay/`
- run `cargo run --features=server -r -- --dev`
- run `teamtype share --iroh-relay http://localhost:3340`, or the respective `join` command

Seems to work for me! :tada:

(See below for DNS + Pkarr relay.)

## Still to do

- [x] Figure out whether we also need/want a flag for custom servers for discovery. -> definitely
- [x] Confirm whether Teamtype then really works without any external servers.
- [x] Document how to run your own relays.

Ideally, we'd provide a `teamtype relay` subcommand, which spins up all required relays and services ourselves – let's probably do that in a follow-up PR.

## Self-review checklist

### DIFF
--- a/book/src/relays.md
+++ b/book/src/relays.md
@@ -1,0 +1,98 @@
+<!--
+SPDX-FileCopyrightText: 2026 blinry <mail@blinry.org>
+SPDX-FileCopyrightText: 2026 zormit <nt4u@kpvn.de>
+
+SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Running your own relays
+
+By default, Teamtype uses a couple of public relays hosted by other people/groups. They facilitate smooth connections between devices in different local networks. All transmitted data is always end-to-end encrypted between peers. After a connection is established, peers connect directly, if possible.
+
+If you don't want to rely on these public relays, you can host your own ones, and instruct Teamtype to use them. Here's how that works for each of the relays. Note that we mostly describe a "development testing setup" here; for a production deployment, please look into the individual relays more carefully.
+
+## Magic Wormhole relay
+
+Teamtype uses a [Magic Wormhole relay server](https://github.com/magic-wormhole/magic-wormhole-mailbox-server) (also known as "mailbox server" or "rendezvous server") to share the initial join codes with other peers.
+
+### Running your own
+
+Install [uv](https://docs.astral.sh/uv/), then run:
+
+    uvx --from twisted --with magic-wormhole-mailbox-server twist wormhole-mailbox
+
+This will start the relay on port 4000 by default.
+
+### Using it
+
+Run
+
+    teamtype share --magic-wormhole-relay ws://localhost:4000/v1
+
+and
+
+    teamtype join --magic-wormhole-relay ws://localhost:4000/v1 <join code>
+
+or set the `magic_wormhole_relay` parameter in `.teamtype/config`.
+
+## Iroh Pkarr relay + DNS server
+
+Teamtype uses Iroh's [Pkarr relay and DNS server](https://github.com/n0-computer/iroh/tree/main/iroh-dns-server) to allow peers to meet on the same Iroh relay (see below). Read [Iroh's discovery documentation](https://docs.iroh.computer/concepts/discovery) for more information.
+
+### Running your own
+
+Using Rust's `cargo`, run `cargo install iroh-dns-server@0.35.0`. Then, run:
+
+    iroh-dns-server
+
+This will start a Pkarr relay on port 8080, and a DNS server on port 5300 by default, which resolves subdomains of `irohdns.example`. You can provide configuration files for production setups, see the [README of iroh-dns-server](https://github.com/n0-computer/iroh/tree/main/iroh-dns-server).
+
+### DNS testing setup
+
+For a development setup, you need to configure your computer to use your local DNS server. Assuming you're on Linux and use `systemd-resolved`, run
+
+    sudo resolvectl dns <interface> 127.0.0.1:5300
+
+to set sets it as a DNS server for a network interface, run `resolvectl status` for your options. Then, run
+
+    sudo resolvectl domain <interface> '~irohdns.example'
+
+so that only domains under `irohdns.example` are resolved using your local server.
+
+### Using it
+
+Run
+
+    teamtype share --iroh-pkarr-relay http://localhost:8080/pkarr
+
+and
+
+    teamtype join --iroh-dns-domain irohdns.example <join code>
+
+or set the `iroh_pkarr_relay` and `iroh_dns_domain` parameters in `.teamtype/config`.
+
+It doesn't hurt to use both flags on both the sharing and the joining peer.
+
+## Iroh relay
+
+Teamtype uses an [Iroh relay](https://github.com/n0-computer/iroh/tree/main/iroh-relay) when direct connections between peers aren't immediately possible. Read [Iroh's relay documentation](https://docs.iroh.computer/concepts/relays) for more information.
+
+### Running your own
+
+Using Rust's `cargo`, run `cargo install cargo install iroh-relay@0.35.0 --features=server`. Then, run:
+
+    iroh-relay --dev
+
+This will start the relay on port 3340 by default.
+
+### Using it
+
+Run
+
+    teamtype share --iroh-relay http://localhost:3340
+
+and
+
+    teamtype join --iroh-relay http://localhost:3340 <join code>
+
+or set the `iroh_relay` parameter in `.teamtype/config`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Add support for JSON-RPC IDs in string to be compliant with the [JSON-RPC specif
 
 Add support for `.teamtypeignore` file to exclude files and directories from synchronization. The syntax is similar to `.gitignore`, supporting file names, directories, and glob patterns.
 
+Add support for custom Iroh relay servers, using the `--iroh-relay` flag.
+
 ## Note for package maintainers: Renamed the Linux binaries
 
 We've renamed the Linux binaries
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Add support for JSON-RPC IDs in string to be compliant with the [JSON-RPC specif
 
 Add support for `.teamtypeignore` file to exclude files and directories from synchronization. The syntax is similar to `.gitignore`, supporting file names, directories, and glob patterns.
 
-Add support for custom Iroh relay servers, using the `--iroh-relay` flag.
+Add support for custom Iroh relay servers, using the `--iroh-relay` flag, and for custom Iroh DNS discovery servers, using `--iroh-dns-domain` and `--iroh-pkarr-relay`. See the "Running your own relays" page in the documentation for how to run all these relays yourself!
 
 ## Note for package maintainers: Renamed the Linux binaries
 
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -29,6 +29,7 @@ SPDX-License-Identifier: CC-BY-SA-4.0
     - [Working with Git](git-integration.md)
         - [Manual](git-integration-manual.md)
         - [Synchronized](git-integration-synchronized.md)
+    - [Running your own relays](relays.md)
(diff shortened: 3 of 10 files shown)


blinry:
The [Iroh documentation](https://docs.iroh.computer/concepts/discovery) explains the Pkarr + DNS construction for discovery pretty well!

The code for the server is here: https://github.com/n0-computer/iroh/tree/main/iroh-dns-server

Should be a straightforward thing to add. :) The [docs for `n0_discovery()`](https://docs.rs/iroh/0.35.0/iroh/endpoint/struct.Builder.html#method.discovery_n0) give hints for the Teamtype-side implementation.

blinry:
I added the flags `--iroh-dns-domain` and `--iroh-pkarr-relay`.

They seem to default to `dns.iroh.link` and `https://dns.iroh.link/pkarr`.

Here's how to test these flags:

- clone https://github.com/n0-computer/iroh
- go to `iroh-dns-server/`
- run `cargo run -r -- -c config.dev.toml`
- if you now run `teamtype share --iroh-pkarr-relay http://localhost:8080/pkarr`, you'll see in the logs of the DNS server that the peer ID is published there
- now we need to configure our system so that it uses our local DNS server:
    - `sudo resolvectl dns wlp1s0 127.0.0.1:5300` sets it as a DNS server for my wireless network card (yours might be called differently, check `ip a`)
    - `sudo resolvectl domain wlp1s0 '~irohdns.example'` should set it so that only <someting>.irohdns.example addresses are resolved using that server
    - you can double-check these settings using `resolvectl status`, and undo them with `resolvectl revert wlp1s0`.
- `teamtype join --iroh-dns-domain irohdns.example <join-code>` now succeeds, and uses our local DNS server!

To "manually" check that you can receive DNS information, use a query like `dig @127.0.0.1 -p 5300 _iroh.pgjce9h5owpyqby4zbwbbnjodnup1ryhyenaygw4chgcb86mrino.irohdns.example TXT`.

blinry:
I added a page to the book that explains how to run each of our 4 (:scream:) relays yourself.

Here's how to confirm that, using all of these, Teamtype no longer needs Internet to make a local connection, on Linux (I'm not completely sure that this is right?):

- `sudo iptables -P OUTPUT DROP` will drop all outgoing packets
- `sudo iptables -A OUTPUT -o lo -j ACCEPT` makes an exception for localhost

(To undo, run `sudo iptables -P OUTPUT ACCEPT`.)

Spin up all relays; then run this monster:

    teamtype share --magic-wormhole-relay ws://localhost:4000/v1 --iroh-pkarr-relay http://localhost:8080/pkarr --iroh-dns-domain irohdns.example --iroh-relay http://localhost:3340

book/src/relays.md
```diff
@@ -0,0 +1,91 @@
...
+
+    teamtype share --iroh-pkarr-relay http://localhost:8080/pkarr
+
+and
+
+    teamtype join --iroh-dns-domain irohdns.example <join code>
+
+or set the `iroh_pkarr_relay` and `iroh_dns_domain` parameters in `.teamtype/config`.
+
+It doesn't hurt to use both flags on both sides.
```
alerque: Sides of what? Both config and CLI? And if that's the case what takes precedence if they are differente? This wording isn't clear to me.

alerque (changes requested):
Nice work on documenting all the self-hosting things. A little silly that it has to be so many services, but I get why we're here...

Just a nitpick on a bit that wasn't clear to me.

book/src/relays.md
blinry: Ah – I mean both the `share` and `join` side. Will clarify!

blinry:
In a review, I'd be happy about critical thinking regarding the names of the three new flags! They make sense to me, but maybe there are better ones?

zormit:
Thanks for finalizing our draft, that's a very helpfully documented PR!

A bit late-stage questioning: When reading this I was wondering if that feature is really what everyone wants... It feels indeed quite bulky.

If I get it right, the main use-case for this is when you're in a more closed (corporate) network. Wouldn't be an easier way there to allow for local connections and skipping all the relay/dns shenanigans? When you're sitting "next to each other" (ok, maybe still remote) and are in a closed network maybe it's okay to share a local address and not have the magic-wormhole experience?

So I see it as three steps of independence:
- 1. pre this PR: Completely dependent on 
- 2. this PR: Be able to self-host relays
- 3. potential future PR: "just" make a local connection, without relays.

I'm not sure how 3 works with Iroh, but my mental model is that it does work. Let's assume we would have it implemented, would (most) people still put in the effort of self-hosting their relays? Do we still need DNS? Are there people who would prefer to self-host their relays in general (in the open internet)?

This is a bit of a devil's advocate questioning, I'm not saying I'm against introducing this. Just wondering if we have explored all alternatives.

(I didn't think about naming yet, as I didn't dive deep into the code yet)

alerque:
> the main use-case for this is when you're in a more closed (corporate) network

That might be *a* use case, but it certainly isn't the only one. There are lots of folks concerned with privacy and meta data leakage for various reasons.

Also the corporate folks are not always local to each other. In fact I suspect many/most these days are remote. Some part of those will be VPNed into the same network, but corporate VPNs often segment or isolate peers from each other even when folks are technically on the same network.

Yes there may be a use case for direct access too, but I see that as somewhat orthogonal to this whole issue.

blinry:
I also think being able to configure all these relays is just plain good for the project! :) Even if just from a "Teamtype should still work if Iroh goes out of business" perspective. It would be a bit more convenient to just be able to specify an "all-in-one relay" with a single flag, I think, and if we build a `teamtype relay` subcommand that starts all of them, we could also do that! But for now, splitting them out to individual flags seems fine to me. The implementation doesn't feel that complex or bulky to me, it's mainly the endpoint configuration.

I also thought about local discovery last night, and it's extremely easy to add, see the (stacked) [local-discovery](https://github.com/teamtype/teamtype/tree/local-discovery) WIP branch. Didn't test that one properly yet. I think it's disabled by default to prevent "leaking" your node ID to the local network, but because we have our password protection, I think we could enable it. See the [mDNS docs](https://docs.iroh.computer/connecting/local-discovery) for how it works.

blinry:
Maybe one thought towards simplification: Instead of having the relay settings as command-line flags, we could (only) allow to specify them via the config. That would result in more overview when running `--help` (and reduce the "interface" of the software, which we could break), but make the options less visible.

Sooner or later I'd love a clean solution for our config file anyway, where you only need to specify them once in the code... and maybe having `[sections]` a la Git and jj, and maybe a `teamtype config` subcommand.

alerque:
> Instead of having the relay settings as command-line flags, we could (only) allow to specify them via the config.

As a general rule I'm against config options that cannot be overridden by CLI args at run time. That doesn't mean they have to all be explicit flags that show up in `--help` though. For example in my [`acceptarium`](https://codeberg.org/plaintextaccounting/acceptarium/) project I started doing what `git` also allows, which is a `-c` that allows access to *anything* in [the config file](https://codeberg.org/plaintextaccounting/acceptarium/src/branch/master/src/defaults.toml), including accessing both top level and a sectioned config keys:

```console
$ acceptarium -c blob-storage git-annex -c git-annex.directory bucket status
```

Or if you were looking at my #522 PR and wanted to audit the bin file changes but don't have my pre-configured Git arrangement, you could do this:

```console
$ git -c diff.hex.textconv='hexdump -v -C' -c diff.hex.binary=true show 36c4c2b -- daemon/src/initial_doc.bin
commit 36c4c2b
Author: Caleb Maclennan <caleb@alerque.com>
Date:   Thu Apr 23 13:45:49 2026 +0300

    refactor(daemon): Move initial automerge document creation to build time

diff --git a/daemon/src/initial_doc.bin b/daemon/src/initial_doc.bin
new file mode 100644
index 0000000..54041a6
--- /dev/null
+++ b/daemon/src/initial_doc.bin
@@ -0,0 +1,9 @@
+00000000  85 6f 4a 83 3d 9d e7 55  00 76 01 10 78 6b 68 2f  |.oJ.=..U.v..xkh/|
+00000010  d7 09 4c 20 84 88 3c 7c  98 78 90 b6 01 8f a4 1f  |..L ..<|.x......|
+00000020  0d 66 3d 8b 7d f6 bd 87  61 10 a7 3f 1e d7 f9 3c  |.f=.}...a..?...<|
+00000030  e3 71 6f 3d 37 8a ea 5e  1e 8e a6 4e fa 06 01 02  |.qo=7..^...N....|
+00000040  03 02 13 02 23 02 40 02  56 02 07 15 0e 21 02 23  |....#.@.V....!.#|
+00000050  02 34 01 42 02 56 02 80  01 02 7f 00 7f 01 7f 02  |.4.B.V..........|
+00000060  7f 00 7f 00 7f 07 7e 05  66 69 6c 65 73 06 73 74  |......~.files.st|
+00000070  61 74 65 73 02 00 02 01  02 02 00 02 00 02 00 00  |ates............|
+00000080
```

That way you don't have to have a CLI flag for every config option, but *anything* that can be configured in the conf can be configured from a command too. In the case of `acceptarium` it also accepts anything as an ENV var too which is handy for CI environments where it is possible to run the same pipeline with different vars depending on various circumstances:

```console
$ env ACCEPTARIUM_BLOB_STORAGE=git-annex ACCEPTARIUM_GITANNEX_DIRETORY=bucket acceptarium status
```

I'm not sure Teamtype would find as much utility in being headless/CI friendly, but the access to config values via a `-c` / `--config` flag might be worth doing. If you'd like I can look into implementing that since I've done it a few times already.

zormit:
I would be ok merging this and open new issues (or directly PRs) for
- the idea of a generic `-c` (and simplifying the flags exposed in `--help` at the same time)
- `teamtype relay`
- local discovery

alerque (approved):
I concur with Moritz, this can go in as is. Surfacing the feature for testing during the development cycle is probably more important that bike shedding about the best way to pass options. We also have until the next tagged release to adjust the UI if we feel the need and/or get around refactoring config handling. Short of completely overhauling the config system I don't have any concrete suggestions to improve this PR, it fits pretty well with the status quo options/config handling.

alerque:
BTW one thing I'm pretty adept at is rebasing and handling merge conflicts. I'm happy to do that for PRs like this where the conflict came up because of other work I merged too if it would facilitate and not hinder your own workflow(s).

blinry:
> BTW one thing I'm pretty adept at is rebasing and handling merge conflicts. I'm happy to do that for PRs like this where the conflict came up because of other work I merged too if it would facilitate and not hinder your own workflow(s).

Did that this time – also to practice it a bit! :) But at least for my PRs, I'd be happy to see you rebase them on your changes, like you mentioned! Thanks for offering that.

(Also, curious: Do you use any specific "supporting tools" for rebasing?)

zormit:
> I'm happy to do that for PRs like this where the conflict came up because of other work I merged too if it would facilitate and not hinder your own workflow(s).

For mine, I'd be happy to do it myself (it's a task that I enjoy) - but if you would "accidentally" do it, it's no big deal - i.e. it's not _that_ important to me that I would expect from you to keep the fine grained preferences in mind :)

alerque:
> (Also, curious: Do you use any specific "supporting tools" for rebasing?)

Yes. I mix `git revise` and `mergiraf` (and sometimes `git absorb`) layered on top of more traditional rebasing helpers like `git rerere`. Besides `mergiraf` handling a *lot* of possible merge conflicts automatically in languages it knows by doing structural AST based comparisons, `git revise` allows me to move around and apply resolved conflict hunks outside the current step in the rebase queue. Rebasing steps you through history changing the working tree as you go, `git revise` allows me to also apply changes directly to commits outside of that tree without affecting the working copy or position in the rebase queue.

Extensive style changes that have deterministic tooling to replicate (like the import ordering change we just did), I iterate though the rebase *backwards* (newest first instead of oldest first) and apply the deterministic changes up through the commit history until I reach the original branch root, and then splice in whatever changes are left that are not relevant to the branch into a new temporary commit. Then rebasing the whole branch on the new HEAD and dropping that extra commit usually results in no merge conflicts at all.

Jujutsu handles some of those things similarly (and possibly with more grace), but I've been so ingrained in my own workflow and tools I haven't been able to make the switch.

PR #525, opened 2026-04-24
Setup rustfmt rules and apply

alerque:
See discussion here:

https://teamtype.zulipchat.com/#narrow/channel/562985-daemon/topic/Ordering.20of.20imports

**Self-review checklist**

### DIFF
--- a/daemon/.rustfmt.toml
+++ b/daemon/.rustfmt.toml
@@ -6,6 +6,7 @@ edition = "2024"
 style_edition = "2024"
 unstable_features = true
 
-# https://teamtype.zulipchat.com/#narrow/channel/562985-daemon/topic/Ordering.20of.20imports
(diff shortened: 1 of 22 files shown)


blinry:
At first, I hesitated whether activating a nightly rustfmt option was the right call. But in combination with not using the nightly toolchain in the CI, thus *not enforcing* the nightly options, I think I'm fine with it.

The effect is that we have a nicer import ordering right now, and, as you said, @alerque, we can occasionally run `cargo +nightly fmt` ourselves to prettify things.

Contributors will get warnings like "Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel", which might surprise them. I'll add a short explanation to `.rustfmt.toml`.

We should probably enforce (stable) formatting in the `integration-tests` crate as well. But let's do that in another PR?

blinry (approved):

PR #527, opened 2026-04-25
Fixup docs generation (local and doc.rs) and include README

alerque:
The docs.rs documentation is both broken and disabled. This fixes both problems. This is useful even for things like LSP enabled editors to be able to link to documentation while working on something using the API surface. In order to also be a useful landing for end users I also included the readme (which renders pretty nicely) and because the readme is used outside of GitHub (already on crates.io for example, not docs.rc) I also added a link to the more end user facing documentation page more prominently at the top.

**Self-review checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ The [releases on GitHub](https://github.com/teamtype/teamtype/releases/latest) c
 <details>
 <summary>Arch Linux</summary>
 
-```
+```bash
 sudo pacman -S teamtype
 ```
 </details>
@@ -65,7 +65,7 @@ sudo pacman -S teamtype
 <details>
 <summary>Homebrew</summary>
 
-```
+```bash
 brew install teamtype
 ```
 </details>
@@ -75,7 +75,7 @@ brew install teamtype
 
 To put `teamtype` in your PATH temporarily, run:
 
-```
+```bash
 nix shell nixpkgs#teamtype
 ```
 
@@ -106,7 +106,7 @@ cargo binstall teamtype
 
 In the directory you want to share:
 
-```
+```console
 $ teamtype share
 
     To connect to you, another person can run:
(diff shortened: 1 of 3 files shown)


zormit (approved):
I think the reasoning why we disabled docs.rs was "we're not a library", so we wouldn't want to maintain a public facing API "contract". I'm not sure how the policy is on "end user software" on docs.rs, it seems you're saying, if you put up a README there, it's okay?

I was surprised to see that we have [something](https://docs.rs/crate/teamtype/latest) there, even though we _tried_ to disable it in the Cargo.toml... So in that case I would also say we should fix it rather than let a broken docs build hang there...

alerque:
> I think the reasoning why we disabled docs.rs was "we're not a library", so we wouldn't want to maintain a public facing API "contract".

This is a point I've been meaning to bring up separately. Personally I think we'd actually benefit from cleaning up the split between our own CLI tool and the library that backs it. Among other things it would enable integrations to build on Teamtype while being self encapsulated without having to run external CLI tooling at all.

> I'm not sure how the policy is on "end user software" on docs.rs, it seems you're saying, if you put up a README there, it's okay?

There is no policy against it that I know of. And when the main user interface is *not* the API surface area it is pretty normal to lead off with whatever *is* most relevant to get people started.

> I was surprised to see that we have [something](https://docs.rs/crate/teamtype/latest) there, even though we _tried_ to disable it in the Cargo.toml... So in that case I would also say we should fix it rather than let a broken docs build hang there...

There are two different interfaces going on, you have one disabled but you can't disable both. Even if you don't have API docs the docs site will at least have a project file browser and some release info. That interface is usually secondary to the API docs but it actually exists as a way to explore the sources of published crates even if they don't publish their sources elsewhere (and to confirm that what they publish to crates.io is the same as their GitHub sources or whatever. But yes, with the API docs disabled they do kind of make it look like a broken window.

Issue #82, opened 2024-07-31
Implement a proper daemon shutdown


blinry:
This would be used to:

- persist the CRDT state one final time
- shut down all child tasks

It would be triggered by Ctrl-C, for example.

blinry:
The [Tokio documentation on "Graceful Shutdown"](https://tokio.rs/tokio/topics/shutdown) gives hints on how to implement this.

alerque:
This also needs to be triggered any time unhandled error bubble up from child functions.

alerque:
Some of the process handling is fixed in #533, but it just traps the right processes and sends shutdown signals to all the threads. This gives a valid place to actually *do* the clean shutdown actions like close out the CRDT state, but it doesn't actually do them it just gives a framework where those things could actually go.

alerque:
c.f. https://github.com/p2panda/p2panda/pull/1139/changes

alerque:
As a side note to review when working an this. Once we handle passing cancellation tokens to each thread we're spawning and each has its own cleanup work to do, we should be able to remove the default panic hook. Currently it's eliding actual panic messages because CTRL-C is triggering some of the threads to panic even when the user requested a shutdown. Basically we're just hiding the panic info as irrelevant. When threads get a proper shutdown message it should become relevant again. Either there won't be panics are if there are they would be 'real' unexpected behavior.

PR #528, opened 2026-04-25
Sanitize dev-facing error and import naming code

alerque:
This comes out of some paper-cuts I got while sorting through the existing code. This shouldn't change *much*. There will be some slight differences in backtraces because there is a difference between directly panicing in a deep function and raising an error that causes a panic higher up, but the additional context information from each layer that errors do successfully propagate should be a good thing. Other than that difference this should all just be develope facing code readability improvements. There are certainly some subjective calls here becuase there is more than one way to skin a cat. I started with points that caught my attention like:

- 'it makes no sense to `use tokio::time::Duration` and thereafter refer to `Duration` but in the same file refer to `tokio::time::Instant`' and
- 'aliasing documented types from other crates when they don't even conflict with anything just makes it hard to reference the appropriate docs'

and expanded the concept to try to make the import/call verbosity balance as consistent as I could without changing any public interfaces or getting into the weeds where actual ambiguity would result.

These commits are best reviewed one by one, combining them makes it hard to see *why* a change is made.

**Self-review checklist**

### DIFF
--- a/daemon/src/document.rs
+++ b/daemon/src/document.rs
@@ -1,11 +1,13 @@
 // SPDX-FileCopyrightText: 2024 blinry <mail@blinry.org>
 // SPDX-FileCopyrightText: 2024 zormit <nt4u@kpvn.de>
+// SPDX-FileCopyrightText: 2026 Caleb Maclennan <caleb@alerque.com>
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
 use std::collections::HashMap;
 
-use anyhow::{Result, bail};
+use anyhow::Result;
+use anyhow::{anyhow, bail};
 use automerge::{
     AutoCommit, ChangeHash, ObjType, Patch, PatchLog, ReadDoc,
     sync::{Message as AutomergeSyncMessage, State as SyncState, SyncDoc},
@@ -378,7 +380,7 @@ impl Document {
         if let Ok(Some((automerge::Value::Object(ObjType::Map), file_map))) = file_map {
             Ok(file_map)
         } else {
-            Err(anyhow::anyhow!(
+            Err(anyhow!(
                 "Automerge document doesn't have a {name} Map object"
             ))
         }
@@ -393,7 +395,7 @@ impl Document {
         if let Some((automerge::Value::Object(ObjType::Text), text_obj)) = text_obj {
             Ok(text_obj)
         } else {
-            Err(anyhow::anyhow!(
+            Err(anyhow!(
                 "Automerge document doesn't have a {file_path} Text object, so I can't provide it"
             ))
         }
@@ -414,7 +416,7 @@ impl Document {
         if let Some((automerge::Value::Object(ObjType::Text), text_obj)) = text_obj {
             Ok(text_obj)
         } else {
-            Err(anyhow::anyhow!(
+            Err(anyhow!(
                 "Automerge document doesn't have a {file_path} Text object, so I can't provide it"
             ))
         }
--- a/daemon/src/document.rs
+++ b/daemon/src/document.rs
@@ -10,7 +10,7 @@ use anyhow::Result;
 use anyhow::{anyhow, bail};
 use automerge::{
     AutoCommit, ChangeHash, ObjType, Patch, PatchLog, ReadDoc,
-    sync::{Message as AutomergeSyncMessage, State as SyncState, SyncDoc},
+    sync::{Message, State, SyncDoc},
     transaction::Transactable,
 };
 use dissimilar::Chunk;
@@ -86,7 +86,7 @@ impl Document {
     pub fn receive_sync_message_log_patches(
         &mut self,
         message: AutomergeSyncMessage,
-        peer_state: &mut SyncState,
+        peer_state: &mut AutomergeSyncState,
     ) -> Vec<Patch> {
         let mut patch_log = PatchLog::active();
         self.doc
@@ -136,7 +136,7 @@ impl Document {
     #[must_use]
     pub fn generate_sync_message(
         &mut self,
-        peer_state: &mut SyncState,
+        peer_state: &mut AutomergeSyncState,
     ) -> Option<AutomergeSyncMessage> {
         self.doc.sync().generate_sync_message(peer_state)
     }
@@ -562,7 +562,7 @@ mod tests {
         #[test]
         fn test_generate_sync_message() {
             let mut document = Document::default();
-            let mut state = SyncState::new();
+            let mut state = AutomergeSyncState::new();
             assert!(document.generate_sync_message(&mut state).is_some());
             // Stops for now and waits for a response
             assert!(document.generate_sync_message(&mut state).is_none());
--- a/daemon/src/document.rs
+++ b/daemon/src/document.rs
@@ -7,8 +7,8 @@
 use std::borrow::Cow;
 use std::collections::HashMap;
 
-use anyhow::Result;
-use anyhow::{anyhow, bail};
+use anyhow::bail;
+use anyhow::{Context, Result};
 use automerge::{
     AutoCommit, ChangeHash, ObjId, ObjType, Patch, PatchLog, ReadDoc, ScalarValue, Value,
     sync::{Message as AutomergeSyncMessage, State as AutomergeSyncState, SyncDoc},
@@ -154,7 +154,7 @@ impl Document {
         let mut offset = 0isize;
 
         let Some(Content::String(text)) = self.current_file_content(file_path) else {
-            panic!("Should have initialized text before applying delta to it");
+            bail!("File {file_path} not initialized before applying delta");
         };
         // Clone so that `self` is no longer borrowed.
         let text = text.clone();
@@ -179,7 +179,7 @@ impl Document {
 
         // Update cached content.
         let Some(Content::String(old_text)) = self.files.get(file_path) else {
-            panic!("Did not find text content in files cache at file_path");
+            bail!("Did not find text content in files cache at file_path");
         };
         let new_text = delta.apply_to(old_text)?;
         self.files
@@ -361,9 +361,7 @@ impl Document {
         if let Ok(Some((Value::Object(ObjType::Map), file_map))) = file_map {
             Ok(file_map)
         } else {
-            Err(anyhow!(
-                "Automerge document doesn't have a {name} Map object"
-            ))
+            bail!("Automerge document doesn't have a {name} Map object")
         }
     }
 
@@ -372,13 +370,13 @@ impl Document {
         let text_obj = self
             .doc
             .get(file_map, file_path)
-            .unwrap_or_else(|_| panic!("Failed to get {file_path} key from Automerge document"));
+            .with_context(|| format!("Failed to get {file_path} key from Automerge document"))?;
         if let Some((Value::Object(ObjType::Text), text_obj)) = text_obj {
             Ok(text_obj)
         } else {
-            Err(anyhow!(
+            bail!(
                 "Automerge document doesn't have a {file_path} Text object, so I can't provide it"
-            ))
+            )
         }
     }
 
@@ -389,13 +387,13 @@ impl Document {
         let text_obj = self
             .doc
             .get_at(file_map, file_path, heads)
-            .unwrap_or_else(|_| panic!("Failed to get {file_path} key from Automerge document"));
+            .with_context(|| format!("Failed to get {file_path} key from Automerge document"))?;
         if let Some((Value::Object(ObjType::Text), text_obj)) = text_obj {
             Ok(text_obj)
         } else {
-            Err(anyhow!(
+            bail!(
                 "Automerge document doesn't have a {file_path} Text object, so I can't provide it"
-            ))
+            )
         }
     }
 
--- a/daemon/src/document.rs
+++ b/daemon/src/document.rs
@@ -4,13 +4,14 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
+use std::borrow::Cow;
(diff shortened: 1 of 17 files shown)


daemon/integration-tests/src/socket.rs
```diff
@@ -77,7 +77,7 @@ impl MockSocket {
             .expect("Could not send message");
     }
 
-    pub async fn recv(&mut self) -> JSONValue {
```
blinry, line 80: I kind of *wanted* to highlight in the function signature that this returns a JSON value, to make it easier to understand what this function does. So a compromise could be to not set up an alias, but to use `serde_json::Value` directly?

daemon/src/peer.rs
```diff
@@ -374,7 +374,7 @@ impl EndpointActor {
         }
     }
 
-    fn handle_incoming_connection(&self, conn: iroh::endpoint::Connection) {
```
blinry, line 377: To be consistent with the suggested `serde_json::Value` above, I would prefer it to stay with the `iroh::endpoint::Connection`, which indeed seems easier to look up in the docs.

daemon/src/peer/sync.rs
```diff
@@ -56,7 +56,7 @@ impl SyncActor {
         let (reponse_tx, response_rx) = oneshot::channel();
         match message {
             PeerMessage::Sync(message_buf) => {
-                let message = AutomergeSyncMessage::decode(&message_buf)?;
```
blinry, line 59: Especially here, it's important to me to highlight that this is not one of "our" message types, but that it's an Automerge message type. So let's go with `automerge::sync::Message`.

daemon/src/peer/sync.rs
```diff
@@ -35,7 +35,7 @@ pub trait Connection<T>: Send + Sync {
 /// Exchanges [`PeerMessage`]s with the connection, and communicates with the document on the other side.
 /// Maintains the sync state.
 pub struct SyncActor {
-    peer_state: SyncState,
```
blinry, line 38: `automerge::sync::State`

daemon/src/document.rs
```diff
@@ -10,7 +10,7 @@ use anyhow::Result;
 use anyhow::{anyhow, bail};
 use automerge::{
     AutoCommit, ChangeHash, ObjType, Patch, PatchLog, ReadDoc,
-    sync::{Message as AutomergeSyncMessage, State as SyncState, SyncDoc},
```
blinry, line 11: Same as in `sync.rs`.

daemon/src/ot.rs
```diff
@@ -500,7 +501,7 @@ mod tests {
     }
 
     mod operational_transform_internals {
-        use operational_transform::Operation as OTOperation;
```
blinry, line 503: This seems like an "exception to the rule", where I'd prefer a shorter name than the full path. So `Operation` is maybe fine.

daemon/src/document.rs
```diff
@@ -4,12 +4,13 @@
...
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
+use std::borrow::Cow;
 use std::collections::HashMap;
 
 use anyhow::Result;
 use anyhow::{anyhow, bail};
 use automerge::{
-    AutoCommit, ChangeHash, ObjType, Patch, PatchLog, ReadDoc,
+    AutoCommit, ChangeHash, ObjId, ObjType, Patch, PatchLog, ReadDoc, ScalarValue, Value,
```
blinry, line 13: For ObjId, ScalarValue and Value, I'd like to keep the automerge prefix, to highlight that they're from that package. We have so many of our own IDs and types…

daemon/src/peer.rs
```diff
@@ -119,10 +121,10 @@ impl ConnectionManager {
     async fn build_endpoint(
         app_config: &AppConfig,
         base_dir: &Path,
-    ) -> Result<(iroh::Endpoint, SecretKey)> {
```
blinry, line 122: Slight preference to keep it `iroh::Endpoint` in this function.

daemon/src/types.rs
```diff
@@ -4,11 +4,14 @@
...
-use std::fmt;
+use std::borrow::Cow;
+use std::{fmt, vec};
 
 use anyhow::{Context, Error, Result};
 use anyhow::{anyhow, bail};
-use automerge::{ConcreteTextValue, Patch, PatchAction, TextEncoding};
+use automerge::{
+    ConcreteTextValue, ObjId, ObjType, Patch, PatchAction, Prop, ScalarValue, TextEncoding, Value,
+};
```
blinry, line 14: Here I'm torn. Not sure what the right mixture of consistency and highlighting the foreign types in the code below would be.

daemon/src/peer/sync.rs
```diff
@@ -116,7 +117,7 @@ impl SyncActor {
                     match doc_ping {
                         Ok(()) => { self.generate_sync_message().await?; }
                         Err(RecvError::Closed) => {
-                            panic!("Doc changed channel has been closed");
+                            bail!("Doc changed channel has been closed");
```
blinry: I think this would introduce a changed behavior: The EndpointActor calls this `run()` without paying attention to the returned result. So while before we definitely would end the daemon here (because of an inrecoverable error), now the daemon would keep running, just without the SyncActor, I think.

It would be much nicer if all Results were handled properly, of course, but that would require a bit more work. So my preference would be to keep the panic!s there for now, while replacing returned Errs with bail! seems good! :)

blinry (changes requested):
I like some of these changes, while for others, I think it makes it less visible what type the code is talking about. Some panic! behaviour indeed changed. See comments below!

I went through this PR with @zormit, but we didn't always agree. He'll add another comment. :)

zormit:
> This comes out of some paper-cuts I got while sorting through the existing code.

I'd be curious to hear what kind of paper-cuts / misunderstandings the current code produced, given that our goal was to make it easily readable :) As in: Where did it go wrong? Maybe a different _way_ of reading the code?

The topic of import consistency feels somehow "dangerously close to bikeshedding" (without desiring to kill the discussion, just being conscious about how much effort to put into it if we don't find agreement :D). I remember that we tried to be very conscious about each of these aliases, while I see that we don't have "coding guidelines" that draw the boundaries, so it's maybe an "almost one person opinionated code" symptom that we see here.

I would be very willing to be more consistent while I also see the "noise" that this produces. And letting go of these previous conscious decisions 😅 

Towards my initial question: If the code says `Value` instead of `JSONValue`, how do you know what kind of "Value" it refers to? Maybe we don't use enough LSP tooling here...

zormit:
btw, the last commit (replacing `panic!` macros) seems related to #82 - currently panicking is a "hack" to ensure everything is properly shut down, when we find a state that doesn't seem recoverable (at least with the existing logic). So if #82 is properly implemented then we can more safely bubble up the errors.

zormit:
Kinda related in the same vein #29

alerque:
I was kind of hoping to be able to at least introduce this talking to both of you, perhaps even working through it in a fast cherry pick session. I'm aware there are some wins and losses here and debating some of the less clear ones could easily turn into bike shedding. I don't want to take it that direction.

I did keep getting tripped up by inconsistencies. Several times I only discovered it when I was struggling to grep dependency source code (e.g. trying to figure out what changed between Iroh releases to make #515 function) only to discover I was looking for Teamtypes alias instead of the original. Instead of picking away at them one by one I decided to just find them all and see if there was a consistent way to handle them.

As you can see here, there is not really a way to be consistent, and trying too hard actually looses some other ground. I think I agree with most of @blinry's specific comments above. It's certainly unfortunate that dependencies sometimes use the most ridiculous generic names for things! Thanks to serde's ubiquity half the Rust world has to cope with the absurdity of `Value` being a blob with a blob intended to be displayed as a fragment of JSON.

I'll look through this and either just cherry pick the clear wins or make another stab at refactoring with some cohesion.

zormit:
Makes sense, would have been good to talk in person, blinry and I had a vivid discussion yesterday :-P Let us know if you want us to have another look.

btw, thinking about it again, I also can see aliasing as a most minimalistic form of a "Newtype", in the sense that you get a little bit more abstracted from a dependency on purpose. But yeah, this makes grepping the code harder, especially if it's not done consistently in one place for a specific dependency.

daemon/integration-tests/src/socket.rs
alerque: That compromise bothers me more than the alias since it hides the import. I'd rather be able to preview at the top of the file any/all external things that will be used (this will come up in relation to your other review comments as well), so for now maybe I'll just revert this.

Long term I think a better approach is never publicly exposing foreign types at all, which means wrapping this is a lightweight type alias. Since that change is a bit more involved than this PR (which I'd like to have as a baseline for other changes sooner rather than later), I'll save that for a future PR.

daemon/src/peer.rs
alerque: If you feel strongly about this please just say so and I'll revert ...

Baring that can you re-review? Per above, I didn't take the suggestion about the Serde type and just reverted to aliasing `JSONValue`, saving what I think will be a more agreeable fix for a later PR. In

Meanwhile, I saw this scenario as a different problem; I wasn't trying to remove unnecessary aliases I was actually *adding* a necessary alias to deconflate the 3 assorted 'Connection' types. On further review I think a more fundamental hiccup is that for some reason WE have a (thankfully private) trait here called `IrohConnection` but it doesn't come from Iroh, it is our trait. It isn't even a wrapper for Iroh's `endpoint::Connection`.

I took another stab at this and kept my alias for `iroh::endpoint::Connection as IrohEndpointConnection`, but renamed our internal trait from `IrohConnection` to `PeerMessageStream` (taking care of one possible conflation) while keeping both from being confused with our own `sync::Connection`.

daemon/src/peer/sync.rs
alerque: Again as with serde I'd rather keep the import and alias than use full import paths in the code, so I just reverted this. With the caveat that I expanded the aliases you had to consistently point out the module they come from. Now the change is just this:

```diff
-use automerge::sync::{Message as AutomergeSyncMessage, State as SyncState};
+use automerge::sync::{Message as AutomergeSyncMessage, State as AutomergeSyncState};
```

daemon/src/peer/sync.rs
alerque: See [previous](https://github.com/teamtype/teamtype/pull/528#discussion_r3168204502).

daemon/src/document.rs
alerque: See [previous](https://github.com/teamtype/teamtype/pull/528#discussion_r3168204502).

daemon/src/ot.rs
alerque: Just like serde, I think this is a flaw in the upstream crate design. Overly generic public types do not play nice downstream ;-)

daemon/src/document.rs
alerque: None of those three are made public here. On the other hand the absurdly generic `Patch` type which you don't have a problem with *is* exposed (via `pub fn receive_sync_message_log_patches`). My suggestion is to just bite the bullet here and assume based on reading the import header that most things in here are going to be foreign, and let that motivate us to properly encapsulate foreign types and hence cut down on this mess. We might for example avoid the same mistake as Serde and clarify what exactly our public `Document` type is here.

daemon/src/peer/sync.rs
alerque: Reverted, you are correct to point out this change is not like the rest and should at least be in a separate commit with different testing. I'll put this in a separate PR with perhaps an attempt to also handle the error properly where it propagates.

daemon/src/types.rs
alerque: Ditto [my comment on document.rs](https://github.com/teamtype/teamtype/pull/528#discussion_r3168317095) re same types.

daemon/src/peer.rs
alerque: Would the fact that this would be the **only** use project wide of a type using a fully qualified import path outside of a lightweight wrapper be enough to overcome your slight preference? We have some function and macro calls that are fully qualified, but those are up top next to the `use` statements, so you can see all types that are used a glance.

daemon/src/peer.rs
blinry: Works for me! I also like the change to `PeerMessageStream`.

daemon/src/document.rs
blinry: I mean, these imports *do* make some of the lines in this file nicer to read, that's for sure. And document.rs now *does* deal mostly with Automerge things anyway right now (plus the "files cache" that's not-yet-split out after #480). It's okay for me to change the imports like this.

daemon/src/peer.rs
blinry: Okay, let's try this style of mostly avoiding fully qualified names :)

daemon/src/types.rs
blinry: I think me using a qualified name in inline code might be a sign that the code doesn't really belong in that file.

In the case of the Automerge types, for example, I might've left the `automerge::` there as a reminder to myself that it might be good to pull that code out, and put it in another function, so that the "Automerge-handling code" is split off cleaner from the rest.

By removing the prefix, it becomes harder to see in the code, but it's still obvious in the import header. I can see why you like being able to see all "dependencies" of a module at the top. Let's try it.

blinry:
(Still wanna give the diff another full read, but ran out of time for today.)

blinry (approved):
Looking fine to me! Thanks for your work on this, @alerque. The import changes probably *do* make the code nicer to read.

PR #529, opened 2026-04-28
Restyle Markdown and copy-edit some docs

alerque:
Not finished yet, but as I'm working through docs I'm also trying to clean up bits of verbiage.

**Self-review checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,12 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 For complete end-user documentation see the [Teamtype Documentation][docs] site.
 
 > [!NOTE]
-> This project used to be called "Ethersync" until October 2025. See [this PR](https://github.com/teamtype/teamtype/pull/436) for our motivation.
+> This project used to be called "Ethersync" until October 2025.
+> See [this PR](https://github.com/teamtype/teamtype/pull/436) for our motivation.
 
-Teamtype enables real-time collaborative editing of local text files. You can use it for pair programming or note-taking, for example. It's the missing real-time complement to Git!
+Teamtype enables real-time collaborative editing of local text files.
+You can use it for pair programming or note-taking, for example.
+It's the missing real-time complement to Git!
 
 ![Demo video for how to make a connection and of collaborating in Neovim](https://files.blinry.org/teamtype-share-join-demo.gif)
 
@@ -31,13 +34,17 @@ We also maintain a list of [other collaborative text-editing software](https://t
 
 ## What Teamtype is not
 
-We are not a company, and don't sell anything. We don't require you to create an account. We don't have access to your data, and don't use it to train AI algorithms. We don't serve you ads, and don't track you.
+We are not a company, and don't sell anything.
+We don't require you to create an account.
+We don't have access to your data, and don't use it to train AI algorithms.
+We don't serve you ads, and don't track you.
 
 We're just a bunch of people building something we want to see in the world.
 
 ## 🚦 Project status
 
-We're currently maintaining Teamtype in our free time. We're often using it for pair programming ourselves, but there's still some [bugs](https://github.com/teamtype/teamtype/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen+%28label%3Abug+OR+type%3ABug%29) to be aware of.
+We're currently maintaining Teamtype in our free time.
+We're often using it for pair programming ourselves, but there's still some [bugs](https://github.com/teamtype/teamtype/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen+%28label%3Abug+OR+type%3ABug%29) to be aware of.
 
 ## 📥 Installation
 
@@ -48,7 +55,8 @@ Teamtype works on Linux, macOS, Android, and on the Windows Subsystem for Linux.
 <details>
 <summary>Binary releases</summary>
 
-The [releases on GitHub](https://github.com/teamtype/teamtype/releases/latest) come with precompiled static binaries. Download one and put it somewhere in your shell's [`PATH`](https://en.wikipedia.org/wiki/PATH_(variable)):
+The [releases on GitHub](https://github.com/teamtype/teamtype/releases/latest) come with precompiled static binaries.
+Download one and put it somewhere in your shell's [`PATH`](https://en.wikipedia.org/wiki/PATH_(variable)):
 
 - `x86_64-unknown-linux-musl` for Linux
 - `universal-apple-darwin` for macOS
@@ -128,7 +136,8 @@ Storing peer's address in .teamtype/config.
 Connected to peer: 5e6b787fff79074735eb9b56939269100de1e37bc7f7a4d29c277cc24f7ee53d
 ```
 
-The directories are now connected, and changes will be synced instantly. You can open text files (using editors with a Teamtype plugin), and start collaborating in real time! :sparkles:
+The directories are now connected, and changes will be synced instantly.
+You can open text files (using editors with a Teamtype plugin), and start collaborating in real time! :sparkles:
 
 ## 🎓 Learn more
 
@@ -160,7 +169,11 @@ Bridges:
 
 ## 🔨 Contributing
 
-We'd love to receive your patches and other contributions! Small patches are very welcome as PRs. Before starting to implement a new big feature, please briefly [check in with us](#contact) so we can discuss how it fits in with our ideas for the project. We have a [CONTRIBUTING.md](CONTRIBUTING.md) file that gives more guidance. Please take note of our [**strict no-LLM policy**](CONTRIBUTING.md).
+We'd love to receive your patches and other contributions!
+Small patches are very welcome as PRs.
+Before starting to implement a new big feature, please briefly [check in with us](#contact) so we can discuss how it fits in with our ideas for the project.
+We have a [CONTRIBUTING.md](CONTRIBUTING.md) file that gives more guidance.
+Please take note of our [**strict no-LLM policy**](CONTRIBUTING.md).
 
 If you're interested in building new editor plugins, read the [editor plugin development guide](https://teamtype.github.io/teamtype/editor-plugin-dev-guide).
 For more information about Teamtype's design, refer to the list of [decision records](docs/decisions/).
@@ -169,7 +182,8 @@ If you find bugs, please [open an issue](https://github.com/teamtype/teamtype/is
 
 ## ☎️ Contact
 
-Feel free to [join us on Zulip](https://teamtype.zulipchat.com) to ask us anything! Other good channels:
+Feel free to [join us on Zulip](https://teamtype.zulipchat.com) to ask us anything!
+Other good channels:
 
 - Mastodon: [@teamtype@fosstodon.org](https://fosstodon.org/@teamtype)
 - Email: <span>t<span title="ihate@spam.com&lt;/span&gt;">e</span>amtype</span>@zormit<i title="&lt;/i&gt;mailto:">.</i>de
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 <!--
 SPDX-FileCopyrightText: 2024 blinry <mail@blinry.org>
 SPDX-FileCopyrightText: 2024 zormit <nt4u@kpvn.de>
+SPDX-FileCopyrightText: 2026 Caleb Maclennan <caleb@alerque.com>
 
 SPDX-License-Identifier: CC-BY-SA-4.0
 -->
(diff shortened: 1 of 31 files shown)


zormit:
Haha so you're also on the "every sentence should be one line"-camp? :) So far I was the only one and I didn't care too much to convince @blinry (who is, for Markdown, using "one paragraph per line") to follow that style, but also right now it's pretty inconsistent, as each of us has often just followed their style.

Curious how you would feel about changing this, @blinry.

blinry:
My main reason for "my" preference is that the documents looks nicer when read as plain text, without Markdown rendering, which I sometimes do when I interact with projects on the command line.

For example, when I want to look up an ADR, I'd just open the file. And it'd be not as pretty (to me) if every sentence was on its own line there.

It would be readable though.

You're both trying to optimize for "cleaner diffs", I presume?

alerque:
> My main reason for "my" preference is that the documents looks nicer when read as plain text, without Markdown rendering, which I sometimes do when I interact with projects on the command line.

I actually *agree* that if you are reading the plain text rendition it is nicer to read an actual paragraph. I just don't think that is the main use case for the README or mdbook docs or anything. Most people will experience those rendered, and the people that are looking at the sources are more likely to be potential editors, so I was optimizing for editing.

I'm not against some optimizations for reading though. For example, that's one reason to use reference URLs to. The constant intrusion of full inline URLs and such make it unpleasant to read, so many of the changes here –like using references for links– instead of full inline URLs should improve plain text readability.

> You're both trying to optimize for "cleaner diffs", I presume?

Sort of. I was optimizing for easier *editing*. This includes the fact that `diff` tooling is usually easier to work with in linewise modes, but it also has other effects. If I do spot a wording problem with a Markdown file it is often when displayed on GitHub or Crates.io or docs.rs or wherever. Correlating the paragraph-wise rending on those sites to paragraph-wise wrapping in my editor is not trivial. On the other hand at least for my brain it is pretty easy to mentally clock "third sentence of paragraph starting <blah>" and spot that in the Markdown source faster than I could even invoke a search. Copy-editing is easier too and I think it's easier to *edit* prose of the paragraph wrapping isn't jumping around all the time and the only thing changing is length of the line I'm actually working on.

alerque:
> but also right now it's pretty inconsistent, as each of us has often just followed their style.

I kind of figured that was how we got here. I saw both styles about 50/50 through the whole project and figured you had both been just winging it. I decided to not even ask if you'd talked about it first because I knew I was just going to try to convince whichever side was the hold out to come around to sentence-wise wrapping anyway, so figured I'd just normalize that direction first and see how it flew ;-) Also [I have tooling](https://github.com/sile-typesetter/casile/blob/master/pandoc-filters/sentence_lines.lua) that mostly just does it.

zormit:
> You're both trying to optimize for "cleaner diffs", I presume?

Yes, plus navigating it in Neovim: When I have to bring my cursor to a specific position, having to switch to `gj` and `gk` or `(` and `)` instead of `j` and `k` somehow feels like it's slowing me down or that it's more tedious to get the cursor to where I want it? Maybe that's a skill problem on my side tho 😬

zormit:
> For example, when I want to look up an ADR, I'd just open the file. And it'd be not as pretty (to me) if every sentence was on its own line there.

Should we publish them somewhere in rendered form btw?

alerque:
> Maybe that's a skill problem on my side tho 😬

Of course it is :stuck_out_tongue_winking_eye:, there is [vim-pencil](https://github.com/preservim/vim-pencil) and aliases:

```vimscript
nnoremap j gj
nnoremap k gk
```

But I would still argue those adoptions are best suited for *writing* long form prose, not *editing* more technical writing. For book length material like fiction that isn't going to be heavily edited over time I'm still down with paragraph wrapping (and the tools above).

> Should we publish them somewhere in rendered form btw?

My first instinct would be to just put them in the mdbook docs nested a bit out of the way so they don't clutter the TOC for people that are just there for the UX.

blinry:
> Should we publish them somewhere in rendered form btw?

This might already be good enough for me: https://github.com/teamtype/teamtype/tree/main/docs/decisions

> Also I have tooling that mostly just does it.

Glad to hear that! :D

---

I just learned that there's a term for this: "semantic line breaks", and there's a spec, of course: https://sembr.org

Even though it's not my preference, I'm open to trying it, if that's important to both of you! We can unify in this direction and see what it feels like.

(Would you also use this style in code comments? Again, I think I'd prefer them to "look nice" when viewed with a text editor :S Maybe we can compromise and not immediately semantically-break those, as well?)

alerque:
> (Would you also use this style in code comments? Again, I think I'd prefer them to "look nice" when viewed with a text editor :S Maybe we can compromise and not immediately semantically-break those, as well?)

At least in Rust, there are two categories of code comments.

* For developer facing comments on the actual code (`//`) I'd actually prefer *not* to default to this, and just wrap them to roughly match the surrounding line length. The primary use case is just leaving hints for other developers, they are rarely longer than a couple of lines, and rarely see ongoing edits, and shouldn't be a source of merge conflicts. They usually only change if the code changes too.
* For Rust's inline documentation options (`///` and `//!`) I'd say the primary use case is reading them in docs.rs, man pages, `--help`, or another documentation viewers and they are technical writing that likely will see ongoing detailed editing. I'd say the same logic applies and leveraging semantic breaks, references, etc. make some sense.

alerque:
I'm marking this as ready to review because I'm starting to hit the same typos and other bits over and over as I work on other branches so it would be nice to get at least some portion of this merged.

If there is anything in here (the semantic breaks, the editorial style, the typographic bits, etc.) where the whole class of changes bothers you guys let me know and I can split it out for separate discussion or discard or whatever. I've tried to keep the commits grouped by concern — and it will almost certainly make more sense to review it commit by commit than as whole. If we let it grow any longer reviewing is going to get unwieldy indeed.

blinry:
I'll try to get a more detailed review in soon – but the one thing that would be good to pull out, to discuss separately (or discard), is the quotes:

mdbook already seems to convert some of the `"` quotes to the typographic ones. (Not sure if all of them.) Where possible, I'd prefer to solve this "on the tooling side", not on the "input side" (as I'm at least not used to typing them, and might forget to use them in the future, same for other contributors).

(That won't work for the READMEs though, as Github's Markdown renderer doesn't seem to touch them.)

alerque:
Done.

CONTRIBUTING.md
```diff
@@ -34,6 +34,10 @@ We will do this until the number of contributors grow too large and then switch
...
-We do not accept any contributions that contain content generated by LLMs.
-This also includes all written communication on GitHub.
-We will close any pull request, issue, or discussion that does not comply to this policy.
+If you used LLMs to contribute to this project, we expect you to disclose its usage.
+This is important for us order for us to figure out with which level of care we should apply when reading the code and providing feedback.
+Transparency helps us gain trust in the code and in you.
+
+You have to take responsibility for your contribution, and make sure that it is high-quality.
+In particular, you must personally understand and test all output generated by your tool.
+We like [Zulip's AI use policy and guidelines](https://zulip.readthedocs.io/en/latest/contributing/contributing.html#ai-use-policy-and-guidelines) in this regard.
```
blinry: Oops, seems that the LLM policy changed back! :D Was the line-break work based on an older commit, maybe?

I scrolled over the diff, but didn't really compare the content word-by-word. Just noticed the different lengths here. Everything else seemed plausible to me.

book/src/relays.md doesn't get reformatted by this commit yet, we should do that, too.

daemon/src/wormhole.rs
```diff
@@ -35,7 +35,7 @@ pub async fn put_secret_address_into_wormhole(address: &str, magic_wormhole_rela
                 warn!("Failed to share secret address. Did your peer mistype the join code?");
             }
 
-            // Print a new join code in the next iteration of the foor loop, to allow more people
+            // Print a new join code in the next iteration of the for loop, to allow more people
```
blinry, line 38: *foor loop* is kinda cute! :D Thanks for catching all these silly mistakes for us!

README.md
```diff
@@ -44,7 +44,7 @@ We're just a bunch of people building something we want to see in the world.
 ## 🚦 Project status
 
 We're currently maintaining Teamtype in our free time.
-We're often using it for pair programming ourselves, but there's still some [bugs](https://github.com/teamtype/teamtype/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen+%28label%3Abug+OR+type%3ABug%29) to be aware of.
+We're often using it for pair programming ourselves, but there's still some [bugs][bugs] to be aware of.
```
blinry: Hmm, I'm not sure about this link style. From the perspective of a person who opens this file in an editor, I'd argue that it makes it *harder* to consume: If I'm interested in following the link, I have to 1. know that this *is* a link syntax in the first place and 2. jump to the correct link below.

When deleting a sentence, it seems that it's easy to forget one of the links at the bottom, and they might accumulate?

Would be curious about your arguments in this style, and whether you feel strongly about this. The sentences are easier to read in plain text format, right?

README.md
```diff
@@ -1,45 +1,46 @@
...
 SPDX-FileCopyrightText: 2024 zormit <nt4u@kpvn.de>
+SPDX-FileCopyrightText: 2026 Caleb Maclennan <caleb@alerque.com>
 
 SPDX-License-Identifier: CC-BY-SA-4.0
 -->
 
 <img src="https://raw.githubusercontent.com/teamtype/teamtype-design/refs/heads/main/logo/regular/on-light.svg" width="250px">
 
-*Multiplayer mode for your text editor!*
+*Multiplayer mode for <ins>your</ins> text editor!*
```
blinry: Is this supposed to underline the word? Doesn't render in GitHub's rich text diff. I feel emphasizing this word in the slogan makes it a bit less clear – all parts of the slogan have equal emphasis in my head.

README.md
```diff
@@ -1,45 +1,46 @@
...
-> See [this PR][formerly] for our motivation.
+> Until October 2025, this project was known as "Ethersync".
+> The motivation for the rename was explained in [this PR][formerly].
 
-Teamtype enables real-time collaborative editing of local text files.
-You can use it for pair programming or note-taking, for example.
-It's the missing real-time complement to Git!
+Teamtype enables real-time peer-to-peer collaborative editing of files using your own text editor or other tools.
+For example, you can use it for pair programming sessions, brainstorming, or note-taking.
+It's the missing real-time complement to version control systems!
```
blinry: I'd rather keep "Git" here, as the by far most popular incarnation of VCS.

README.md
```diff
@@ -1,45 +1,46 @@
...
 > [!NOTE]
-> This project used to be called "Ethersync" until October 2025.
-> See [this PR][formerly] for our motivation.
+> Until October 2025, this project was known as "Ethersync".
+> The motivation for the rename was explained in [this PR][formerly].
 
-Teamtype enables real-time collaborative editing of local text files.
-You can use it for pair programming or note-taking, for example.
-It's the missing real-time complement to Git!
+Teamtype enables real-time peer-to-peer collaborative editing of files using your own text editor or other tools.
```
blinry: Good, but maybe "<ins>local</ins> text files", to keep highlighting that it really works with files on your hard drive? Or do you see another way to mention that?

README.md
```diff
@@ -1,45 +1,46 @@
...
 
-- 👥 Edit files at the same time, across different text editors
-- 📍 See your peers' cursors and selections
-- 🗃️ Work on entire projects, the way you're used to
-- 🔒 Encrypted peer-to-peer connections, no need for a server
-- ✒️ Local-first: You always have full access, even offline
-- 🧩 [Simple JSON-RPC protocol][dev-guide] for writing new editor plugins
+- 👥 Edit files concurrently by multiple users using different text editors
+- 📍 See your peers' cursors, selections, and edits in real time
+- 🗃️ Work seamlessly across project workspaces, the way you're used to
```
blinry: Not sure this makes it clearer. I wanted to highlight that it works on "entire folders".

Replace the bullet point with "Collaborate on entire directories, not just single files", maybe?

("Seamlessly" feels a bit too markety to me here.)

README.md
```diff
@@ -1,45 +1,46 @@
...
+- 🔒 Rest easy with encrypted peer-to-peer connections, no need for a server
+- ✒️ Maintain full access with a local-first model that retains normal files on disk even if you go offline
```
blinry: "Normal files" confused me at first. Just "that retains all content on disk"?

README.md
```diff
@@ -1,45 +1,46 @@
...
 We also maintain a list of [other collaborative text-editing software][related-projects].
 
-## What Teamtype is not
+## What Teamtype is *not*
 
-We are not a company, and don't sell anything.
-We don't require you to create an account.
-We don't have access to your data, and don't use it to train AI algorithms.
-We don't serve you ads, and don't track you.
+We are not a company and don't have anything to sell.
```
blinry: You avoided contractions below (for a more "serious tone", maybe?). For consistency, also "do not" here?

README.md
```diff
@@ -1,45 +1,46 @@
...
+We do not require you to create an account.
+We do not have access to your data, and don't use it to train AI algorithms.
+We do not serve you ads or track you in any way.
 
-We're just a bunch of people building something we want to see in the world.
+We are just a bunch of people building something we want to use ourselves.
```
blinry: Hmm, that's also true, but I'm also building it for others! <3 What's your criticism with the original version?

blinry (changes requested):
Thanks for this PR! Review in the comments.

CONTRIBUTING.md
alerque: Corrected bad merge conflict resolution and applied sembr to relays.md.

README.md
alerque: Removed, it was supposed to do something but I already confirmed it didn't work, but apparently I `fixup ...`'ed rather than `drop ...`'ed the commit.

README.md
alerque: I put 'Git' back in but hedged a bit:

```diff
-It's the missing real-time complement Git!
+It's the missing real-time complement Git and other snapshot based version control systems!
```

Not marking this as resolved because it's a minor point of pedantic idealism on my part and I'm fine with just reverting entirely if it doesn't sound right.

README.md
alerque: I think I decided to remove the emphasis on "text" since a surprising number of people seem to struggle with "plain text" including things like code and markup languages and all sorts of things. True enough, but easily dismissed as not being relevant. Keeping 'local' makes sense though.

README.md
blinry: It's not quite as catchy, but I'm okay with that. (As a fan of jj, I'm also happy about a wider perspective here.)

- You missed a "to"
- I'd take out the "snapshot based", to make this sentence a bit simpler.

README.md
alerque: Per Jitsi discussion I'll pull this out.

README.md
alerque:
```suggestion
We are just a bunch of people building something we want to use ourselves and share with others.
```

alerque:
@blinry Since the last review from you was "changes requested" I can't close this out, but the rebase I did a few hours ago should be just the changes we tasked about, no more no less. If it looks good to you feel free to rebase & merge ... or approve if it all seems copacetic and I'll take care of it later.

<sub>(And yes I stashed the links rewrite commit in case we ever talk you into changing your mind. :wink:)</sub>

blinry (approved):

alerque:
@blinry I got a chuckle out of the sheer serendipity when [this 2021 comment](https://github.com/sembr/specification/pull/7#issuecomment-839474064) hit my GH notification radar just after seeing this PR closure this morning. I guess I'm not *alone* in the universe.

PR #530, opened 2026-04-28
Enable local discovery

### LINKED ISSUE #410: Allow local connection of daemons
Instead of using the wormhole join codes and some relays, we could allow peers to find themselves with iroh local discovery, via the secret address.

blinry:
Fixes #410.

Still to do:

- [x] Test
- [x] Add to documentation

**Self-review checklist**


blinry:
Here's what I did for testing that local discovery works:

- On main, I launched a share on one laptop, and a join with the option `--iroh-dns-domain example.com` on another. As expected, this prevents the discovery from working – the joining peer can't find the sharing peer.
- Then, on this branch, I did the same. Now the peers successfully connect! `RUST_LOG=debug` shows mDNS messages being exchanged.

blinry:
Future ideas:

- Add an option to prevent any external network activity apart from the local discovery.
- Add an option to disable the local discovery – it "broadcasts" your peer ID into the local network, which is why it's not on by default, I assume. We protect connections with a password, but still.

alerque:
I am *not* able to verify that this is working right. Using your testing setup (a bogus iroh-dns-domain) I am able to connect with this branch, and also not able to connect using a build from the default branch, but that isn't the end of the story. The failure case does not convince me that this mechanism is actually working, because even in the original code the peers still do find each other. They even connect and negotiate an initial connection, they only then fail with a "peer ___ lost" message.

I think this needs testing in an offline network. I can do that too, just not right this second.

book/src/connection-making.md
```diff
@@ -27,6 +27,8 @@ After using a join code, the secret address is stored in your `.teamtype/config`
 You can directly connect across different local networks, even when each of you is behind a router.
 This way of connecting is more "ad hoc" and useful if you want to collaborate over a short period of time (as described in more detail in the [pair programming scenario](pair-programming.md)).
 
+Teamtype also uses a "local discovery" mechanism based on mDNS, which allows peers in the same local network to find each other.
```
alerque, line 66: This might be true in a very technical sense, but even I misunderstood it when first seeing this PR, and I think at the very least the documentation should be updated to reflect what this does and does *not* do. Yes there is some local peer discovery now, but this does not close the linked issue and is not usable offline—at least not the way I read that issue or expected it to work. I would expect a local peer discovery to enable completely local discovery. This is eliminating the need for the Iroh DNS to and some of it's hole-punching functioning functionality, but it isn't enabling users to find and connect to each other in a disconnected notwork. It still needs a magic wormhole relay.

I'm not saying this code can't go in as-is because it does move the bar forward and eliminate one of the necessary orchestration servers you'd need to run for peers to connect in an airgapped or firewalled network ... all I'm saying is that somehow this documentation should be more explicit about what piece in the puzzle is being addressed here so people don't expect it to work with no relay servers yet.

alerque (changes requested):
Needs to document exactly which piece(s) in the 4 relay server equation can be eliminated in the case of a closed network.

book/src/connection-making.md
blinry: Right! Most of that knowledge is only in our heads yet. I'll take it on to write that out a bit better.

alerque:
I just rebased this for you using the new crate workspace layout. It's currently stacked on #564 just because that has a fresh lock file update and the correct lock file locations, but that will be irrelevant as far as this feature is concerned. We can always rebase it using an updated lock file at any time.

alerque:
This got behind the Iroh 1.0 release. I just rebased it to address that, which now involves a separate crate instead of a feature flag on the Iroh crate. Otherwise it's not substantially different as a feature.

I'm also splitting the commit situation up a little and stacking the mermaid stuff in the docs into a separate PR.

PR #533, opened 2026-04-30
Setup trap early enough to actually catch signals

### LINKED ISSUE #404: ethersync share should not crash on ending with ctrl+c
When ending the process in command line with "ctrl+c", I have seen repeatedly a crash message:
```
^C
thread 'notify-rs inotify loop' panicked at src/watcher.rs:39:36:
called `Result::unwrap()` on an `Err` value: SendError { .. }
```

Unless ctrl+c is not the intended way to end the process ?

alerque:
See [this comment on PR 490](https://github.com/teamtype/teamtype/pull/490#issuecomment-4350921007) for an explanation of the problem this addresses.

Closes #404.

**Self-review checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 
 # 0.9.2 (unreleased)
 
+Fix CLI thread handling to correctly listen for SIGINT (<Ctrl>-C) and SIGTERM (typically invoked with `kill`) while the main daemon or client loop is running.
+
 Add support for JSON-RPC IDs in string to be compliant with the [JSON-RPC specification](https://www.jsonrpc.org/specification).
 
 Add support for `.teamtypeignore` file to exclude files and directories from synchronization. The syntax is similar to `.gitignore`, supporting file names, directories, and glob patterns.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 
 # 0.9.2 (unreleased)
 
+Handle POSIX specific process signalling without assuming everybody is running a Unix platform.
+
 Fix CLI thread handling to correctly listen for SIGINT (<Ctrl>-C) and SIGTERM (typically invoked with `kill`) while the main daemon or client loop is running.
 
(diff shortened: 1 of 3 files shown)


alerque:
I'm pretty sure the first two commits here are solid. For non-Unix platforms the termination dummy will just never resolve. I'm less sure about the last commit which I coded kind of blindly coping from search results. I don't actually have a Windows environment setup to test if it works. The Linux end of things still builds and works right, but I have not confirmed the same is true for Windows.

alerque:
The Windows build is likely still blocked on

```rust
use std::os::unix::fs::PermissionsExt;
```

in the sandbox module (see #258). Should I leave the Windows specific commit here as a better starting place for the next step, or take it off and stash it somewhere pending being able to test Windows builds? The *current* state of the code is completely broken and either way the Unix end should work, but we won't know if the Windows bit is right or needs further fixes until it can be tested.

alerque:
I answered my own question and moved the untested Windows specific changes to a separate PR and kept this one to just the testable fixes to the existing code.

alerque:
The GitHub diff viewer isn't doing this justice because it's showing all the whitespace changes and lost track of what was important. To make this I basically just:

1) Took the two huge arms of the runtime mode `match` flow control and copied them into their own utility functions.
2) Used the existing `match` and arms to just detect the mode without running anything at that point.
3) Used Tokio's `select!` macro to run the appropriate function with all the runtime logic in parallel with the thread that is listening for shutdown events. 
4) Make sure of those two things running in parallel whatever one ends first for whatever reason informs the other one to shut down too.

I didn't touch the actual logic of what is happening in the runtime logic, I just moved it from sequential code run in an arm of the match expression and then later listening for signals to a form that was easy to trigger in parallel so the signal listener was running while the other logic was being run.

alerque:
In order to ease review I just split the commit that did the Tokio thread re-arrangement into two commits. The first commit should be 100% equivalent to the current logic, it just copies the code out of the huge match expression so each arm gets its own helper function. This should make it pretty obvious where the signals were being trapped and why they were not working. The second commit then doesn't need to move huge blocks of code around, it just converts the runtime mode match expression into something we can more ergonomically pass to Tokio for it to pick a thread to run, then sets up the Tokio threads and signalling in the right order.

blinry (approved):
Looking good! Thanks for making this easy to review! <3 The `trap_shutdown()` code is very elegant!

I also don't have a way of easily testing this on Windows right now, but I see that you're working on that in #534. :)

PR #534, opened 2026-04-30
Gate Unix specific bits and add Windows specific bits

### LINKED ISSUE #258: Windows support

### LINKED ISSUE #485: tracking issue: Windows Support
This is an issue to discuss what will be necessary in order to support Windows.

- [ ] `teamtype client` needs to work with [`NamedPipes`](https://learn.microsoft.com/en-us/windows/win32/ipc/named-pipes)
- [ ] the daemon (`teamtype share`/`join`) works with `NamedPipes`
- [ ] shutdown mechanism that is not dependent on Unix Signals
- [ ] sandbox / file system module is multi-platform
- [ ] Continuous integration runs on Windows to confirm it build and run the tests

Open Questions:
- How to secure the `NamedPipe` properly?
- Which editor should be used in the Fuzzer on Windows? Can we use some version of Neovim? Or should we go with unit tests only, as a start?
- Overall, how do we want to approach the multi-platform code? I think thinking about this will help us get our interfaces clearer and the code structured better.

alerque:
Stacked on ~~#533, #541, #603, #611, #601,~~ and #560.

Closes #258
Closes #485.

Currently untested. Passes Cargo checks in cross compile and builds in Windows on CI, but untested.

**Self-review checklist**

- [ ] I have manually tested and self-reviewed my changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -849,6 +849,33 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "61803da095bee82a81bb1a452ecc25d3b2f1416d1897eb86430c6159ef717c17"
 
 [[package]]
+name = "crossterm"
+version = "0.29.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d8b9f2e4c67f833b660cdb0a3523065869fb35570177239812ed4c905aeff87b"
+dependencies = [
+ "bitflags",
+ "crossterm_winapi",
+ "derive_more",
+ "document-features",
+ "mio",
+ "parking_lot",
+ "rustix",
+ "signal-hook",
+ "signal-hook-mio",
+ "winapi",
+]
+
+[[package]]
+name = "crossterm_winapi"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "acdd7c62a3665c7f6830a51635d9ac9b23ed385797f70a83bb8bafe9c572ab2b"
+dependencies = [
+ "winapi",
+]
+
+[[package]]
 name = "crunchy"
 version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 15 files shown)


alerque:
I can confirm this does *not* build, at least not cross compiled from Linux for Windows via a mingw-x64 toolchain and `cargo build --target x86_64-pc-windows-gnu` — so that's something. But the failure is promising. It actually *does* compile, but then later fails in linking ... something about too many linker args for the toolchain. The issue is in linking Iroh (and yes the 0.9x releases have the same issue with my cross compile linker). The fact that it got through compilation suggests this code is probably right and we just need to figure out the right cross compile setup or compile it natively. Then test it for real.

alerque:
I did a bit more work on this. In particular I've been cleaning up some of the naming and locations of things to cut down on the amount of duplicated code and miss-named bits (like `socket_path` being used to create a named pipe). Additionally for convenience running tests I rebased in on #552 since I expect that to land soon and this to take a bit longer.

alerque:
I think this is ready for review — not as the end of the road but at least as a huge step forward. Windows support may or may not be fully functional yet but it needs to get some test users looking at this and giving it a thumbs up or a thumbs down with some comments on their experience.

As of now the `cargo check` and `cargo build` targets work for Windows (tested locally with cross compilation and in CI natively, both work) but it hasn't actually been test driven. The e2e testing also is Unix only at this point.

Note CI should have artifacts that can be downloaded and tested by anybody on Windows without needing to setup tooling to build it themselves. I'll post a link to that in an issue when Ci finishes but it should be here under the checks tab.

alerque:
This was already stacked on #541 which hasn't gotten a complete review yet. Since we actually did get through a top to bottom review of #560 today and I'm hopeful that will land in quick succession from 541, I just went ahead and rebased this to also be stacked on #560. The PR substance isn't really dependent on it, but there are a large number of conflicts between the two channels of work. Since the merge conflicts would have to be dealt with either way and this will land behind the others I went ahead and worked through the conflicts now.

alerque:
I was able to test this in a VirtualBox VM (bone stock Windows 11) and confirmed this does run ... but there are some path handling problems.

image

alerque:
> *Hello ~~darkness~~ Windows, my old friend;  
> I've come to talk with you again.*

image

As seen in the screen shot the Windows named pipe works, the networking stuff works, I can join in from another system, create files, send file edits, etc.

The remote side gets the Windows files once, but file create/edit events don't trigger sending the files again. I assume the RPC interface would work for an editor, but the file system watcher is evidently not working yet.

alerque:
The connecting both a Windows share daemon and Windows client works, but as soon as it gets edit messages from a remote peer the Windows daemon crashes:

> Should be able to create File URI from absolute path: File URI 'file://\\?\C:\Users\Administrator\Desktop\from_linux.txt' does not start with 'file:///'

alerque:
Closing in favor of a new PR #631 from the canonical repo so this can be managed with GH Stacks.

PR #535, opened 2026-05-01
Fix integration tests to not load user's Neovim config

alerque:
Here's an easy one. I've been unable to test locally even when CI was passing. This is why ;-)

**Self-review checklist**


zormit:
Good catch ;)

alerque:
This didn't go nearly far enough. It fixed the user *configs* from getting in the way, but it still uses the user's `&runtimepath` and hence *can* load all sorts of plugins, plus it requires the user to have the `teamtype-nvim` plugin installed. Furthermore, it goes on to use the user's system `teamtype` binary for at least some of the tests. At the end of the day this fix may have helped me pass tests locally, but it is not actually testing anything useful — in the ends it's testing a random hodge podge of mismatched bits, some from the user's personal `$HOME` environment, others from the repo at the time of testing.

PR #536, opened 2026-05-01
Avoid 'path must be shorter than SUN_LEN' when sharing long winded paths

alerque:
Closes #462

**Self-review checklist**

- [x] I described the changes of the PR in the [CHANGELOG](https://github.com/teamtype/teamtype/blob/main/CHANGELOG.md).
- [x] I have manually tested and self-reviewed my changes.
- [x] I have added myself to the REUSE headers in the files where I made significant changes, see [CONTRIBUTING.md](https://github.com/teamtype/teamtype/blob/main/CONTRIBUTING.md).
- [x] This PR does not contain content generated by LLMs, see our [generative AI policy](https://github.com/teamtype/teamtype/blob/main/CONTRIBUTING.md).

PR #537, opened 2026-05-02
Avoid 'path must be shorter than SUN_LEN' when sharing long winded paths

### LINKED ISSUE #462: share cannot create sockets on a long path (path must be shorter than SUN_LEN)
**Describe the bug**
The `share` command is unable to create a unix socket on folders whose full path is longer than a set amount (still not exactly sure on the number), where `$(pwd)/.teamtype/socket` is longer than `SUN_LEN` (108 in Linux, 104 in macOS).

**To Reproduce**
1. Create a folder with a name longer than 100 chars: `mkdir --verbose $(openssl rand -hex 100)`. Alternatively, create a sequence of nested folders whose full path exceeds 100 chars. The limit (108 or 104) will be reached when teamtype adds `.teamtype/socket`.
2. `cd` into it.
3. `teamtype share`
4. See error: `path must be shorter than SUN_LEN`

**Expected behavior**
The folder should be shared.

If the path is too long for teamtype to create a socket, the socket should be created on `$XDG_RUNTIME_DIR` with a hash of the folder path or another suitable location, and a softlink can be made from `.teamtype/socket` to the actual location.

**Versions:**
- `teamtype --version`: 0.9.0
- Which text editor are you using to collaborate, and which version? None
- Which version of the editor plugin are you using? None

alerque:
Closes #462

(This is a repost of #536 which I accidentally opened as a PR from my fork’s default branch.)

**Self-review checklist**

### DIFF
--- a/daemon/src/editor.rs
+++ b/daemon/src/editor.rs
@@ -1,11 +1,14 @@
 // SPDX-FileCopyrightText: 2024 blinry <mail@blinry.org>
 // SPDX-FileCopyrightText: 2024 zormit <nt4u@kpvn.de>
+// SPDX-FileCopyrightText: 2026 Caleb Maclennan <caleb@alerque.com>
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
 //! This module is all about daemon to editor communication.
 
-use std::{fs, os::unix::fs::PermissionsExt, path::Path};
+use std::os::unix::fs::PermissionsExt;
+use std::path::{Path, PathBuf};
+use std::{env, fs};
 
 use anyhow::{Context, Result, bail};
 use futures::StreamExt;
@@ -82,6 +85,14 @@ fn is_user_readable_only(socket_path: &Path) -> Result<()> {
     Ok(())
 }
 
+fn strip_current_dir(path: &Path) -> PathBuf {
+    let Ok(cwd) = env::current_dir() else {
+        return path.to_path_buf();
+    };
+    path.strip_prefix(&cwd)
+        .map_or_else(|_| path.to_path_buf(), Path::to_path_buf)
+}
+
 /// # Panics
 ///
 /// Will panic if we fail to listen on the socket, or if we fail to accept an incoming connection.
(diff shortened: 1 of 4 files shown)


alerque:
The second commit here fixes the `teamtype client --directory /some/other/long/path` use case. The code for this is not DRY and more convoluted than it needs to be, but fixing it properly requires some other refactoring that I decided not to try to mix into this PR. Instead I opened #538 with some notes to track that issue.

daemon/src/editor.rs
```diff
@@ -85,7 +85,7 @@ fn is_user_readable_only(socket_path: &Path) -> Result<()> {
     Ok(())
 }
 
-fn strip_current_dir(path: &Path) -> PathBuf {
+pub fn strip_current_dir(path: &Path) -> PathBuf {
```
alerque: Use `pub(crate)` here since we don't need to be exporting this as a public API. That's going to get cluttered enough without our internal convenience functions joining the party.

daemon/src/jsonrpc_forwarder.rs
```diff
@@ -15,11 +15,11 @@
 //! - takes content-length encoded data from stdin (as sent by an LSP client) and writes it
 //!   "unpacked" to the socket
 
-use std::env;
-use std::path::{Path, PathBuf};
+use std::path::Path;
 
 use async_trait::async_trait;
 use futures::{SinkExt, StreamExt};
+use teamtype::editor::strip_current_dir;
```
alerque: use super::strip_current_dir; I think will do the trick.

blinry:
I rebased this PR onto main after we merged #540, and verified that they fix the problem! :tada:

I also added two suggestions to the branch:

- make `strip_current_dir` public in the teamtype crate, so that the client can re-use it
- slight clarification of a comment, to make the other referred comment easier to find

Feel free to squash these into the others, if you'd feel happy about them!

blinry (approved):

alerque:
Sure ... and I can apply the scoping nit I just commented on.

alerque:
Gah. I went down this rabbit hole once and forgot. When I saw your suggestion I somehow assumed you'd gotten this into the same module, but no such luck.

The namespaces here are bonkers because the jsonrpc_forwarder is actually in a private module in the bin crate, the editor is is the lib crate. They are not part of the same crate at all in spite of the code proximity. I have another PR in the works to untangle the workspace issue, so I guess I'll put off the scope issues until then.

`pub(crate) fn ...` works fine but since we can actually *only* access the public interface here we either have to repeat ourselves or just make it public. I guess for now I'll make it public even though I want it to go away, but document that it shouldn't be used somehow.

PR #540, opened 2026-05-03
Fix client shut down and daemon startup

blinry:
We introduced a problem in #533: `teamtype client` wouldn't shut down on its own after errors, for example, not being able to connect to the daemon's socket. 

run_daemon returned a handle to the daemon, but the daemon is doing its work in other threads/tasks. So we want to wait for an external shutdown signal after calling it.

run_client, on the other hand, returns only when it errors. The returned Result should determine the exit code of the main program. We don't want an additional shutdown capture after that. `teamtype client` doesn't need to do any cleanup, so going with the default behaviour where shutdown signals just kill the process seems fine - we don't need to set up the shutdown trap ourselves.

After restructuring this, it seemed to me that the Mode enum was no longer needed and I removed it again.

I added a couple of comments to help future readers understand this. The different behaviour of the two modes is a tad confusing, but I'd rather push a quick fix now and possibly tackle this in a future "proper shutdown" PR.

**Self-review checklist**

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -40,12 +40,6 @@ fn has_teamtype_directory(dir: &Path) -> bool {
     sandbox::exists(dir, &teamtype_dir).expect("Failed to check") && teamtype_dir.is_dir()
 }
 
-#[derive(PartialEq)]
-enum MainMode {
-    Daemon,
-    Client,
-}
-
 #[tokio::main]
 async fn main() -> Result<()> {
     let default_panic = std::panic::take_hook();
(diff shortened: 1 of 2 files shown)


alerque:
Yes, something was/is off here. I agree there is a new problem introduced by #533, but at least at first glance this seems to partially revert to the bug we were trying to fix. In testing to see if that is really the case just now I ran into a different problem entirely. In fact I think you might have misdiagnosed what the new problem even is and introduced a new glitch by trying to fix the wrong problem. Before I try to explain I need to sleep though! I'll prioritize this tomorrow and either work out a reproduction of what I think I just saw or set myself straight with fresh eyes and get this approved if this fix is actually working right.

alerque:
Notes as I work through this...

- The issue may not be with the *client* at all. The share server works and can be used for remote collaboration, but it is creating a local socket, then instantly terminating it leaving the client with nothing to connect to in the first place.
- The file watcher is blocking async execution, and also has an unhandled panic (see `unwrap()` on watcher.rs:59:36)
- There are way to many levels of nested thread spawning!
- The watcher panic seems to happen because the thread it's running in is already in a termination mode due to the socket having gone away, the elided error is a `SendError` after several success and I think it's failing because the node went away while the notify watcher is still being created, basically a race condition
- db2e713 is reverting the important bits of the original fix and even though it is **eventually** trapping the kill signals, it is doing so *sequentially*. For example "CTRL+C" first kills off the daemon thread, *then* spawns a thread where it traps signals and figures out that a "CTRL+C" is already in progress. It is not catching the signal in parallel and leaving a chance for clean shutdown.
- None of the original, my 533 work, or this db2e713 fixup correctly remove the socket in the `join` mode.
- The commit message and code comment don't reflect what is actually happening with the daemon's handle. The handle isn't being captured, it is the output result of the joined handle, and the issue isn't that it needs to be captured, it is that it needs to *not* go out of scope and get dropped until other factors have spelled the end of life for the thread.

alerque:
After wasting¹ significant chunks of two days banging my head against this I've come to the conclusion it is un-fixable without a significant refactor of other internals. Given that won't happen overnight I concur with @blinry's conclusion that a quick hack to make it behave better even though it isn't *correct* is better done sooner than later. This is still better than what we had before where shutdown signally was totally perfunctory and gives us a little bit of room to grow and flex as we root out the real problem.

Squashing about 50 test commits together in the end the only thing I ended up changing was moving the early return vs. signal catching back into the main loop, including a messaging channel where nested threads can start listening for events, sending a clean shutdown event to anybody listening, and being a bit more precise in code comments about why we're being hanging onto the joined handle results.

<sub>¹ I suppose I shouldn't say "wasted". I've used Tokio before but never dived quite so deep down the rabbit hole of nested threading!</sub>

alerque (approved):

alerque:
I didn't put it it this branch yet, but while working on the main loop cleanup I came up with a small follow up that clarifies the scope of data a tiny bit: I split up the share & join config handling into helper functions. Eventually the `run_daemon()` should probably be split up and the logic for it downstreamed into separate modules to make them easy to call from library functions, but again that is out of scope for here. You can have a look in 093770a and cherry pick it into this branch if you like it.

PR #541, opened 2026-05-05
Colorize log messages from tracing used as user facing messages

### LINKED ISSUE #405: trace/log output should be colored to enable warnings to stick out
the non debug output is all one color which makes it hard to filter for important things

alerque:
Closes #405

**Self-review checklist**

### DIFF
--- a/crates/teamtype/src/config.rs
+++ b/crates/teamtype/src/config.rs
@@ -11,6 +11,7 @@ use std::path::{Path, PathBuf};
 
 use anyhow::bail;
 use anyhow::{Context, Result};
+use docstr::docstr;
 use git2::{Config as GitConfig, ConfigLevel, Error as GitError, Repository};
 use ini::{Ini, Properties};
 use tracing::info;
@@ -238,13 +239,14 @@ fn get_username_from_git(base_dir: &Path) -> Option<String> {
     let username = get_git_username(base_dir);
     if let Some(ref username) = username {
         info!(
-            "Using the Git username '{username}' as username, to display next to the cursors other people see."
+            "{}",
+            &docstr!(format!
+                /// Using the Git username '{username}' as username, to display next to the cursors other people see.
+                /// Teamtype uses the Git username as username by default.
+                /// You can set the configuration value `username` in your `.teamtype/config` to override this username.
+                /// You can also use the flag `--username` when using the `share`/`join` subcommands.
+            )
         );
-        info!("Teamtype uses the Git username as username by default.");
-        info!(
-            "You can set the configuration value `username` in your `.teamtype/config` to override this username."
-        );
-        info!("You can also use the flag `--username` when using the `share`/`join` subcommands.");
     }
     username
 }
@@ -252,12 +254,13 @@ fn get_username_from_git(base_dir: &Path) -> Option<String> {
 fn get_username_from_fallback_value() -> String {
     let username = USERNAME_FALLBACK.to_string();
     info!(
-        "Using the fallback value for username '{username}' as username, to display next to the cursors other people see."
+        "{}",
+        &docstr!(format!
+            /// Using the fallback value for username '{username}' as username, to display next to the cursors other people see.
+            /// You can set the configuration value `username` in your `.teamtype/config` to override this username.
+            /// You can also use the flag `--username` when using the `share`/`join` subcommands.
+        )
     );
-    info!(
-        "You can set the configuration value `username` in your `.teamtype/config` to override this username."
-    );
-    info!("You can also use the flag `--username` when using the `share`/`join` subcommands.");
     username
 }
 
--- a/crates/teamtype/src/logging.rs
+++ b/crates/teamtype/src/logging.rs
@@ -5,11 +5,19 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
 use std::env;
+use std::fmt;
 
 use anyhow::Result;
+use nu_ansi_term::{Color, Style};
 use time::macros::format_description;
 use tracing::level_filters::LevelFilter;
 use tracing::subscriber;
+use tracing::{Event, Level, Subscriber};
+use tracing_subscriber::fmt::{
+    FmtContext,
+    format::{FormatEvent, FormatFields, Writer},
+};
+use tracing_subscriber::registry::LookupSpan;
 use tracing_subscriber::{EnvFilter, FmtSubscriber, fmt::time::UtcTime};
 
 pub fn initialize() -> Result<()> {
@@ -18,9 +26,8 @@ pub fn initialize() -> Result<()> {
     if simplified_logging {
         let subscriber = FmtSubscriber::builder()
             .with_env_filter(EnvFilter::new("teamtype=info,fuzzer=info"))
-            .without_time()
-            .with_level(false)
-            .with_target(false)
+            .with_ansi(true) // Let the rest of the builder understand we are colorized.
+            .event_format(ColoredLine)
             .finish();
         subscriber::set_global_default(subscriber).expect("Setting default log subscriber failed");
     } else {
@@ -32,9 +39,59 @@ pub fn initialize() -> Result<()> {
             .with_env_filter(filter)
             .with_thread_ids(true)
             .with_timer(timer)
+            .with_level(true)
+            .pretty()
             .finish();
         subscriber::set_global_default(subscriber).expect("Setting default log subscriber failed");
     }
 
     Ok(())
+}
+
+/// Modified clone of `Compact` formatter from tracing-subscriber. Eschews the optional level
+/// labels, time formatting, and target options for hard coded defaults and uses the same ANSI
+/// handling internals to colorize the whole line instead of only the level label.
+struct ColoredLine;
+
+impl<S, N> FormatEvent<S, N> for ColoredLine
+where
+    S: Subscriber + for<'a> LookupSpan<'a>,
+    N: for<'a> FormatFields<'a> + 'static,
+{
+    fn format_event(
+        &self,
+        ctx: &FmtContext<'_, S, N>,
+        mut writer: Writer<'_>,
+        event: &Event<'_>,
+    ) -> fmt::Result {
+        let mut buf = String::new();
+        let buf_writer = Writer::new(&mut buf);
+        // We told the event builder we're using ANSI escapes here, but we are about to wrap the
+        // output of the existing `Compact` formatter and we do *not* want it to colorize stuff
+        // for us. We'll do that by wrapping the whole buffered output.
+        let fmt = tracing_subscriber::fmt::format()
+            .compact()
+            .with_ansi(false)
+            .without_time()
+            .with_target(false)
+            .with_level(false);
+        fmt.format_event(ctx, buf_writer, event)?;
+        // We want the ANSI coloring to be inside the trailing newline of each message.
+        let line = buf.trim_end_matches('\n');
+        let style = if writer.has_ansi_escapes() {
+            match *event.metadata().level() {
+                Level::ERROR => Color::Red.normal(),
+                Level::WARN => Color::Yellow.normal(),
+                // The typical color for info logs would be `Green`, but that is overbearing when
+                // colorizing the whole line, so just render info messages in the terminal default.
+                Level::INFO => Style::default(),
+                Level::DEBUG => Color::Blue.normal(),
+                Level::TRACE => Color::Purple.normal(),
+            }
+        } else {
+            return writeln!(writer, "{line}");
+        };
+        write!(writer, "{}", style.paint(line))?;
+        writeln!(writer)
+    }
 }
--- a/crates/teamtype/src/main.rs
+++ b/crates/teamtype/src/main.rs
@@ -14,6 +14,7 @@ use std::{env, panic};
 use anyhow::bail;
 use anyhow::{Context, Result};
 use clap::{CommandFactory as _, FromArgMatches as _};
+use docstr::docstr;
 use microxdg::XdgApp;
 use teamtype::jsonrpc_forwarder::{JSONRPCForwarder, UnixJSONRPCForwarder};
 use teamtype::{
@@ -99,9 +100,12 @@ async fn run_daemon(app_config: AppConfig, init_doc: bool) -> Result<Daemon> {
     config::ensure_teamtype_is_ignored(&app_config.base_dir)?;
 
     if app_config.sync_vcs && config::has_local_user_config(&app_config.base_dir).is_ok_and(|v| v) {
-        warn!(
-            "You have a local user configuration in your .git/config. In --sync-vcs mode, this file will also be synchronized between peers. If your version \"wins\", all peers will have the same Git identity. As a workaround, you could use `git commit --author`."
-        );
+        warn!(docstr!(
+            /// You have a local user configuration in your .git/config.
+            /// In --sync-vcs mode, this file will also be synchronized between peers.
+            /// If your version "wins", all peers will have the same Git identity.
+            /// As a workaround, you could use `git commit --author`.
+        ));
     }
 
     debug!("Starting Teamtype on {}.", app_config.base_dir.display());
--- a/crates/teamtype/src/wormhole.rs
+++ b/crates/teamtype/src/wormhole.rs
@@ -6,6 +6,7 @@
 use std::{borrow::Cow, str::FromStr, time::Duration};
 
 use anyhow::Result;
+use docstr::docstr;
 use magic_wormhole::{AppConfig, AppID, Code, MailboxConnection, Wormhole, transfer};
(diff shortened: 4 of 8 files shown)


alerque:
Output formats approved, waiting on merging sequential log messages and making sure the indentation carries over...

alerque:
This is all wrong. The `tracing::info!` macro is just entirely wrong for this use case. The CLI is abusing the macro as a substitute for actual STDIO output, not logging messages at all.

Logging is supposed to have levels and you can typically control the output by setting a filter level.

1. None (silent modes)
1. Error
1. Warn
1. Info
1. Debug
1. Trace

Typically an app should be usable at level `NONE`, but would default to running at `WARN` level. You could then ramp up the verbosity to various levels depending on how seriously your were debugging things. An `INFO` level should be useful for people casually wondering what the system is up to under the hood and should explain what's being done why, but should not be essential for function of the program. At the moment we've got stuff being output via `info!` that you basically can't use the app without (e.g. share codes). Additionally we have a number of things that are less about informing about the current status and more just general usage help messages. In other words the logging system is being overloaded to be a CLI user interface.

There are several ways to format multi line strings that work in one or more of the log modes, but they stop making sense really fast when you try to make *multiple* log formats work. I think this is a problem because the message consumer may be a CLI, but it also may be an editor interface that is catching an relaying them.

I think we need to overhaul the CLI interactions to keep all log messages to one line like most log message viewing system expect. Required user interactions like join codes should go straight to STDOUT, probably not through the logging system at all. We can also log errors, but we shouldn't be substituting an error message in the log for bubbling up an error type (we have `error!(); return;` in the code!).

alerque:
This is now stacked on #543 and #551. As of now it solves [the major architectural problem from my last comment](https://github.com/teamtype/teamtype/pull/541#issuecomment-4387854869) by splitting up the user facing messaging required to function (share codes, connect/disconnect notices, etc.) and outputting those to the console, and using the tracing based log messages as logging was meant to be. By default no logging is enabled at the warn level with concise messages and that can easily be bumped up to the info level with the new --verbose flag. The more verbose logging output format with source locations can still be triggered by setting the RUST_LOG variable.

Additionally several fixes to the actual messages slipped in here such as noting the actual peer line that would need to be added to a config to connect without the share code, etc.

Next step is to get all of this out of the library crate.

alerque:
As of now the `tracing` macros are only used for actual logging, all the log messages that were user interactions the user *had to see* (and a few they probably don't) have been converted to be actually handled by the UI. In the case of the CLI that just means printing them or reading prompts, but now the UI layer is completely transferable. We could write a TUI that wraps the output in something fancy and interactive instead of just printing it to the console. We can write library bindings where the linker is expected to give us an interface callback, etc.

alerque:
This PR started off using the `indoc` crate which I've used before and like and greatly improves the ergonomics of multiline strings that are supposed to have internal indenting. I'm quite happy with that and we can go that way too. Then today I discovered @nik-rev's `docstr!` macro and decided to try it on for size here. I like it even better! It has *different* ergonomic trade-offs and of course all of this is going to be up for discussion, but as of now I'm even happier with `docstr` than `indoc`.

alerque:
Now that this isn't stacked on anything and ready for review as a stand alone chunk of refactoring, I rebased it to change the commit order. It was roughly the order I did the work in, now it is in a more logical progression for review on a per-commit basis.

alerque:
Gah! The GitHub interface is laggy and when you load pages it moves stuff around. The stupid review requests UI keeps moving CoPilot to be the link under cursor after any update. I meant to tag Moritz not CoPilot!

crates/teamtype/src/wormhole.rs
```diff
@@ -6,11 +6,18 @@
 use std::{borrow::Cow, str::FromStr, time::Duration};
 
 use anyhow::Result;
+use docstr::docstr;
```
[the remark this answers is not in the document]
alerque: Why LLMs suck for jobs that can be (and in this case already are) done by deterministic tooling, *Exhibit A*: **THIS IS THE `docstr#` MACRO YOU MORON!**. :clown_face:

blinry:
> Gah! The GitHub interface is laggy and when you load pages it moves stuff around. The stupid review requests UI keeps moving CoPilot to be the link under cursor after any update. I meant to tag Moritz not CoPilot!

Has happened to me, as well! I've hidden the comment.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 
 # 0.9.2 (unreleased)
 
+Make daemon share and join modes less verbose by default; add a flag `-v`/`--verbose` to the CLI fear easy access to more chatty output.
```
zormit:
```suggestion
Make daemon share and join modes less verbose by default; add a flag `-v`/`--verbose` to the CLI for easy access to more chatty output.
```

crates/teamtype/src/cli.rs
```diff
@@ -29,6 +29,9 @@ fn get_version() -> &'static str {
 #[command(version=get_version(), about, long_about = None)]
 #[command(propagate_version = true)]
 pub struct Cli {
+    /// Output extra informational log messages
+    #[arg(short, long)]
+    pub verbose: bool,
```
zormit: I'm not sure I understand the architecture of the verbose output. It is now mixing the `UserInterface` output and the logging output. Is that on purpose? What I would expect is that I don't see any logging as a CLI user (after your PR).

zormit: What I mean, I would expect to see info/debug stuff only with the `RUST_LOG` approach.

crates/teamtype/src/config.rs
```diff
@@ -211,53 +213,61 @@ fn get_username(
...
+    general_section: &Properties,
+    ui: &UserInterface,
+) -> Option<String> {
     general_section
         .get("username")
         .map(ToString::to_string)
         .map(|username| {
-            info!("Using the username '{username}' from `.teamtype/config` as username, to display next to the cursors other people see.");
+            info!("Using the username '{username}' from `.teamtype/config` as username");
+            ui.inform(&format!(
```
zormit: This is in multiple places, so doesn't seem to be an accident - but it feels a bit redundant to log a very similar string. Maybe it's how it is, if you do it seriously, but it has a strange look.

Also when using the `--verbose` flag, both get shown, that's a bit noisy? If the first gets written to a file that would make more sense.

zormit (commented):
Thank you for preparing this PR! This looks overall pretty neat to me. Now the goal is to find the right balance and purpose for the "split up" logging vs. CLI UI output.

I didn't look thoroughly through everything yet, but I gathered some questions already.

alerque:
> Now the goal is to find the right balance and purpose for the "split up" logging vs. CLI UI output.

Yes ... and this isn't quite there yet. I did most of this work a month ago and just tonight while working on Lua bindings tonight ran into something that is off here. The "fix" is pretty easy—I might be able to get a commit out tonight to address the lack of errors outside of the logging system.

crates/teamtype/src/cli.rs
alerque: You're right, the split is incomplete here. I would expect the `--verbose` flag to enable logging to go to the console (as it is now) but the default non-verbose output should include the warning messages through a *non* logging channel as well. Stand by.

crates/teamtype/src/config.rs
alerque: Yes, there are quite a few of these. I started out trying *not* to do a bunch of copy editing in the process of refactoring how messages were handled, but I ended up having to do some.

Realistically we can *probably* get rid of most of the places logging and user messaging happens at exactly the same point. Where they do end up both being around, the purpose and wording of messages should be different. The normal user facing output can be a bit chatty and explain thing, logging should just cut straight to the chase and potentially include much more technical detail.

There are also a few more places (even after my new commits a few minutes ago) with `warn!` level logging that don't seem to be something the user *needs* to see in normal use, so they have not been converted to UI messages, but part of this PR review should probably be a second pair of eyes on those.

alerque: > If the first gets written to a file that would make more sense.

Adding an option to log to a file sounds like a reasonable thing to do at this point.

crates/teamtype/src/main.rs
```diff
@@ -294,16 +306,20 @@ fn get_directory(temporary_directory: Option<&TempDir>, cli: &Cli) -> Result<Pat
     Ok(directory)
 }
 
-fn setup_teamtype_directory(directory: &Path, temporary_directory: Option<&TempDir>) -> Result<()> {
+fn setup_teamtype_directory(
+    directory: &Path,
+    temporary_directory: Option<&TempDir>,
```
alerque: The message buried in logging about what the temp dir name ended up being should be a user information message not a debug log.

alerque: Fixed in amended commit, 5b85604 by promoting message about the temporary directory name from a log message to an essential user information message.

alerque:
> Thank you for preparing this PR! This looks overall pretty neat to me. Now the goal is to find the right balance and purpose for the "split up" logging vs. CLI UI output.

Yes. An @blinry found something on the wrong side of that split today. I [just upgraded the log message](https://github.com/teamtype/teamtype/compare/637a1a28c6a847179fd27985a40a8d1651a4ab78..6fe9906f25b8e855faae7e51af70efcd2881d1b2) that has the name of a generated temporary directory to a user message since it is basically essential to the expected workflow there.

We also discussed default verbosity a little bit and he suggested perhaps making verbose the default (with the info level log messages going to the console) and making the options flag be `-q`/`--quiet` instead. I'm fine with that either way. My only concern about that is it might lead us to not making good choices about what is essential user interface messaging that must be surfaced to the user during normal use vs. what is just informational output the user could monitor or not monitor. Defaulting to the chatty log info might make us hesitant to log things that might be actually useful in a debugging situation and mixing in things that are actually essential into the log stream.

blinry:
Gave this a cursory first try from the user side. Here's some things I noticed:

- I'm surprised by the new (y/n) confirmation behavior where you don't need to press enter anymore. What's your reasoning? Users not realizing that the need to press enter? It will be a bit jarring to old users at first, I think.
- The new (y/n) dialog doesn't look very pretty on my terminal, it seems to "repeat" the question after confirmation on narrow terminals:

		Detected an existing socket '/home/blinry/tmp/playground/.teamtype/socket'. There might be a daemon running already for this directory, or the previous one
        Detected an existing socket '/home/blinry/tmp/playground/.teamtype/socket'. There might be a daemon running already for this directory, or the previous one crashed. Do you want to continue? yes
- The new `-v` flag can only be provided directly after `teamtype`, not after the subcommands, which makes it less discoverable (in the `--help` entries of the subcommands, for example). Can we promote it to go after share/join, like `--directory`?
- With `--verbose`, the output contains mixed "regular" output and lines with "INFO:" prefix. Seems @zormit also was a bit surprised a bit. From a user perspective, it'd be nice if the messages were formatted the same, probably?
- When using `--verbose`, some messages are duplicated, like the Git username, or the join code.
- What is logged as "share code" in one place, we call "join code" instead.
- I'd miss some messages in the non `--verbose` mode: Mosty the messages about which files are initialized/changed/deleted. The "Editor #n connected" was also often useful as a confirmation. If all plugins would have a good status display, maybe we wouldn't need it, but I think we're not there yet.

Some general thoughts on the design of logging/showing messages to the user:

- While the "debug" messages probably don't need to be displayed in a user interface in "regular usage", I feel like maybe most messages we previously showed with "info!" should be displayed even in a TUI?
- To avoid having to duplicate messages going to the user and going to the logs in the code, might it make sense to have some kind of wrapper that does both?

alerque:
Downgraded to draft status because we talked about both splitting this up into two concerns and changing the approach to one of them a bit. I have in my head how to do that but since we're not going to merge until after the next release I've been working on that with what bits of time I've had this week.

alerque:
I just attempted to split up this PR into a couple stacked ones using `gh stack`. It seems to have succeed in adding this existing PR to a new stack, but please excuse the mess while I figure out how that tooling is supposed to work.

alerque:
This long running branch has been split into several PRs (usage `gh stack`). The only bits left in *THIS* PR is colorization of the existing log messages using almost identical formatting, just adding ANSI escape sequences for color to make `debug!` and `warn!` stand out from `info!`. It also adds a crate I think is convenient for handling multiline strings. It does not change the level of any existing messages or address the overloading of logging functions for UI. It is just ground work in an attempt to move on and make the later refactoring easier to reason about.

It does reformat one message from being a long run-on line that will wrap to a narrower pre-wrapped version that should be easier to read in many/most terminals.

This is stacked on #608 but due to a technicality I can't add that to the GH Stack handling used here for the PRs higher in the stack.

crates/teamtype/src/config.rs
```diff
@@ -238,26 +239,28 @@ fn get_username_from_git(base_dir: &Path) -> Option<String> {
...
     if let Some(ref username) = username {
         info!(
-            "Using the Git username '{username}' as username, to display next to the cursors other people see."
+            "{}",
+            &docstr!(format!
+                /// Using the Git username '{username}' as username, to display next to the cursors other people see.
+                /// Teamtype uses the Git username as username by default.;
+                /// You can set the configuration value `username` in your `.teamtype/config` to override this username.
+                /// You can also use the flag `--username` when using the `share`/`join` subcommands.;
+            )
```
blinry, line 248: This is surprisingly involved – you've probably tried/compared alternatives to have nice multiline strings?

One I see is something like

```rust
info!(concat!(
    "Using the Git username '{username}' as username, to display next to the cursors other people see.\n",
    "Teamtype uses the Git username as username by default.\n",
    "...\n",
));
```

Edit: Ah, this doesn't do the `format!`…

crates/teamtype/src/config.rs
```diff
@@ -238,26 +239,28 @@ fn get_username_from_git(base_dir: &Path) -> Option<String> {
     let username = get_git_username(base_dir);
     if let Some(ref username) = username {
         info!(
-            "Using the Git username '{username}' as username, to display next to the cursors other people see."
+            "{}",
+            &docstr!(format!
+                /// Using the Git username '{username}' as username, to display next to the cursors other people see.
+                /// Teamtype uses the Git username as username by default.;
```
blinry: A semicolon sneaked into the end of the line here, it seems! :>

crates/teamtype/src/logging.rs
```diff
@@ -32,9 +39,59 @@ pub fn initialize() -> Result<()> {
...
+        fmt.format_event(ctx, buf_writer, event)?;
+        // We want the ANSI coloring to be inside the trailing newline of each message.
+        let line = buf.trim_end_matches('\n');
+        let style = if writer.has_ansi_escapes() {
+            match *event.metadata().level() {
+                Level::ERROR => Color::Red.normal(),
+                Level::WARN => Color::Yellow.normal(),
+                // The typical color for info logs would be `Green`, but that is overbearing when
+                // colorizing the whole line, so just make it white here.
+                Level::INFO => Color::White.normal(),
```
blinry: In my current color scheme, this makes these log lines *gray*.

Image

Looking at a [table of default colors](https://en.wikipedia.org/wiki/ANSI_escape_code#3-bit_and_4-bit), it seems that foreground color 37 is usually not the same as "pure white".

So I'd suggest we use `nu_ansi_term::Style::default()` for INFO, which does not add escape codes to the line. All colored lines output by `nu_ansi_term` seem to end with `<Esc>[0m`, which resets the styles.

blinry (changes requested):

crates/teamtype/src/logging.rs
alerque: Good call, will adjust.

crates/teamtype/src/config.rs
alerque: Yikes. I know what rebase snafu that came from.

crates/teamtype/src/config.rs
alerque: Yes I evaluated lots of alternatives. The one you suggest is the closest to feeling okay for me, but I still don't like having to include newline escapes manually to get the string to work and I think it leads to inconsistencies such as when to include a trailing newline in a multiline message and when not to. Also it gets hard to visualize what is happening with indentation and blank lines. Also it needs another level of format expansion, which makes it messier than your example.

I agree this specific code *is* a bit over complex, but I'll also note it gets a simpler in a later iteration of this PR stack. Part of the problem here is feeding the multi-line string into these macros that require a fixed string, hence the dumb looking indirect expansion via a `format!("{}", ...)` macro instead of just using the result directly. I went ahead and used this as a placeholder at this point in the commit stack because the proper fix that makes it a lot cleaner in actual usage doesn't happen until later in the sequence.

blinry (approved):

blinry:
Thanks for cooking this up! 🍲

PR #542, opened 2026-05-06
Follow up on signal handling refactor, cleanup variable scope

alerque:
This is a follow up on #540. Just code refactoring, no user facing changes. The goal here is to differentiate what belongs to what mode by  gradually cleaning up the config scope daemons, clients, etc. can send/recieve their respective signals, be cleanly managed from the library crate API surface, etc. This change avoids the dual conditional constructs where both join & share modes are iterating over the same data in the same function. The result makes it clear, for example, that the `init` flag coming from the CLI is being passed though to the daemon function, but not via the `AppConfig` construct — the construction of which dominated that section of code.

**Self-review checklist**

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -62,14 +62,21 @@ async fn main() -> Result<()> {
     setup_teamtype_directory(&directory, temporary_directory.as_ref())
         .context("Failed to find .teamtype/ directory")?;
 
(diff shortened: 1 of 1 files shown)


blinry:
I like it. Chopping long functions into smaller ones is great. Thanks!

PR #543, opened 2026-05-06
Use Cargo virtual workspace instead of nesting, rename crates

alerque:
Not finished yet, to be discussed.

**Self-review checklist**


blinry:
(Just quickly commenting to say that I like this direction – having a directory for crates would also allow us to pull out the "one-sided OT" functionality, for example, which is mainly independent from the rest of Teamtype, but could be helpful for other projects dealing with the Teamtype editor protocol!)

alerque:
> having a directory for crates would also allow us to pull out the "one-sided OT" functionality, for example, which is mainly independent from the rest of Teamtype

Yes. The details of what this looks like and what we call it (e.g. this paves the way to get away from the 'daemon' naming [per discussion](https://teamtype.zulipchat.com/#narrow/channel/562985-daemon/topic/Is.20.22daemon.22.20a.20misnomer.3F/with/585626089) but it doesn't commit to calling it anything in particular) is up for discussion. I've been "trying on" different names for test crates and other bits.

But specifics aside, I think the current repository layout is very inflexible and I expect this will not only cleanup the tool usage as developers but open things up to splitting concerns up into separate crates. We can compartmentalize some of our own internal interfaces, but I don't feel strongly about needing to be too granular. I was more thinking about bindings and setting up shared libraries that can be used directly for Rust editors and through language specific binding crates for others; e.g. NeoVIM could use crate that generates a Lua module via `mlua` dependencies, Python, JS, WASM, and others could likewise do the same for other editor environments. This is easier managed with a bindings crate than feature gating in the main library.

alerque:
I'm going to break this PR up as it has started to grow past the initial scope and I'm now working on more things branched off of it — and I definitely like the workspace layout better and think we should move this direction.

Do I need to outline the case for actually getting everything into one workspace and not having the crate published as 'teamtype' in a directory called 'daemon' — or is that all self evident at this point?

Whether or not to use a nested `crates` directory or top level directories is of course up for discussion. It can be made to function about the same either way. Additionally the name of the e2e tests crate is up for grabs, although long term I think the best approach for that may be to split it up into more crates splitting out the parts that remote control editors (our meta control channel) from the actual tests we want to do and also separate from fuzzing, etc.

crates/e2e-tests/Cargo.toml
```diff
@@ -6,41 +6,33 @@
...
+authors.workspace = true
+license.workspace = true
+edition.workspace = true
+rust-version.workspace = true
 
 [[test]]
 name = "fuzzer"
 # We set process-global state in this test, so we want to run our own main method.
 harness = false
+test = false # don't include fuzzer in default test runs
```
blinry:
```suggestion
test = false # Don't include fuzzer in default test runs.
```

crates/teamtype/Cargo.toml
```diff
@@ -0,0 +1,114 @@
...
+[[bin]]
+name = "teamtype"
+required-features = ["cli"]
+
+[lib]
+name = "teamtype"
+crate-type = ["rlib", "cdylib"]
+
+[features]
+default = ["cli"] # this is unfortunate, but library users should use `default-features = false`
```
blinry:
```suggestion
default = ["cli"] # This is unfortunate, but library users should use `default-features = false`.
```

crates/teamtype/Cargo.toml
```diff
@@ -0,0 +1,114 @@
...
+[dependencies]
+anyhow.workspace = true
+async-trait.workspace = true
+automerge = "0.8"
+constant_time_eq = "0.4"
+dissimilar = "1"
+futures.workspace = true
+ignore = "0.4"
+iroh = "0.35" # https://www.iroh.computer/blog/iroh-0-90-the-canary-series
+magic-wormhole = "0.7" # 0.8 fails on macos due to a bad interaction with iroh's netwatcher usage
```
blinry: While we're at it, maybe we can make these comments into "full sentence"-style, as well :}

blinry (changes requested):
Reminder as to the comment style :)

Cargo.toml
```diff
@@ -0,0 +1,86 @@
...
+authors = [
+    "Moritz Neeb <nt4u@kpvn.de>",
+    "blinry <mail@blinry.org>",
+    "Caleb Maclennan <caleb@alerque.com>",
+]
+homepage = "https://teamtype.github.io/teamtype/"
+repository = "https://github.com/teamtype/teamtype"
+license = "AGPL-3.0-or-later"
+edition = "2024"
+rust-version = "1.85"
```
zormit: It seems this was changed back to 1.85 from 1.89, but my (MacOS) `cargo msrv find` evaluates it to 1.89:

```
Compatibility Check #4: Rust 1.89.0
  [OK]     Is compatible

Compatibility Check #5: Rust 1.86.0
  [FAIL]   Is incompatible

  ╭────────────────────────────────────────────────────────────────────────╮
  │ error: rustc 1.86.0 is not supported by the following packages:        │
  │   automerge@0.8.0 requires rustc 1.89.0                                │
  │   automerge@0.8.0 requires rustc 1.89.0                                │
  │ Either upgrade rustc or select compatible dependency versions with     │
  │ `cargo update <name>@<current-ver> --precise <compatible-ver>`         │
  │ where `<compatible-ver>` is the latest version supporting rustc 1.86.0 │
  │                                                                        │
  │                                                                        │
  ╰────────────────────────────────────────────────────────────────────────╯

....

Result:
   Considered (min … max):   Rust 0.11.0 … Rust 1.95.0
   Search method:            bisect
   MSRV:                     1.89.0
   Target:                   x86_64-apple-darwin

```

alerque:
The gitignore needs to get moved to the top level.

crates/e2e-tests/Cargo.toml
alerque: I actually prefer punctuated fragments for line-end comments and full punctuated sentences for stand alone comment lines, but obviously this was a bad middle ground—a full sentence but tacked on the end without punctuation. I'll just promote this comment to a line of its own which is also consistent with other comments in the same section.

Cargo.toml
alerque: I don't know how this happened now, I suspect I must have re-tested it part way through a rebase or something like that, but clearly this wrong.

alerque:
I think all the nits are addressed now. In retrospect the comment style changes probably should have been separately from the manifest reorganization, but untangling it into a separate commit now is probably more trouble than it's worth.

alerque:
This just got a small update. Cargo has some new workspace manifest linting ability, so I figured I'd double check this with that tooling. It highlighted the fact that I didn't inherit the homepage or clippy lint rules into the crate(s). That was a small fix, but fixing that meant that CI actually started linting again with the project specific lint preferences, which it hadn't been. That along with the new Rust version is surfacing a few new lints, so now there is a commit to fixup those as well.

blinry:
Added two things I felt were missing on top. Please check whether those make sense. Thumbs up from me if the CI comes out green.

While testing, I noticed that the e2e tests no longer work for me locally, but that seems the case since before this PR. Will need to bisect.

blinry (approved):

zormit (approved):
Just carefully read through everything again and I also approve :)
thanks for reorganizing and laying it out in easily digestible commits

PR #544, opened 2026-05-07
Fixup e2e tests to leave user's plugin environment behind

alerque:
My previous PR #535 was onto something but [didn't go far enough](https://github.com/teamtype/teamtype/pull/535#issuecomment-4400820934). It turns out running tests locally really wasn't isolated at all.

Draft mode for the moment because this still needs a way to have the plugin as tested use the binaries as built at time of testing, not just a random system one that the dev may or may not have just built and installed in the `$PATH` to match. That's happening in CI right now, but not necessarily locally. That part *might* have to wait for my workspace fixes in #543 that get the testing into the same workspace as the bin & lib crates, but I'll at least look into whether it is viable to fix here first.

**Self-review checklist**

### DIFF
--- a/daemon/integration-tests/src/actors.rs
+++ b/daemon/integration-tests/src/actors.rs
@@ -27,7 +27,7 @@ pub trait Actor: Send {
 }
 
 pub struct Neovim {
-    nvim: nvim_rs::Neovim<Compat<ChildStdin>>,
+    pub nvim: nvim_rs::Neovim<Compat<ChildStdin>>,
     buffer: nvim_rs::Buffer<Compat<ChildStdin>>,
 }
 
@@ -35,24 +35,35 @@ impl Neovim {
     /// # Panics
     ///
     /// Will panic if Neovim cannot be started or if the file cannot be opened.
-    pub async fn new(file_path: PathBuf) -> Self {
+    pub async fn new(file_path: Option<PathBuf>) -> Self {
         let handler = Dummy::new();
         let mut cmd = Command::new("nvim");
-        cmd.arg("--headless").arg("--embed");
-        // Disable loading user RC files, to prevent unrelated local test failures
-        cmd.arg("-u").arg("NORC");
-        // Disable ShaDa files, to prevent CI failures related to them.
-        cmd.arg("-i").arg("NONE");
-        // Disable Swap file, to prevent CI failures related to them.
+        cmd.arg("--headless");
+        cmd.arg("--embed");
+        // Use factory defaults, disable user config, plugins, and shada.
+        cmd.arg("--clean");
+        // Also disable fallback from shada to swap files.
         cmd.arg("-n");
+        // Add the local Neovim plugin tree to the plugin path.
+        let workspace_root = PathBuf::from(env!("CARGO_WORKSPACE_DIR"));
+        let workspace_root = workspace_root.to_string_lossy();
+        cmd.arg("-c").arg(format!(
+            r#"let &runtimepath="{workspace_root}/nvim-plugin,".&runtimepath"#
+        ));
+        // Load the Neovim plugin into the instance.
+        cmd.arg("-c").arg(format!(
+            "source {workspace_root}/nvim-plugin/plugin/teamtype.lua"
+        ));
         let (nvim, _, _) = new_child_cmd(&mut cmd, handler).await.unwrap();
 
         // We canonicalize the path here, because on macOS, TempDir gives us paths in /var/, which
         // symlinks to /private/var/. But the paths in the file events are always in /private/var/.
-        let file_path = file_path.canonicalize().unwrap();
-        nvim.command(&format!("edit! {}", file_path.display()))
-            .await
-            .expect("Opening file in nvim failed");
+        if let Some(file_path) = file_path {
+            let file_path = file_path.canonicalize().unwrap();
+            nvim.command(&format!("edit! {}", file_path.display()))
+                .await
+                .expect("Opening file in nvim failed");
+        }
 
         let buffer = nvim.get_current_buf().await.unwrap();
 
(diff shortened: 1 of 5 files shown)


alerque:
This should be fixed up in two steps:

1. Use the Neovim plugin directly from the repository for tests, not the one the user installed in their home profile.
2. Use the `teamtype` binary from the test run, not whatever happens to be in `$PATH`.

The caveat here is that part 2 uses a Cargo feature that is not stabilized yet. For my part I'd be fine with requiring the e2e test crate requiring `+nightly` as long as the user-facing crates (and unit tests) work on `+stable`. That being said I will look into what the alternatives would be. I suspect we'd have to hack together the path handling in a `build.rs` and that it would be possible but that it would be ugly to make this work without the built in artifact handling. I'm just out of time for the moment to look into that.

daemon/integration-tests/src/actors.rs
```diff
@@ -27,32 +27,52 @@ pub trait Actor: Send {
...
+        cmd.arg("--embed");
+        // Disable loading user RC files plugins
+        cmd.arg("-u").arg("NONE");
         // Disable ShaDa files, to prevent CI failures related to them.
         cmd.arg("-i").arg("NONE");
         // Disable Swap file, to prevent CI failures related to them.
         cmd.arg("-n");
+        // Add the local Neovim plugin tree to the plugin path
+        let project_root = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        let project_root = project_root
```
alerque: This unsightly parent().parent() jumble will go away in workspace refactoring as in #543 because the integration tests will end up properly housed the same workspace and hence Cargo will be able to find the root manifest (technically the 'virtual workspace manifest' in Cargo lingo) obviating the need for this hard coded jumping around.

alerque:
I backed out the part 2 fix because it needs discussion or experimentation, this part 1 should be a pretty cut and dry win.

alerque:
@zormit Does this PR by chance help you at all with being able to run tests locally on your macOS? If it does, great. If it does not, can you open a new issue with your output from `cargo test --manifest-path=daemon/integration-tests/Cargo.toml --locked` so I can an eye out for what the root cause of that really is?

daemon/integration-tests/src/actors.rs
```diff
@@ -27,32 +27,50 @@ pub trait Actor: Send {
...
         let mut cmd = Command::new("nvim");
-        cmd.arg("--headless").arg("--embed");
-        // Disable loading user RC files, to prevent unrelated local test failures
-        cmd.arg("-u").arg("NORC");
-        // Disable ShaDa files, to prevent CI failures related to them.
-        cmd.arg("-i").arg("NONE");
-        // Disable Swap file, to prevent CI failures related to them.
+        cmd.arg("--headless");
+        cmd.arg("--embed");
+        // Use factory defaults, disable user config, plugins, and shada
```
blinry: Here's a coding convention that I care about for some reason: We try to end comments that are full sentences with a period. I think we probably picked that up from the [Rust Style Guide](https://doc.rust-lang.org/style-guide/#comments).

Might be cool to document that somewhere. I opened #549 to track that.

Full review of this PR coming up next time! :) Cool that you got this to work!

alerque: I can get on board with that! I've gone both ways on projects before and find some projects tend to have lots of small comments than are sentence fragments, but given this project's tendencies to longer form descriptions (if a comment is needed at all) punctuating sentences makes perfect sense to me.

daemon/integration-tests/src/actors.rs
```diff
@@ -27,32 +27,50 @@ pub trait Actor: Send {
...
         cmd.arg("-n");
+        // Add the local Neovim plugin tree to the plugin path.
+        let project_root = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        let project_root = project_root
+            .parent()
+            .ok_or("bad manifest")
+            .unwrap()
+            .parent()
+            .ok_or("bad manifest")
+            .unwrap()
```
blinry: Is there a reason you didn't use `expect()` here?

```suggestion
            .parent()
            .expect("Unable to find parent of CARGO_MANIFEST_DIR")
            .parent()
            .expect("Unable to find grandparent of CARGO_MANIFEST_DIR")
```

daemon/integration-tests/src/actors.rs
```diff
@@ -27,32 +27,50 @@ pub trait Actor: Send {
...
-        cmd.arg("-i").arg("NONE");
-        // Disable Swap file, to prevent CI failures related to them.
+        cmd.arg("--headless");
+        cmd.arg("--embed");
+        // Use factory defaults, disable user config, plugins, and shada.
+        cmd.arg("--clean");
+        // Also disable fallback from shada to swap files.
         cmd.arg("-n");
+        // Add the local Neovim plugin tree to the plugin path.
+        let project_root = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
```
blinry: We could call this variable something like `integration_tests_dir`, as it's not the project root yet.

blinry (approved):
Have two non-blocking nits. Feel free to optionally pick up my suggestions & merge! \o/

daemon/integration-tests/src/actors.rs
alerque: Fair enough. I was actually working on this in #543 where it *is* the root and then porting to here.

daemon/integration-tests/src/actors.rs
alerque: Okay okay okay this was a dirty hack, but glossing over it with more verbose panic messages isn't the fix. I already found a better fix that is in the #543 draft already, I'll just drop it here for good measure while that cooks.

daemon/integration-tests/src/actors.rs
alerque: This is now `workspace_root` and it actually *is* the workspace root, not a temporary value that needs tweaking to get to the root.

PR #545, opened 2026-05-07
Fixup e2e tests to use test build's binary, not $PATH

### LINKED ISSUE #562: Can't set `--username` containing spaces via CLI
**Describe the bug**

```
$  tt share --username "foo bar"
error: unexpected argument 'bar' found
```

**Expected behavior**

This should work. I'm very surprised that it doesn't. Why doesn't clap leave the quoted argument alone?

**Versions:**
- `teamtype --version`: teamtype 0.9.2-dev (60d270426 2026-05-27)

alerque:
Follow up to #544 (and #535 before that) to use the binaries built for this test run, not whatever is in the user's `$PATH` at the time of the test run. This helps local tests to be isolated from the user's environment.

Per [this comment](https://github.com/teamtype/teamtype/pull/544#issuecomment-4401614773) this is split into a draft PR because it uses nightly features and I want to research what our options are for avoiding that.

**Self-review checklist**

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -25,6 +25,18 @@ set unstable
 
 profile := "dev"
 
+# With positional arguments enabled, we can pass all the arguments to the bash
+# shell in a way that will get expanded to the original 'word' breakdown. However,
+# when we do this blindly in all cases and the job's positional arguments happen
+# to be empty the shell decides we must have wanted a placeholder for an empty
+# string argument — a construct that is invalid for many of our commands. The
+# solution is to decide up front whether we have any positional arguments at all
+# and then either not pass anything or pass them in a way that will get expanded
+# properly. As a caveat we can't use this workaround for nested jobs that pass
+# positional arguments to other jobs since one layer of quoting is lost, but we
+# don't need to because none of those happen to use spaces in arguments anyway.
+maybe-pass(args) := if args != "" { '"$@"' } else { "" }
+
 [group('check')]
 [parallel]
 check *ARGS: (check-cargo ARGS) check-typos
@@ -123,12 +135,13 @@ perfect: check lint test fuzz
 #     alias nvim='just --justfile ~/path/to/teamtype/Justfile nvim'
 #
 # Run Neovim with the plug-in for testing (can be used from outside the project).
+[script]
 [no-cd]
 nvim *ARGS: build-test
     {{ nvim }} --clean \
         --cmd {{ quote("let &runtimepath=\"" + justfile_directory() + "/nvim-plugin,\" . &runtimepath") }} \
         --cmd 'runtime plugin/teamtype.lua' \
-        {{ ARGS }}
+        {{ maybe-pass(ARGS) }}
 
 # This task will build (if necessary) and run the Teamtype CLI via the development version from this repository.
 # This is especially useful for manual testing and can be used from anywhere by invoking the Justfile externally,
@@ -137,6 +150,7 @@ nvim *ARGS: build-test
 #     alias teamtype='just --justfile ~/path/to/teamtype/Justfile teamtype'
 #
 # Build and run Teamtype for testing (can be used from outside the project).
+[script]
 [no-cd]
 teamtype *ARGS: build-test
-    $TEAMTYPE_BINARY {{ ARGS }}
+    $TEAMTYPE_BINARY {{ maybe-pass(ARGS) }}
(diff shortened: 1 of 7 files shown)


alerque:
Now stacked on #543 since we're going ahead with the workspace shuffle, and also #552 because building this with local developer tooling shortcuts just makes it easier to test.

alerque:
This works now without any nightly features. The solution isn't nearly as elegant as with the nightly features but it works. If you run `cargo test` in a way that includes any e2e tests it will panic. The error describes the missing binary. Invoking `cargo build` first will clear that up. Note the test won't refresh the build, just use it. For that use `just test` instead which knows it has to do a build first and hence will refresh itself on each run. Per above it's stacked on two other PRs so I'm not marking it ready for review yet.

alerque:
This now has a nice little developer convenience for development. Usage looks something like:

```console
# In any shell
$ alias nvim='just --justfile ~/path/to/teamtype/Justfile nvim'

# In whatever directory you want to test sharing with Neovim
$ nvim ...
```

The result will be that a debug binary will be built (or refreshed as needed) from the current Rust code and a clean Neovim instance will be fired up with just the latest development version of the plugin that should use just that debug binary, not whatever is in your system $PATH.

alerque:
Besides fixing up the e2e tests to not be dependent on the user's $PATH, this also now adds two convenience jobs to the Justfile. Here is how I use them:

```zsh
alias teamtype='just --justfile ~/projects/teamtype/teamtype/Justfile run'
alias nvim='just --justfile ~/projects/teamtype/teamtype/Justfile nvim'
```

Now I can in any directory and (as long as those shell aliases are active) I can run `teamtype` and `nvim` as I would normally invoke them in a shell but get the test versions of the binary and plugin (including building them if necessary).

Justfile
```diff
@@ -96,27 +105,38 @@ lint-rust:
...
+
+# This task will build (if necessary) and run the Teamtype CLI via the development version from this repository.
+# This is especially useful for manual testing and can be used from anywhere by invoking the Justfile externally,
+# e.g. with an alias such as:
+#
+#     alias teamtype='just --justfile ~/path/to/teamtype/Justfile run'
+#
+# Build and run Teamtype for testing (can be used from outside the project).
+[no-cd]
+run *ARGS: build-test
```
alerque:
```suggestion
teamtype *ARGS: build-test
```

This job could potentially mirror the main CLI name. I was thinking about how it stands in for `cargo run` but with  being able to be used outside of the project and in conjunction with the Neovim plugin development, but it also stands in for the CLI binary so that name might make sense too.

alerque:
Amended to address #562. Whether or not that issue was actually a case of `alias tt='just --justfile ~/projects/teamtype/teamtype/Justfile run'` or not, the Justfile job exhibits the same argument quoting problem.

zormit (approved):

alerque:
Needs proper fix for when to expand arguments and when not to, c.f #562.

alerque:
Still not right...

alerque:
This is one reason (of many) I like Zed shell! It handles argument and array expansion so much more gracefully by default and has more robust options for when the defaults aren't what you want.

Zed shell would have been:

```Justfile
teamtype *ARGS:
  teamtype $@
```

Bash doesn't retain the word boundaries when expanding this way, so quoted things in the input args come out split. Bash's solution is the expand inside quotes, then the inner quoting is retained and the shell then strips off one layer of quoting anyway:

```Justfile
teamtype *ARGS:
  teamtype "$@"
```

But that falls very flat in the case the positional arguments array is empty, because then Bash decides you must have meant a literal empty string and sends `teamtype ""`, which is bogus. Zed shell would correctly expand to no arguments at all, but for bash we have to get fancy and decide whether the positional arguments is empty or not before we decide to expand it or not:

```Justfile
maybe-pass(args) := if args != "" { '"$@"' } else { "" }

teamtype *ARGS:
    teamtype {{ maybe-pass(ARGS) }}
```

Now bash either gets `"$@"` that will expand to *something* or `` (without any quotes) which will expand to nothing instead of no an empty string.

I'm including the workaround to make this cooperate with `bash` for this project, and will continue to be even happier with insistence on using `zsh` for my own projects ;-)

zormit:
I'd be okay pivoting this to `zsh` to make it nicer to read, but as you observed elsewhere, it's default on a macOS, so @blinry has to make the call :)

Thanks for fixing it eventually!

PR #546, opened 2026-05-08
Fix socket creation for additional use case where share --directory is deeply nested

alerque:
PR #537 did not go quite far enough fixing #462. We caught the issue for socket creation from share/join when in the target directory and for the client even when using an extrenal `--directory`, but I did not catch the case of using `--directory` with the share subcommand.

This should catch that by just changing the whole CLI context –which should be good for other reasons too– but also adds a workaround directly to the call sites so that future use as a linked library will also function and not mess up the parent thread's location in the process.

**Self-review checklist**

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -191,10 +191,6 @@ fn parse_share_config(command: Commands, directory: PathBuf) -> AppConfig {
 }
 
 async fn run_client(directory: PathBuf) -> Result<()> {
-    // See comment in editor, but the TL;DR is that referencing a socket node from a deeply
-    // nested or overly verbose path will fail on some platforms. By *changing* into the
-    // target directory first we enable the use of relative paths in socket creation calls.
-    env::set_current_dir(&directory)?;
     let jsonrpc_forwarder = jsonrpc_forwarder::UnixJSONRPCForwarder {};
     jsonrpc_forwarder
         .connection(&directory)
--- a/daemon/src/daemon.rs
+++ b/daemon/src/daemon.rs
@@ -958,15 +958,15 @@ impl Daemon {
 
         let document_handle = DocumentActorHandle::new(&app_config, init, is_host, persist);
 
+        let base_dir = &app_config.base_dir;
+
         // Start socket listener.
-        let socket_path = app_config
-            .base_dir
+        let socket_path = base_dir
             .join(config::CONFIG_DIR)
             .join(config::DEFAULT_SOCKET_NAME);
         editor::spawn_socket_listener(&socket_path, document_handle.clone())?;
 
         // Start file watcher.
-        let base_dir = app_config.base_dir.clone();
         spawn_file_watcher(&app_config, document_handle.clone());
 
         if persist {
@@ -976,7 +976,7 @@ impl Daemon {
 
         // Start connection manager.
         let connection_manager =
-            peer::ConnectionManager::new(&app_config, document_handle.clone(), &base_dir)
+            peer::ConnectionManager::new(&app_config, document_handle.clone(), base_dir)
                 .await
                 .expect("Failed to start connection manager");
         let address = connection_manager.secret_address();
--- a/daemon/src/editor.rs
+++ b/daemon/src/editor.rs
@@ -128,8 +128,19 @@ pub fn spawn_socket_listener(
     // The std library function used to create sockets requires a path shorter than SUN_LEN, but the
     // length that matters is only the segment it is asked to handle. If passed an absolute path
     // here we can potentially be run in a path that exceeds the maximum (~100 chars). Passing it a
-    // relative path effectively sidesteps this limitation.
+    // relative path effectively sidesteps this limitation. Stripping the leading path segments will
+    // result in a relative path that won't have a long cumbersome prefix that fails safety checks.
+    // The extra song and dance to change into the parent directory first is not needed by our CLI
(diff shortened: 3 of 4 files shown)


cargaji:
@alerque, don't know if you appreciate this feedback, but this fix makes teamtype work in my manual testing. TeXstudio throws another error, I'll throw the logs here, but I think its related to an incorrect use of the client from their side (I'll raise it with them):

```
Collab Server Output: "Using the Git username 'MY_GIT_NAME' as username, to display next to the cursors other people see.\nTeamtype uses the Git username as username by default.\nYou can set the configuration value `username` in your `.teamtype/config` to override this username.\nYou can also use the flag `--username` when using the `share`/`join` subcommands.\n"
Collab Server Output: "Detected an existing socket '$DIR/.teamtype/socket'. There might be a daemon running already for this directory, or the previous one crashed. Do you want to continue? (y/N): "
Collab Server Output: "\n\tOne other person can use this to connect to you:\n\n\tteamtype join 6-cherokee-berserk\n\n"
Collab Server Output: "Editor #0 connected.\n"
Collab Server Output: "Error for JSON-RPC request: RequestError { id: None, error: EditorProtocolMessageError { code: -32700, message: \"Invalid request: EOF while parsing a string at line 1 column 58608\", data: None } }\n"
Collab Server Output: "Editor #0 disconnected.\n"
Collab Client Output: "{\"error\":{\"code\":-32700,\"data\":null,\"message\":\"Invalid request: EOF while parsing a string at line 1 column 58608\"},\"id\":null,\"jsonrpc\":\"2.0\"}"
Collaboration client finished with exit code 0 and status QProcess::NormalExit
"Error: JSON-RPC forwarder failed\n\nCaused by:\n    No such file or directory (os error 2)\n"
```

The file being edited is 56kB, but I assume that is not a limitation of teamtype. Thanks again for taking the time to fix this.

alerque:
Are you able to share the file in question manually and use a different editor to actually confirm if the problem is unique to TeXstudio? Does it work with other projects / files? If so can you raise an issue with the contents of the actual file? It isn't clear to me from that log which app is actually having a problem.

daemon/src/main.rs
```diff
@@ -62,6 +62,13 @@ async fn main() -> Result<()> {
     setup_teamtype_directory(&directory, temporary_directory.as_ref())
         .context("Failed to find .teamtype/ directory")?;
 
+    // Change into the project base directory before we do anything else. We mostly use absolute
+    // paths and most filesystem IO operations are handled by the sandbox, but having the process
+    // running in the shared project base directory helps avoid other potential problems.
+    // See comment about SUN_LEN in editor.rs, but the TL;DR is that referencing a socket node
+    // from a deeply nested or overly verbose path will fail on some platforms.
+    env::set_current_dir(&directory)?;
```
blinry: Could we solve this problem "at the root" for both daemon and client?
I think what you did for the daemon is exactly the right call – to make it impossible to call the code "in a wrong way". So could we also add a "briefly change into the socket's directory" wrapper to `jsonrpc_forwarder.rs`, as, well?

I think if we do, we could drop this here. I'd prefer that for simplicity reasons.

blinry (changes requested):
I confirmed the `--directory` problem, and that this PR solves that! :) Left a wish as a comment.

daemon/src/main.rs
alerque: Did I not already do that? I thought I did... did I botch the commit or something? Looking.

I do think keeping the `cd` in the CLI's main function will be the right approach for other reasons even though it "doubles" up the fix for this particular case. Eventually I think we should be using relative paths anyway since we're supposed to be sandboxed in a project.

alerque: Yikes, this branch is not my last iteration of this fix. Not only did I make this fix in the JSON/RPC socket creation point, but I changed the comment a bit too. Reflog here I come.

alerque: I don't know what I was smoking when. I think I probably copied the fix once, then didn't get it staged and then got distracted rewriting the comments in another branch and then with the changes in "two" places I spaced the fact that there should be "three"!

alerque: > I'd prefer that for simplicity reasons.

I think the tradeoff is simplicity vs. correctness. Yes it works now without being where we should be, but we aren't where we *should* be. Ideally I'd like to move this blanket fix into the sandboxing as a sort of init phase for *all* public API surface (as we process the config and deduce where are share directory is, `cd` into it and everything after that doesn't even need the directory passed around, it can just assume it is where it belongs—and that it *cannot* do anything elsewhere). At that point the CLI wouldn't need to care, but since the entry points to the API don't yet do that in my mind it makes sense to start shaping the assumptions in that direction.

I you disagree still I'll pull it out since the fix will work either way and we don't have tests that prove the correctness one way or the other.

blinry: I view this with another goal in mind: In the future, the daemon could share multiple projects at the same time – if we "properly" daemonize it. In that world, I'd only want to solve our socket problem at the "innermost" level like we did here.

In any case, keeping code around that's not required is something I think I'd like to avoid in most cases. I removed this "cautionary directory" change from the commit, and rewrote the commit message a bit to fit. (Is such an approach fine with you, if it's quick touch-ups before merging?)

blinry (approved):

daemon/src/main.rs
alerque: Works for me.

PR #550, opened 2026-05-14
Update status of known advisories

alerque:
The recent features we added using new Iroh dependencies surfaced a couple more advisories. These shouldn't be a security risk for us, and will be resolved when we can eventually bump Iroh.

Although I do have `cargo-deny` checks locally I wasn't paying attention to to them when #523 merged, and they are not yet in CI either. Given that they were introduced to our SBOM in that PR, these should have been added there along with the lock file bump, but we overlooked that. Of course I had to bump into these on another branch where I have the tooling set to run automatically. Eventually this will be checked without having to remember specifically! Thankfully nothing serious for our security model.

**Self-review checklist**

### DIFF
--- a/daemon/deny.toml
+++ b/daemon/deny.toml
@@ -24,7 +24,8 @@ ignore = [
     "RUSTSEC-2026-0119", # cpu exhaustion: hickory-proto (via iroh)
 ]
 
-# Not exhaustive, just whitelist the current dependency tree
+# This is not an exhaustive list of licenses we could accept, it is just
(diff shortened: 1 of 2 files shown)


blinry (approved):
I confirmed that, after this PR, `cargo deny check` is happy.

(Is there a way to check that we don't have "too many" ignored advisories, which are no longer present, by the way?)

alerque:
Yes, it will actually throw a warning when an ignored advisory is no longer relevant so you can see when a dependency upgrade takes care of it.

PR #551, opened 2026-05-15
Move code into appropriate crates

alerque:
Stacked on #543.

* Library crates SHALL NOT directly read/write to STDIO streams.
* Bin crates SHALL NOT package API surface useful to other possible UIs.

This touches on issues I started running into when working on log messages in #541. Neither that PR nor this (yet) finishes extricating the user-facing messaging from the library, but these are steps in the right direction.

**Self-review checklist**


alerque:
This can be marked ready for review as soon as #543. The only reason I'm not doing so now is the GH review tooling doesn't make any sense with the stacked commits from the other branch showing up here.

crates/e2e-tests/tests/fuzzer.rs
```diff
@@ -44,6 +44,10 @@ fn initialize_directory() -> (TempDir, PathBuf, PathBuf) {
     (dir, dir_path.to_path_buf(), file)
 }
 
+fn prompt_bool_dummy(_: &str) -> Result<bool> {
```
blinry, line 47: I think this function should carry the "no to everything" property in its name: `prompt_bool_always_false`?

crates/teamtype/src/editor.rs
```diff
@@ -100,6 +99,7 @@ pub(crate) fn strip_current_dir(path: &Path) -> PathBuf {
 pub fn spawn_socket_listener(
     socket_path: &Path,
     document_handle: DocumentActorHandle,
+    bool_prompter: &dyn Fn(&str) -> Result<bool>,
```
blinry, line 102: For consistency, I'd prefer to call this function the same everywhere. `prompt_bool` doesn't sound like a function to me, maybe `prompt_book_callback`? (Is there some convention in other Rust projects, maybe?)

crates/teamtype/src/main.rs
```diff
@@ -26,8 +27,6 @@ use tracing::{debug, info, warn};
 use self::cli::{Cli, Commands, ShareJoinFlags};
 
 mod cli;
-mod jsonrpc_forwarder;
```
blinry, line 29: I'm not sure I agree with this move. To me, the JSON-RPC forwarder doesn't feel like "library material". Would you expect other projects want to make use of it? Also, doesn't it break your first rule ("Library crates shouldn't deal with stdin/out")?

Maybe a bigger question, but because I don't remember super well, @zormit: What was the reason to split Teamtype into a binary crate and a library crate in the first place? Was it because it's not possible to write integration tests for binary crates?

And an even bigger one: Do we even want to publish the crate? In my mind, so far, all code in this repo was only there to support our binary, and we wouldn't have breaking changes to the library crate bump the major version. Maybe worthwhile to continue this discussion on Zulip…

blinry (changes requested):
I don't love giving the Daemon this additional argument, but you're right to point out the unclear separation of concerns! So I can accept iterating towards a cleaner interface step by step!

It's interesting that you tend to "build future capacity" much more than we did before, @alerque! I think so far I would only have made this change as soon as a second "user" of the prompting function existed. But also here, I'm not against this.

alerque:
> I think so far I would only have made this change as soon as a second "user" of the prompting function existed.

On the side I'm working on code for two other users: a TUI and Lua bindings. I also have a weather eye out for other possible bindings and compilation to WASM. These things are popping up as blockers to make those things feasible.

*Edit:* Additionally I think it's worth noting that none of this is **just** stubs or places to hang future work. The current code is scrambled and having `println!` and direct STDIO stream read/writes all mixed through the library source as well as overloading the logging macros for UI output is chaotic and at some points outright wrong. Besides tripping up our own fuzzer and it is current making it hard to differentiate errors from info logging from actual CLI output. These changes and tho follow up ones as well are all to get things where the actually belong and the functions we actually have work better / more correctly. That adding additional UIs will be easier to do after these refactorings is a secondary benefit. Right now trying to add anything on (like a TUI) quickly starts setting off foot-guns.

crates/teamtype/src/main.rs
alerque: > I'm not sure I agree with this move. To me, the JSON-RPC forwarder doesn't feel like "library material". Would you expect other projects want to make use of it?

Yes. If you have an optional TUI, GUI, or headless system daemon interface for running the listener, it should have access to all the things the CLI does without having to duplicate it or reaching into the binary module.

> Also, doesn't it break your first rule ("Library crates shouldn't deal with stdin/out")?

No. All the STDIN/OUT will come out of both this bit of module and all the other library ones, see the follow up PR #541 which does just that.

> Maybe a bigger question, but because I don't remember super well, @zormit: What was the reason to split Teamtype into a binary crate and a library crate in the first place? Was it because it's not possible to write integration tests for binary crates?

It makes perfect sense actually.

> And an even bigger one: Do we even want to publish the crate? In my mind, so far, all code in this repo was only there to support our binary, and we wouldn't have breaking changes to the library crate bump the major version. Maybe worthwhile to continue this discussion on Zulip…

Yes of course we want to publish it. Perhaps under a different version scheme, but "all this code is only to support our CLI binary" is upside down, the CLI binary is only there as one possible UI for people to access the underlying functionality. In Cargo's headspace the CLI is just the Clap argument processing + STDIN handling + Terminal escape handling (log colors) + whatever else is unique to the CLI as a wrapper around your core functionality. The TUI mode will have a different set of interactions with the user but expose the saying underlying functionality, just presented differently. The same goes for editors that link the library to offer a more integrated UX rather than calling the CLI behind the scenes.

I do agree what things are marked as public exports needs overhauling and probably greatly reducing. You might have noticed I've already started doing this for functions and types I've been touching: marking them as private or `pub(crate)` wherever possible so we don't expose more API surface than we need to and burden ourselves with the headaches of not being able to change it.

crates/teamtype/src/editor.rs
alerque: Consider this a placeholder to refactor the scope before I fixed the actual API. A follow up that sets up some actual UI hooks is in #541 already. Once the cheese was moved around to get things in the right crates (this PR) it was a lot easier to drop in a more ergonomic API. This function signature ends up looking like this:

```rust
pub fn spawn_socket_listener(
    socket_path: &Path,
    ui: &Interface,
) -> Result<()>;
```

crates/e2e-tests/tests/fuzzer.rs
alerque: See #541, this gets refactored into a trait object and will need to follow the same function naming for both the real (CLI) and dummy (fuzzer) implementations.

crates/teamtype/src/main.rs
alerque: > we wouldn't have breaking changes to the library crate bump the major version

Why not? Because the current CLI version number is actually representative of a protocol version that isn't represented in the protocol itself? And because currently the binary and library are versioned in sync?

blinry (approved):

crates/teamtype/src/main.rs
blinry: (We discussed in-person that while the jsonrpc-forwarder move does not fulfill the promise of "library shall not touch stdin/stdout *right now*, that's the long-term ambition.)

PR #552, opened 2026-05-15
Add developer-facing local tooling

alerque:
Stacked on #543.

Still in the creative experimentation phase.

Changelog and release workflow bits not yet in play.

For the Justfile, start with just `just` (or `just list`) to see a list of possible jobs. For my own personal projects I tend to lump a lot of stuff together (one lint task that does all the linting, one format task that does all the formatting, etc.) but when other people are involved I finde they often want to be able to run things on a more granular level or in different increments than I do, so this is on the more granular side. For example you can run `just lint` to do all the linting, but it is split into `just lint-rust` and `just lint-lua` and `just lint-license` and more under the hood, so you can see those jobs avoilable too.

**Self-review checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,23 +7,23 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 
 # 0.9.2 (unreleased)
 
-Fix issue blocking directories that are deeply nested (or have very long winded names) from being shared.
+Fix issue blocking directories with deep nesting (or very long winded names) from being shared.
 
 Handle POSIX specific process signalling without assuming everybody is running a Unix platform.
 
-Fix CLI thread handling to correctly listen for SIGINT (<Ctrl>-C) and SIGTERM (typically invoked with `kill`) while the main daemon or client loop is running.
+Fix CLI thread handling to correctly listen for `SIGINT` (<kbd>Ctrl</kbd>-<kbd>C</kbd>) and `SIGTERM` (typically generated with `kill`) while the main daemon or client loop is running.
 
 Add support for JSON-RPC IDs in string to be compliant with the [JSON-RPC specification](https://www.jsonrpc.org/specification).
 
-Add support for `.teamtypeignore` file to exclude files and directories from synchronization. The syntax is similar to `.gitignore`, supporting file names, directories, and glob patterns.
+Add support for a `.teamtypeignore` file to exclude files and directories from synchronization. The syntax is similar to `.gitignore`, supporting file names, directories, and glob patterns.
 
 Add support for custom Iroh relay servers, using the `--iroh-relay` flag, and for custom Iroh DNS discovery servers, using `--iroh-dns-domain` and `--iroh-pkarr-relay`. See the "Running your own relays" page in the documentation for how to run all these relays yourself!
 
 ## Notes for package maintainers
 
-### Renamed the Linux binaries
+### Renamed binary release assets
 
-We've renamed the Linux binaries
+We've renamed the prebuilt Linux binaries
 - `teamtype-x86_64-unknown-linux-musl` to `teamtype-x86_64-linux-static`
 - `teamtype-aarch64-unknown-linux-musl` to `teamtype-aarch64-linux-static`
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,3 +90,19 @@ too_many_lines = "allow"
 [workspace.lints.rust]
 unsafe_code = "forbid"
 unused_qualifications = "warn"
+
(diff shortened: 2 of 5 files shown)


alerque:
Also a note on the `Justfile`: I'm having it take the extra step of checking for and expecting *all* af the local tools it might us up front so if you don't have something (e.g. `cargo deny` or `luacheck` or whatever) it will error right out of the gate and let you know what you're missing. However it doesn't have to be setup that way, you can also just save the errors for when you actually try to run something. That would enable partial use (e.g. somebody who wanted to use some of the local tooling but not all of it).

Given that using `just` at all is completely optional and people can run the individual commands too, and also considering this is developer-only tooling not something an end user is expected to have or use, I've always felt this was a reasonable compromise.

blinry:
Can you show an example of what it would look like to check for the existence of a tool inside of a "target"? I didn't *love* that I needed to install all tools (including zsh :O, would it hurt much to use bash or sh?) just to get the list of available targets just now. Maybe enabling a partial use case would indeed be nice, but I'm assuming it comes with a downside?

alerque:
> Can you show an example of what it would look like to check for the existence of a tool inside of a "target"?

It isn't quite that simple, but I'll look into it. The main options are `which()` and `require()` which play the same role on a global scope, one failing to an empty string and the other producing an error. There isn't something like a per-job requirement, but I'll see if something like it can be rigged up.

> I didn't _love_ that I needed to install all tools [...] just to get the list of available targets just now.

Fair enough. Per my [previous comment](https://github.com/teamtype/teamtype/pull/552#issuecomment-4463196691) this was adapted from my own workflows where I do prefer being stopped in my tracks if I'm on a machine that doesn't have the right tooling for a project, but I can understand wanting to use things piecemeal and start slow too.

>  (including zsh :O, would it hurt much to use bash or sh?)

Not much, just the depth of my familiarity with Zed shell scripting. I can adapt it for Bash. I thought Zsh was available by default on macOS though so that's a bit of a surprise to me.

> Maybe enabling a partial use case would indeed be nice, but I'm assuming it comes with a downside?

Just more cumbersome to setup that way.

alerque:
> Can you show an example of what it would look like to check for the existence of a tool inside of a "target"? I didn't _love_ that I needed to install all tools (including zsh :O, would it hurt much to use bash or sh?) just to get the list of available targets just now.

I just had a go at this. Since there isn't a builtin function for this approach it requires a bit of ad-hoc noise in the Justfile code which I think feels cluttered. But it does work. The obvious gain is that you no longer have to have all the possible developer tools installed to run *any* developer jobs using `just`, just the ones it is going to use for that job. The downside is the clutter. Obviously my first pick was to just require everything up front, but I can live with this too if you guys are more comfortable with that. Remember all of this is optional anyway, everything can be run by hand too.

zormit:
> I thought Zsh was available by default on macOS though so that's a bit of a surprise to me.

I'm the one on macOS right now, blinry is on NixOS iirc

alerque:
I gave this a once-over per our conversation today. It is back to the "require all the dev tooling up front" approach. Also I ditched Zed shell and made the default shell environment Bash.

One detail I didn't mention is that in the event you do want to run some jobs without all the tooling, you can skip the checks. For example a rust dev without Lua tooling might use `just --set luacheck false --set stylua false lint-rust`. Not super convenient but might be useful for some situations. It's more useful to me to be able to swap out tooling to e.g. use a test version of a tool I might `just --set luacheck "lua ~/projects/lunarmadules/luacheck/bin/luacheck.lua" lint-lua`. Perhaps that's only useful to me (Did you know I was the maintainer for Luacheck?) but it can be used for any of these values and will be useful for testing editor integrations.

blinry:
Seems neat, and works as expected for me. If we find any issues, we can keep iterating.

I wrote a Nix flake for myself to give me all the required tools, we can think about putting it into the project later. Maybe at that point we can also put a sentence about the Justfile into CONTRIBUTING.md.

```nix
# SPDX-FileCopyrightText: NONE
#
# SPDX-License-Identifier: CC0-1.0

{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
  };

  outputs = {nixpkgs, ...}: let
    system = "x86_64-linux";
    pkgs = nixpkgs.legacyPackages.${system};
  in {
    devShells.${system}.default = pkgs.mkShell {
      packages = with pkgs; [
        cargo-deny
        just
        luaPackages.luacheck
        stylua
        reuse
        stylua
        typos
      ];
    };
  };
}
```

blinry (approved):

PR #553, opened 2026-05-16
Exit main loop when daemon error is propogated

alerque:
This is subtle but egregious. In working with signal handling we neglected to check that other abort conditions are handled correctly. The `bail!` macro and other error propogation techniques are used in quite a few places that bubble errors that SHOULD go all the way to the main loop. Well they are going to the main loop, but the way Tokio is handling them it is delaying the resolution of the daemon's Result<> until it absolutely has to.

By preserving the unwrapped() daemon handle instead of the wrapped version we force any possible errors to be propogated into the main loop sooner rather than later.

No changelog necessary since this just makes the things we already fixed since the last release work like they should.

**Self-review checklist**

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -79,7 +79,7 @@ async fn main() -> Result<()> {
             let share_config = parse_share_config(cli.command, directory.clone());
             run_daemon(share_config, init).await
         }
-    };
+    }?;
 
     trap_shutdown().await;
 


blinry:
Good catch!

Did  you find a way of testing this? I tried a bunch of things to crash the daemon, but failed! :D

I'll merge this, in any case!

alerque:
> Did you find a way of testing this?

I discovered this while I was working on unraveling logging vs. use facing messaging (see #541). I was deliberately *trying* to bump into warning and error messages to see how the formatting looked, especially when asking for user feedback. You can trigger this bug by trying to launch a share in a directory where you already have an active share daemon. It throws a warning and is supposed to die, but it didn't.

We don't have any test harness at all that test things like exit conditions of the CLI. We need one but I was trying to do things like get the crates workspace fixed before I worked on what harness to use for those sort of tests.

PR #555, opened 2026-05-20
Fix fuzzer to play nice with macOS temp dir being a symlink

### LINKED ISSUE #178: Fix fuzzer on macOS
We suspended it from the CI because we couldn't find out why it's hanging forever. We can observe the same effect locally. Maybe try bisecting?

alerque:
Closes #178.

Untested because I don't have an Apple device to test on, but the issue report now includes details which make sense.

@zormit ball is in your court to test.

**Self-review checklist**

### DIFF
--- a/.github/workflows/general.yml
+++ b/.github/workflows/general.yml
@@ -57,7 +57,7 @@ jobs:
         run: cargo test --locked --no-default-features -p teamtype
       - name: Run default e2e tests
         run: cargo test --locked --no-default-features -p e2e-tests
-      - if: ${{ matrix.os == 'ubuntu-latest' && matrix.nvim-version == 'v0.10.4' }}
+      - if: ${{ matrix.nvim-version == 'v0.11.2' }}
         name: Run e2e fuzzer
         run: cargo test --locked --no-default-features -p e2e-tests --test fuzzer
 
--- a/crates/e2e-tests/tests/fuzzer.rs
+++ b/crates/e2e-tests/tests/fuzzer.rs
@@ -29,16 +29,19 @@ async fn perform_random_edits(actor: &mut (impl Actor + ?Sized)) {
     }
 }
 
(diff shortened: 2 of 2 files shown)


alerque:
Enabling CI for this pretty much proves the macOS blocker was resolved, @zormit I presume your local results will be the same but of course you can confirm.

We don't know how *stable* CI will be for this, but it can't be any less stable than the other tests so I suggest not factoring that into consideration at the moment.

Obviously this is stacked on #543 so that needs to be merged first, but the 2 commits unique to this PR are ready for review.

zormit:
nice one :)
and seeing your fix, this somehow seems so obvious! 🥹 
we had put in some time to fix it on the daemon level, I believe...
but just resolving the path early, in the fuzzer, I think that didn't occur to us.

I can confirm it working. We have to wait for #543 to be merged, as it's stacked on that one.

alerque:
@zormit If you mark this PR review request as approved I can take care of it –including any necessary rebasing– after the other PR lands.

alerque:
> we had put in some time to fix it on the daemon level, I believe...

The CLI/daemon expects to be in the project directory and sandboxed to stay there. The sandbox code doesn't put any effort into detecting if path A and path B both canonicalize to the same thing before kicking you out, nor do I necessarily think it should. By the time we sandbox the thread it *should* be settled on a config that it is going to stick to for the lifetime of the thread. The fuzzer was initializing the daemon one way, then passing it messages about file to work on that was initialized a different way.

It is still possible the daemon and sandbox path handling should be completely reworked. I have a nagging feeling that always using absolute paths is fundamentally a bad habit and we need to default all the tooling to use all relative paths (with the sandbox doing a check that normalizing the relative paths doesn't end up outside of the project base dir) but I haven't had time to mess around with what that might entail to implement or what implications it could have security wise.

zormit (approved):

PR #558, opened 2026-05-27
Update security advisories for workspace

alerque:
When I re-arranged everything into one proper workspace, I neglected to recheck the current advisories. It turns out the shuffle brought a couple changes. Because the e2e tests with Neovim and fuzzing came into scope, running the same advisory checks against them brings a new license (a new-old licensse with a deprecated SPDX format, but that isn't our fault) and a new advisory. Additionally one other advisory is no longer being hit, and this seems to be related to how the Cargo resolver worked out which patch version of `rand` to use. With the e2e tests being in the same workspace even though we don't have them using the same version it has resolved the usages to both an older and a newer version and resolved one advisory for us.

**Self-review checklist**


blinry:
Thanks for detailing the changed advisories! I confirmed that with this PR, `just perfect` runs through cleanly.

PR #560, opened 2026-05-29
Trim down the library API surface

alerque:
This is a logical follow up and stacked on ~~#551, #541, #603, #611, and #601~~.

The split between binary and library crates and the exposed public API surface was clearly not considered while this was developed. Not knocking the work—it's just time for a refactor so multiple contexts can be supported. As of this posting I'd say I'm about half way there removing public exports that clearly belong, then we can start considering what the ergonomics are for the bits that are left.

image

( BEFORE → AFTER )

In normal circumstances taking away published APIs would be considered breaking, but since I don't think we have any consumers yet and the publication was somewhat accidental I don't think we must identify it as such *per se*.

**Self-review checklist**

### DIFF
--- a/crates/e2e-tests/tests/fuzzer.rs
+++ b/crates/e2e-tests/tests/fuzzer.rs
@@ -12,7 +12,7 @@ use e2e_tests::actors::{Actor, Neovim};
 use futures::future::join_all;
 use pretty_assertions::assert_eq;
 use rand::RngExt;
-use teamtype::config::{self, AppConfig};
+use teamtype::config::{self, Config};
 use teamtype::daemon::{Daemon, TEST_FILE_PATH};
 use teamtype::logging;
 use teamtype::sandbox;
@@ -85,21 +85,21 @@ async fn main() -> Result<()> {
     let (_handle2, dir2, file2) = initialize_directory();
 
     // Set up the actors.
-    let mut app_config = AppConfig::default();
-    app_config.base_dir = dir1;
-    let daemon = Daemon::new(app_config, true, false, ui).await?;
+    let mut config = Config::default();
+    config.base_dir = dir1;
+    let daemon = Daemon::new(config, true, false, ui).await?;
 
     // Wait until iroh's DNS discovery (hopefully) works.
     sleep(Duration::from_millis(1000)).await;
 
     let nvim = Neovim::new(Some(file1)).await;
 
-    let mut app_config2 = AppConfig::default();
-    app_config2.base_dir = dir2;
-    app_config2.peer = Some(config::Peer::SecretAddress(
+    let mut config2 = Config::default();
+    config2.base_dir = dir2;
+    config2.peer = Some(config::Peer::SecretAddress(
         daemon.secret_address().to_string(),
     ));
-    let peer = Daemon::new(app_config2, false, false, ui).await?;
+    let peer = Daemon::new(config2, false, false, ui).await?;
 
     // Wait until file2 appears.
     while !file2.exists() {
--- a/crates/e2e-tests/tests/fuzzer.rs
+++ b/crates/e2e-tests/tests/fuzzer.rs
@@ -96,7 +96,9 @@ async fn main() -> Result<()> {
 
     let mut app_config2 = AppConfig::default();
     app_config2.base_dir = dir2;
-    app_config2.peer = Some(config::Peer::SecretAddress(daemon.address.clone()));
+    app_config2.peer = Some(config::Peer::SecretAddress(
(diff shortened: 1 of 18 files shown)


alerque:
This PR is still subject to scope creep, but that's because I'm working down other rabbit holes and then pushing stuff up to this branch when the scope makes sense here. All my other working branches at this point are using this as a starting branch as it saves a lot of duplicate work and rebasing.

Any time this can get reviewed it probably should because the sooner any of this lands of main the less friction there will be to work on other fixes. In fact if it's easier to split this up and cherry pick commits as they get reviewed we can do that too.

crates/teamtype/src/lib.rs
```diff
@@ -6,19 +6,26 @@
...
 pub mod types;
-pub mod watcher;
-pub mod wormhole;
+
+// Used by e2e test crate, but not officially public
+#[doc(hidden)]
+pub mod document;
+#[doc(hidden)]
+pub mod editor_protocol;
+
```
blinry, line 24: (The plan is to un-publish these modules in the long run.)

crates/teamtype/src/setup.rs
```diff
@@ -18,12 +18,23 @@ use crate::types::UserInterface;
...
 
 pub fn setup_teamtype_directory(
-    directory: &Path,
-    temporary_directory: Option<&TempDir>,
+    directory: Option<&Path>,
     ui: &UserInterface,
-) -> Result<()> {
-    if has_ethersync_directory(directory) {
-        let old_directory = directory.join(LEGACY_CONFIG_DIR);
+) -> Result<(PathBuf, Option<TempDir>)> {
```
blinry, line 23: (@alerque plans to refactor this further, see https://github.com/alerque/teamtype/commit/33ac007f4786fff763fad8d62cee8475cc50676d)

crates/e2e-tests/tests/fuzzer.rs
```diff
@@ -12,7 +12,7 @@ use e2e_tests::actors::{Actor, Neovim};
 use futures::future::join_all;
 use pretty_assertions::assert_eq;
 use rand::RngExt;
-use teamtype::config::{self, AppConfig};
+use teamtype::config::{self, Config};
```
blinry, line 15: (Note: We were probably inspired by `magic_wormhole::AppConfig`.)

blinry:
(Caleb and I reviewed the commits in this PR that are not part of #541 together, and fixed any objections I had.)

alerque:
This was a pretty major rebase after already reviewing all the relevant commits with @blinry, but [this compare link](https://github.com/teamtype/teamtype/compare/29668a1ed1709cbccf7e66ed64b94a00c9a98fef..d3067805f691b48aaaace5a76d819d1bd61c02d2) shows just the rebase differences and it's relatively easy to skim that and see that it is all stuff from the base branch(es).

blinry:
Eek, merging #601 (which automatically deleted the branch) made this PR close again. :( Still don't understand why.

To fix, I:

- re-created the `change-default-verbosity` branch
- re-opened this PR
- changed the base branch to `main`

alerque:
The PR closure happened because this was not part of the stack the other PR series we just worked through was. It was a branch I created before the stack thing existed and it is from a branch on *my fork*, not from the canonical repository. The stacking feature that updates base branches on base branch merges instead of closing them only works for PRs of branches in the canonical repository and marked as part of the relevant stack.

alerque:
I just rebased this to strip out the legacy ethersync handling that was originally *moved around* with this refactoring since the PR that *removed* them merged first. This should be substantially unchanged since @blinry and I reviewed it commit by commit once already except for the necessary rebasing, but I haven't stepped through it very closely (e.g. with clippy on each commit) since the rebasing either so it's possible some rebase mess slipped in. Other than checking for that nonsense, I consider this ready to go.

Do note that some things (e.g. the temp dir handling) are changed in this PR a little bit, but a *much improved* version of those changes appears in another branch (originally #571, now #620). Don't worry to much about the final shape of the directory handling, the point here is to get stuff into the right crates to enable unpublishning a bunch of things.

alerque:
Converted to draft because something *is* off in the rebasing, something gets moved and not marked as `use` in a matching commit. I'll fix that and the lints before a review is needed.

alerque:
Okay I cleaned up some artifacts from bad rebasing (such as copy-edited log messages that were moved from one module to another, then editing in the base branch but the rebase didn't get the message updates copied yet) and fixed the lint issues commit-by-commit. It should be good to review and merge again now.

PR #563, opened 2026-06-01
Adjust Lua formatting rules and restyle

alerque:
As noted [in this PR review comment](https://github.com/teamtype/teamtype/pull/427#discussion_r3330723372) the Lua identifier casing is slightly inconsistent. The first commit here fixes that. The remaining camelCase instates are table keys used externally by LSP and so forth.

After that this takes a quick dive into possible bikeshed territory. Please don't knock it till you try it!

1. Use a space between the identify and the signature when defining functions, and no space when calling them.

    Lua is very tolerant of whitespace variations. I have a strong preference for this particular combination in relation to function definitions and it is consistent with some other language defaults.

    ```lua
    -- Like this:
    function foo (arg) end
    foo(6)

    -- Never like this or other variants:
    function foo(arg) end
    foo (6)
    ```

2. Use 3-space indents.

   Yes three is an odd number, and yes very few language consider this normal. But Lua kind of does. Many large Lua code bases including many of the most used Lua ecosystem tools have standardized on this: `luarocks`, `luacheck`, `busted`, etc.

   Many rationals for this preference could be invented. I personally thought it was fantastically unorthodox when I was first exposed, then in quickly grew on me for Lua specifically. I don't think it works well for most languages, but it suites Lua nicely. Lua tends to have more deeply nested control structures than many languages; so 4 space indents tend to waste a lot of horizontal space but (IMO) 2 space indents don't give enough visual separation. 3 might be odd but it turns out to be a sweet spot. Many of the language constructs end up with a very satisfactory alignment:

    ```lua
    -- comment leaders are 3 spaces wide
    -- if conditionals with continuations like and/or line up with the initial expression
    if a == 1
       and b == 2
       or a == b
       then

       -- string concatenation operators are 3 wide
       local foo = "bob"
          .. a
          .. b

       -- Deep nesting doesn't eat too much space but is more distinct than 2 space:
       local bar = func({
          foo = "bar",
          nested = {
             foo = "bar",
             nested = {
                foo = "bar",
                nested = {
                   foo = "bar",
                },
             },
          },
       })

       -- ending nested control structures cascade
       for i in ipairs(_) do
          for j in ipairs(_)  do
             print(i, j)
          end
       end

    end
    ```

Obviously these last two changes are low priority and entirely based on my preferences, but having worked with a lot of Lua code I do suggest trying them on on for size.

**Self-review checklist**

### DIFF
--- a/.github/workflows/nvim-plugin.yml
+++ b/.github/workflows/nvim-plugin.yml
@@ -25,7 +25,7 @@ jobs:
       - uses: JohnnyMorganz/stylua-action@v5
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          version: 0.20.0
+          version: 2.5.2
           args: --config-path nvim-plugin/.stylua.toml --check nvim-plugin
       - name: Luacheck Linting
         uses: lunarmodules/luacheck@v1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
     -   id: cargo-check
         args: []
 -   repo: https://github.com/JohnnyMorganz/StyLua
-    rev: v2.1.0
+    rev: v2.5.2
     hooks:
     -   id: stylua
         args: ['--config-path', 'nvim-plugin/.stylua.toml', '--check', 'nvim-plugin']
--- a/nvim-plugin/.stylua.toml
+++ b/nvim-plugin/.stylua.toml
@@ -3,5 +3,6 @@
 # SPDX-License-Identifier: CC0-1.0
 
 indent_type = "Spaces"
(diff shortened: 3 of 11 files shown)


blinry:
Lol, *three* spaces! :D That's delightfully weird.

I agree that it looks kinda nice, but I'd say let's not cater to your taste for now, @alerque, if that's okay. For me, the whole point of automatic formatters is not having to discuss styling so much.

If anything, we could try to go closer to how Neovim styles their source code, which seems to be two spaces, to allow easier code sharing. But my suggestion would be to stick to what we have right now, and to drop the style setting changes in this PR. I'll move it a bit in my heart, though; maybe we can come back to it at some point.

PR #567, opened 2026-06-03
Avoid messing with Cargo's default handling of profiles

alerque:
Per our session today, here are the profile changes I suggest. The commits have more detailed reasoning for each.

As an example in regard to stripping, Arch Linux builds all packages without explicitly stripping and various compiler default options set to *not* strip binaries. Then at the last stage of the build, it runs `split` on the binaries that would be installed, but stashes the debug symbols that get split out. If there are any, it will package them in a separate package (`teamtype-debug`). This goes in a separate repository and most people will never download or use it.

But there is magic. It is possible to enable the debug repositories and enable some hooks so that if user runs `gdb teamtype` or triggers various other debug scenarios, the package that actually has the debug symbols will be downloaded and installed first, then the debugger will combine the stripped symbols with the original binary so you can actually run a debugger on the final stripped release artifact and still get meaningful debug information.

What will happen in the current scenario is that the `teamtype` package will be stripped (expected) and the `teamtype-debug` package will be empty (because the symbols were removed before the packaging even saw the artifact) and people will not be able to debug it. Here is what I see with Arch's current package:

Debian, Gentoo, and many other distros have similar systems. Not all of them package symbols or wire things up so they are easy to get, but even the ones that just discard the debug information at least typically strip binaries so the project build system doesn't need to worry about it.

**Self-review checklist**

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,6 +39,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: taiki-e/upload-rust-binary-action@v1
+        env:
+          CARGO_PROFILE_RELEASE_CODEGEN_UNITS: 1
+          CARGO_PROFILE_RELEASE_LTO: fat
+          CARGO_PROFILE_RELEASE_STRIP: symbols
         with:
           bin: teamtype
           include: README.md,LICENSE.md
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,15 +38,6 @@ default-features = false
 [workspace.dependencies.teamtype]
 path = "crates/teamtype"
 
-# Stripping symbols (not debuginfo) reduces binary size by ~1.5x.
-# This is the default profile for `cargo install`:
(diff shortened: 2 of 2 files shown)


blinry:
Looks reasonable to me. Deleting stuff always makes me happy. :)

Did you find a way to test the added environment variables in the CI? Otherwise, we can check whether this succeeds when doing the next release.

alerque:
> Did you find a way to test the added environment variables in the CI? Otherwise, we can check whether this succeeds when doing the next release.

Negative. Those only run on release and I just applied the settings that made sense according to my reading of the upstream docs. I don't have a way to test those artifacts, and doing so requires a complete re-hash of CI.

...which I think we need anyway. For example we could be building the release artifacts any time things get merged to *main* (probably *not* on every PR or branch, just *main*) and have them available as CI artifacts. Then the only difference for actual releases would be attaching those artifacts to the releases, not generating all new ones using a different system.

You can see in the Windows PR #534 I already started posting debug artifacts from CI runs which is a step in that direction.

PR #568, opened 2026-06-03
Baby sit the version of stuff running in CI

alerque:
This isn't the complete overhaul the CI jobs needs, but at least it should get rid of the constant warnings about EOL actions we're running that have old Node versions and make GHA very unhappy.

**Self-review checklist**

### DIFF
--- a/.github/workflows/general.yml
+++ b/.github/workflows/general.yml
@@ -32,7 +32,7 @@ jobs:
         nvim-version: [v0.9.5, v0.10.4, v0.11.2]
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
       - name: Set up Rust toolchain
         run: rustup toolchain install stable --profile minimal
       - uses: Swatinem/rust-cache@v2
@@ -62,7 +62,7 @@ jobs:
     name: Check Rust formatting
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
       - run: cargo fmt --check
 
   clippy:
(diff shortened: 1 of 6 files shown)


alerque:
With this PR I think we'd be able to enable the CI checks from the 'general' workflow as required to pass for all PRs again. c.f. https://teamtype.zulipchat.com/#narrow/channel/563011-continuous-integration/topic/adapting.20the.20branch.20ruleset

.github/workflows/general.yml
```diff
@@ -7,16 +7,8 @@ name: General tests
...
-    paths:
-      - 'crates/**'
-      - 'nvim-plugin/**'
-      - '.github/workflows/general.yml'
+      - main
   pull_request:
-    paths:
-      - 'crates/**'
-      - 'nvim-plugin/**'
-      - '.github/workflows/general.yml'
```
blinry, line 19: I'm not sure I like dropping these paths, but I get your point about making green "general" tests mandatory again. But I'm open to trying this, and see if it's annoying having to wait several minutes for documentation fixes… if it is, maybe we can put these back in, or find another way of only testing the relevant workflows?

alerque: Yes. And by "find another way" it should be finding a way that actually works, because this does not. I don't know what's wrong with the expressions, but they don't run sometimes when they should and run sometimes that would seem to have been precluded.

One thing to mitigate the couple minutes wait is to enable automerge so that somebody can choose to go ahead and merge *pending* passing checks/reviews. Especially if the required PR state is CI green plus 1 approving review, having automerge enabled would keep the wait time for CI from being much of a friction point.

blinry:
Merging, will try to look into making the required repo settings another time :)

PR #569, opened 2026-06-04
Retry instead of hang indefinitely on failure to reach Magic Wormhole

alerque:
I was hanging out on a bad internet connection today and trying to work on user messaging issues. I ran across this `error!` which procedes to hang indefinitely and make it impossible to connect or add peers. I don't understand why this should be a show stopper. If the daemon needs restarting we should be actually raising an `Err()` here. I was going to fix the function signature to bubble up an error, but that didn't seem right either since it might disconnect peers that *are* connected even if the network segment with the MW relay is actually down.

In the end I don't see any realy reason not to retry here, at least not at a conservative rate.

**Self-review checklist**

### DIFF
--- a/crates/teamtype/src/wormhole.rs
+++ b/crates/teamtype/src/wormhole.rs
@@ -8,7 +8,9 @@ use std::{borrow::Cow, str::FromStr, time::Duration};
 use anyhow::Result;
 use magic_wormhole::{AppConfig, AppID, Code, MailboxConnection, Wormhole, transfer};
 use tokio::time::sleep;
-use tracing::{error, info, warn};
+use tracing::{info, warn};
+
+const NETWORK_RETRY: Duration = Duration::from_secs(300);
 
 pub async fn put_secret_address_into_wormhole(address: &str, magic_wormhole_relay: Option<String>) {
     let payload: Vec<u8> = address.into();
@@ -17,10 +19,12 @@ pub async fn put_secret_address_into_wormhole(address: &str, magic_wormhole_rela
     tokio::spawn(async move {
         loop {
             let Ok(mailbox_connection) = MailboxConnection::create(config.clone(), 2).await else {
-                error!(
-                    "Failed to share join code via Magic Wormhole. Restart Teamtype to try again."
+                warn!(
+                    "Failed to register a new join code via Magic Wormhole. Automatic retry in {:?}. Peers who joined before can still re-connect without a code.",
+                    NETWORK_RETRY
                 );
-                return;
+                sleep(NETWORK_RETRY).await;
+                continue;
             };
             let code = mailbox_connection.code().clone();
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 
 # 0.9.2 (unreleased)
 
+Continue trying to contact Magic Wormhole on a 5 minute cycle in the event join codes fail to register.
+
 Add status widget to the Neovim plugin with basic information to be used in statuslines.
 
 Fix issue blocking directories with deep nesting (or very long winded names) from being shared.


blinry:
I don't think it's a showstopper – `put_secret_address_into_wormhole` spawns its own Tokio task. I tried it on main and can still connect to the peer, even when using Magic wormhole fails because of a `--magic-wormhole-relay https://example.com` flag.

I like being explicit that this did not work here, and giving the user guidance on how to try again better than waiting several minutes. In the future, a nicer TUI that gives the user the option to generate new join codes on keypress will also allow the user to try again without having to restart the daemon.

Would it be an improvement in your point of view if we extended the warning to say "Peers who already know the secret address can still connect, however."?

alerque:
Of course future alternative UIs (TUI / editor embeded, etc.) might be given deeper access to this with a direct way to ask for single codes even before previous ones are used. But that's out of scope for here.

By "hanging" I didn't mean that this was blocking the main thread and the process was hanging, I just meant that this particular thread / loop was never getting rerun. I would argue the "explicit" advice to the use is counter productive since restarting the daemon would disconnect any current peers and require them to restart too when the can potentially recover from a temporary failure to reach the wormhole server. Personally I consider it obvious that if they were impatient they could kill the process and try again immediately, but I wouldn't want to encourage that as the way they *have to* handle things when we can try to recover gracefully for such a relatively low cost.

Adding the message clarifying the difference between new joiners using a code vs. existing joiners that know the peer might make sense to add here. I'll look at amending to improve the message.

alerque:
How does that revised verbiage strike you?

crates/teamtype/src/wormhole.rs
```diff
@@ -17,10 +19,12 @@ pub async fn put_secret_address_into_wormhole(address: &str, magic_wormhole_rela
     tokio::spawn(async move {
         loop {
             let Ok(mailbox_connection) = MailboxConnection::create(config.clone(), 2).await else {
-                error!(
-                    "Failed to share join code via Magic Wormhole. Restart Teamtype to try again."
+                warn!(
+                    "Failed to register a new join code via Magic Wormhole. Automatic retry in {:?}. Peers who already know the secret address can still connect.",
```
zormit: Somehow I think "know the secret address" is a little bit technical, but it generally seems to be a concept that users have to be taught: If you connected once, you can continue to connect.

I'm wondering if this version is more clear, but also maybe there's no need to overthink this, as it's a very low probability path:

```suggestion
                    "Failed to register a new join code via Magic Wormhole. Automatic retry in {:?}. Peers who joined before can still re-connect without a code.",
```

crates/teamtype/src/wormhole.rs
```diff
@@ -8,7 +8,9 @@ use std::{borrow::Cow, str::FromStr, time::Duration};
 use anyhow::Result;
 use magic_wormhole::{AppConfig, AppID, Code, MailboxConnection, Wormhole, transfer};
 use tokio::time::sleep;
-use tracing::{error, info, warn};
+use tracing::{info, warn};
+
+const NETWORK_RETRY: Duration = Duration::from_secs(300);
```
zormit, line 13: 5 minutes seems like a pretty long time, but maybe it doesn't matter if it's just about keeping the connection open for re-connections.

zormit (approved):
to me this looks good, considering the comments.

crates/teamtype/src/wormhole.rs
alerque: There is already a pretty long network timeout loop trying to get the connection through the fist time (30s? 60s?) and people won't be hitting this unless that has completely failed, so if the network didn't start working in 60s starting another connection attempt in the same or less time the first one ran didn't make sense to me.

alerque: The network retry loop runs for 30s not 60s, but I still assert that's a *long* time in network failure modes and should cover most very ephemeral conditions. Flooding the network with retry attempts as soon as handshake timeouts have failed is not productive; and 10× the active packet retry loops to start over from scratch seems like the about the right spot to me. We can adjust later if it feel wrong for any use case.

Issue #305, opened 2025-07-10
Nix tests on CI take over 10 minutes


blinry:
## Problem description

The Nix tests take over 10 minutes on the CI: https://github.com/ethersync/ethersync/actions/runs/16169633430

For comparison, the "general tests", thanks to caching with [rust-cache](https://github.com/Swatinem/rust-cache), take under 2 minutes: https://github.com/ethersync/ethersync/actions/runs/16169633438

This sometimes slows down our development process, as we want to wait for the CI to become green.

## Potential solution

@MangoIV do you maybe know a way of doing similar caching for Nix builds? Very often, the dependencies wouldn't need to be rebuilt, just the ethersync crate itself.

MangoIV:
yes - two main things 
- using a cache: you can click a free cache on cachix.org, they have some space for open source projects
- build more separately, i.e. at least a separate derivation for the dependencies and the project itself because dependencies change more rarely, there's also projects for building all of the crates separately like e.g. with Cargo.nix (https://nix-community.github.io/crate2nix/00_guides/31_auto_generating/)

MangoIV:
another potential solutions is to run the nix *build* only after merge on main, that way you can go back and fix it if something breaks but you're not blocked for merging the actual thing.

PR #570, opened 2026-06-06
Add Nix flake

alerque:
All the usual goodies like `nix run`, `nix develop`, `nix shell`, and such should work.

This is the first time I've used this particular combination of flake overlays for working with Rust. Most of my projects use autotools and my usual go-to for flakes is the *naersk* overlay. See for example the [flake for `decasify`](https://github.com/alerque/decasify/blob/master/flake.nix). I'm generally happy with that, but I ran across [this example here](https://log.woodweb.ca/articles/rust-flake/) that included the dev shells split for different Rust toolchain versions and I thought that might be useful so I decided to give it a spin. This flake is largely derived from there with some modifications of my own and a dev dependency list from a comment of @blinry that had a dependency flake.

A caveat on the LLM issue here, this actually does have a very small GenAI contribution. I beat my head against a nix language parse error for a long time and regular web searches for the problem were not getting me anywhere. The Nix compile errors were completely inscrutable to me and I decided to see if an LLM could find the cause. Somewhat to my surprise OpenCode w/ Big Pickle cut straight to the root problem. The issue wasn't my Nix syntax at all or and all the workarounds I'd been trying were way off. The issue is the TOML parser built into Nix (`builtins.fromTOML`) has some limitations and couldn't handle the otherwise valid inline table formatted over multiple lines. GAH! The resulting fix is technically LLM generated code but also the fix so trivial that once I knew what the problem is impossible to unsee and the only way to fix it without exactly copying the code would be to break the table out into a new section which is even more unnecessary for a single-entry table.

**Self-review checklist**

- [-] This PR does not contain content generated by LLMs, see our [generative AI policy](https://github.com/teamtype/teamtype/blob/main/CONTRIBUTING.md).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 
 # 0.9.2 (unreleased)
 
+Add a Nix Flake to the project so the repository can be run directly, used as a dependency, or just to supply a developer environment.
+
 Continue trying to contact Magic Wormhole on a 5 minute cycle in the event join codes fail to register.
 
 Add status widget to the Neovim plugin with basic information to be used in statuslines.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,9 +71,8 @@ unused_qualifications = "warn"
 
 [workspace.metadata.typos.default]
 locale = "en-us"
-extend-words = {
-    ot = "ot"
-}
+# Note: the TOML parser used in the Nix flake can't handle inline tables on multiple lines.
+extend-words = { ot = "ot" }
 extend-ignore-identifiers-re = [
     "2nd",
 ]
--- a/Justfile
+++ b/Justfile
@@ -6,6 +6,7 @@ cargo := require('cargo')
 cargo-deny := require('cargo-deny')
 just := just_executable()
 luacheck := require('luacheck')
+nix := require('nix')
 nvim := require('nvim')
 reuse := require('reuse')
 stylua := require('stylua')
@@ -68,12 +69,16 @@ build-test *ARGS:
 
 [group('format')]
 [parallel]
-format: format-lua format-rust
+format: format-lua format-nix format-rust
 
 [group('format')]
 [working-directory("nvim-plugin")]
 format-lua:
     {{ stylua }} --respect-ignores .
+
+[group('format')]
+format-nix:
+    {{ nix }} fmt flake.nix
 
 [group('format')]
 format-rust:
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -2,6 +2,6 @@ version = 1
 
 # Computer-generated files.
 [[annotations]]
-path = ["**/Cargo.lock", "**/package-lock.json", "**/initial_automerge_doc.bin"]
+path = ["**/Cargo.lock", "**/package-lock.json", "**/initial_automerge_doc.bin", "flake.lock"]
 SPDX-FileCopyrightText = "NONE"
 SPDX-License-Identifier = "CC0-1.0"
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,96 @@
+{
+  "nodes": {
+    "flake-parts": {
+      "inputs": {
+        "nixpkgs-lib": "nixpkgs-lib"
+      },
+      "locked": {
+        "lastModified": 1778716662,
+        "narHash": "sha256-m1Yf0wZ8j1OHjTc2UwHwyQRSnNeSgLJOd7q5Y45hzi4=",
+        "owner": "hercules-ci",
+        "repo": "flake-parts",
+        "rev": "f7c1a2d347e4c52d5fb8d10cb4d94b5884e546fb",
+        "type": "github"
+      },
+      "original": {
+        "owner": "hercules-ci",
+        "repo": "flake-parts",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1780243769,
+        "narHash": "sha256-x5UQuRsH3MqI0U9afaXSNqzTPSeZlRLvFAav2Ux1pNw=",
+        "owner": "nixos",
+        "repo": "nixpkgs",
+        "rev": "331800de5053fcebacf6813adb5db9c9dca22a0c",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nixos",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "nixpkgs-lib": {
+      "locked": {
+        "lastModified": 1777168982,
+        "narHash": "sha256-GOkGPcboWE9BmGCRMLX3worL4EMnsnG8MyKmXNeYuhQ=",
+        "owner": "nix-community",
+        "repo": "nixpkgs.lib",
+        "rev": "f5901329dade4a6ea039af1433fb087bd9c1fe14",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-community",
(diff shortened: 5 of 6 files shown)


alerque:
And yes, it was *after* working this up I discovered that there had previously been a Flake in this project that was [removed here](https://github.com/teamtype/teamtype/issues/305#issuecomment-3327399790).

I would argue the that removal was solving the wrong problem the wrong  way. The two reasons I see surfaced seem to be:

1) The nixpkgs repo picked it up as an upstream package, so people could install/run releases that way.
2) The related CI build job was taking way too long.

I would say that 1 represents the wrong problem: the main objective with the Flake isn't for end users to run stable tagged releases that way. The goal is developers and tinkerers being able to build and run any branch or commit they want in a reproducible way with known inputs. This is quite handy, but you can't use it unless the Flake has been committed prior to the commits and branches you want to test drive.

In addition to the convenience of being able to run, test, and cache arbitrary versions the Flake also facilitates developer tooling with the shell and develop environments such that the project can be easily worked on without needing to adjust the host system to add tooling or change versions of anything. Nobody is required to use it, and I'm not on NixOS and don't primarily use nix shells to provide anything, but even so still find them useful to have around.

I would say that 2 can be solved other ways: notably not needing to run a build in CI at all for it at all, and also the speed issue likely being a result of not having proper Nix store caching. I keep Nix Flakes around in basically all my projects and usually the build times can be made comparable with other build jobs with proper caching. In the event that doesn't pan out I'd just say don't run the job at all or at least make it non required so nobody was ever forced to wait for it. Realistically the speed of CI isn't our bottleneck anyway.

blinry:
I tried using this flake, activating the (default) dev shell, and running `just perfect`.

- This task failed:

```
$ just lint-format-rust 
/nix/store/39s3cxs9s1l0vqfr0n5xpzmdapn9hzib-rust-default-1.98.0-nightly-2026-06-06/bin/cargo +nightly fmt --check
error: no such command: `+nightly`
```

- `just lint-manifests` also currently fails with the `git2` crate (of course out-of-scope for this PR)

Can't get in a proper review yet, but some more thoughts:

- Your LLM usage seems to fall into the grey area that we didn't explicitly disallow. For me, if you used it to get "unstuck", and then used the trick you learned to solve your problem, this seems okay!
- The reason we deleted the Flake before was indeed motivated by it being packaged in nixpkgs. We just didn't need it anymore, and I feel if no-one uses something, I'd rather remove it than maintain it. Of course there might have been "unknown-to-us" users. For me, it'd indeed now be useful again to run the just tasks.
- We used to use the flake for tests and builds in the CI, but didn't try to set up proper caching, which made the CI times a hassle. This probably could be solved! I'd love to keep the CI times as short as possible though.
- I was surprised that the devshell defaults to nightly. I'd have expected default = stable. Do we really need to be that "bleeding edge"? When developing on nightly, at which point would we learn that the features that our clippy/LSP suggested we use in our code is not supported on stable?
- Could we easily also add the "aarch64-darwin" platform?

alerque:
> I tried using this flake, activating the (default) dev shell, and running `just perfect`.
> 
>     * This task failed:
> 
> 
> ```
> $ just lint-format-rust 
> /nix/store/39s3cxs9s1l0vqfr0n5xpzmdapn9hzib-rust-default-1.98.0-nightly-2026-06-06/bin/cargo +nightly fmt --check
> error: no such command: ``+nightly``
> ```

I'll look into that.

>     * `just lint-manifests` also currently fails with the `git2` crate (of course out-of-scope for this PR)

Yes, I noticed that working on the release tooling too, there are two new RUSTSEC advisories. They do go away if we bump libgit2 because patches were out quickly, but we should probably do that.

> Can't get in a proper review yet, but some more thoughts:
> 
>     * Your LLM usage seems to fall into the grey area that we didn't explicitly disallow. For me, if you used it to get "unstuck", and then used the trick you learned to solve your problem, this seems okay!

Back in the day a web search would probably have been more effective at solving this because more people posted their problems and solutions. These days so many people check with an LLM first half that knowledge never becomes public.

Also search engines have started being terrible at finding things that are public.

And even if search wouldn't have worked, StackOverflow would have been a resource that solved this is 5 minutes, but that's sadly a wasteland now.

>     * The reason we deleted the Flake before was indeed motivated by it being packaged in nixpkgs. We just didn't need it anymore, and I feel if no-one uses something, I'd rather remove it than maintain it. Of course there might have been "unknown-to-us" users. For me, it'd indeed now be useful again to run the just tasks.

Also it should be useful for regression testing. I'd like to get this is before we tag the next release just so that when testing the known-breaking Iroh stuff we can make sure we're warning properly and this will be an easy way to run old versions and new ones side by side.

>     * We used to use the flake for tests and builds in the CI, but didn't try to set up proper caching, which made the CI times a hassle. This probably could be solved! I'd love to keep the CI times as short as possible though.

Yes I agree.

>     * I was surprised that the devshell defaults to nightly. I'd have expected default = stable. Do we really need to be that "bleeding edge"? When developing on nightly, at which point would we learn that the features that our clippy/LSP suggested we use in our code is not supported on stable?

That was probably for the sake of the formatting rules we went with that are still experimental, but I'll double check.

>     * Could we easily also add the "aarch64-darwin" platform?

Sure. And soon to be Windows. I just didn't have it to test ;-)

alerque:
> I tried using this flake, activating the (default) dev shell, and running `just perfect`.

That's a bit of a catch 22. Allowing the Justfile to work with both hard coded toolchain versions and Rustup is not trivial. So instead of having the flake provide a nightly toolchain itself I just opened up the *dev* shell to use rustup there. That should ease the mix of mostly stable but for some dev jobs reaching into nightly that we have going on without making it complicated on other platforms outside of the Nix shell.

That also eliminates the need for the multiple possible dev shell variants.

> $ just lint-format-rust 

Works now.

>     * `just lint-manifests` also currently fails with the `git2` crate (of course out-of-scope for this PR)

I rebased this on #579 to address that.
 
>     * I was surprised that the devshell defaults to nightly. I'd have expected default = stable. Do we really need to be that "bleeding edge"? When developing on nightly, at which point would we learn that the features that our clippy/LSP suggested we use in our code is not supported on stable?

Removed per the first topic in this comment.

I think we're dependent on Clippy not trying to force something on us that doesn't work in our MSRV. Most of it's suggestions are things that will work even in older versions even though there were not lints about them at the time.

Besides for Clippy by default we're only asking it to pass stable rules. Only the format jobs are using nightly features that differ from stable output.
 
>     * Could we easily also add the "aarch64-darwin" platform?

Done.

alerque:
In order to work on release automation stuff I'm trying to use `nix run` as one way to test the version string generation. Hence I've had to stack this PR on #580 so that I can stack the release versioning stuff on it. The only commit relevant to this PR is the single "add nix flake" one.

blinry:
What do you think of defining the Nix formatter in the flake? What you've been using seems very close what `nixfmt` does. I added a commit that sets it, and adds a Just job for it. Feel free to tweak it if it seems suboptimal.

I verified that `just perfect` works now in the shell, and `nix build .` also works for me.

blinry (approved):

alerque:
> What do you think of defining the Nix formatter in the flake?

Great idea.

> What you've been using seems very close what `nixfmt` does.

Very close? It should have been the same as my usual incantation is to run `nix run nixpkgs#pkgs.nixfmt *.nix`. The only change I see in your diff a blank line you took out at the top of the flake, and I suspect that was a relic of an older version of the formatter, maybe back when in was in the RFC stage. I think the final approved version allows whitespace outside of the main block.

> I added a commit that sets it, and adds a Just job for it. Feel free to tweak it if it seems suboptimal.

Looks good to me! Putting it in the flake itself has the advantage of linking the formatting operation to the nix lock file so we'll all get the same version of that tooling too.

PR #571, opened 2026-06-07
Cleanup initialization order, use dedicated types

alerque:
Stacked on #541 and #560.

**Self-review checklist**

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -136,6 +136,9 @@ lint-manifests:
 lint-rust:
     {{ cargo }} clippy --all-targets --all-features
 
+doc:
+    {{ cargo }} doc --all-features --no-deps --open -p teamtype
+
 [group('test')]
 test *ARGS: (test-cargo ARGS)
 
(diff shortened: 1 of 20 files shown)


alerque:
I started trying to fix the STDIO handling vs. other stuff that we talked about in regard to the `run_client()` command moving from the bin crate to the lib crate, but ran into some other problems before that could be addressed. See the commit message for fe5ebfef.

Question for either of you: was there ever a use case for the `teamtype client` subcommand to work on a directory that had not previously been initialized with a teamtype directory and config? It's currently creating one and then dying anyway when it discovers there is no socket to read. I'm rather of the opinion that it should error earlier and just refuse to run at all if asked to start where it isn't going to be able to run rather than creating directories and files and then failing anyway. Does that make sense?

alerque:
Since this PR got incorrectly closed (thanks to terrible GitHub limitations) and the branch name was originally meant for some other task I couldn't get to for running into these issues first, I'm going to re-create this as a PR stack.

PR #575, opened 2026-06-16
Use custom word lists for join codes

### LINKED ISSUE #419: Have more fun whisper links
We're currently using the magic wormhole wordlist, which probably is based on the [PGP word list](https://en.wikipedia.org/wiki/PGP_word_list).  We would like to have something more cheerful, like the [words that screensy](https://github.com/screensy/screensy/blob/main/screensy-website/screensy.ts#L630-L635) is using :)

blinry:
Fixes #419.

To do:

- [x] Give the word lists another good read-through and filter out the worst ones

**Self-review checklist**

- [x] Conventional commit style


crates/teamtype/src/wormhole.rs
```diff
@@ -7,16 +7,32 @@ use std::{borrow::Cow, str::FromStr, time::Duration};
...
-            let Ok(mailbox_connection) = MailboxConnection::create(config.clone(), 2).await else {
+            let adjective = adjectives
+                .iter()
+                .choose(&mut OsRng)
+                .expect("Word list is not empty");
+            let noun = nouns
+                .iter()
+                .choose(&mut OsRng)
+                .expect("Word list is not empty");
+            let password = format!("{adjective}-{noun}");
```
alerque: Does the Wormhole side of this automatically prepend the random number? Is it using the number prefix like a *nonce*? I should probably go read some documentation...

blinry: Yup, Wormhole calls it "post box".

blinry:
Ready for review, I think! Some questions to ponder:

- Is `assets/` the best place for the word lists?
- Magic Wormhole's `Password` newtype is a very cool idea – but are we confident that our passwords will have [enough entropy](https://docs.rs/magic-wormhole/latest/src/magic_wormhole/core.rs.html#795)?
- Feel free to give the word lists themselves another read. We have room for removing more of them, if needed. And we can easily change them later.

zormit:
> Is assets/ the best place for the word lists?

I think it's alright. Magic Wormhole puts it directly parallel to the wordlist.rs that uses it, but in our case it would mean it's just in `src/`.

> Magic Wormhole's Password newtype is a very cool idea – but are we confident that our passwords will have [enough entropy](https://docs.rs/magic-wormhole/latest/src/magic_wormhole/core.rs.html#795)?

I'd be curious to run this code with our wordlist as input as well. But I think as we have more than 256 values in each, we should pass the 2^16 required entropy?

crates/teamtype/src/wormhole.rs
```diff
@@ -17,9 +18,26 @@ pub async fn put_secret_address_into_wormhole(address: &str, magic_wormhole_rela
...
+    let adjectives: Vec<&str> = include_str!("../assets/adjectives.txt").lines().collect();
+    let nouns: Vec<&str> = include_str!("../assets/nouns.txt").lines().collect();
+
     tokio::spawn(async move {
         loop {
-            let Ok(mailbox_connection) = MailboxConnection::create(config.clone(), 2).await else {
+            let adjective = adjectives
+                .iter()
+                .choose(&mut rand::rng())
+                .expect("Word list is not empty");
```
alerque, line 29:
```suggestion
                .expect("Word list cannot be empty");
```

This is a minor nit, but I find this wording confusing when reading the code. I understand the expectation that `.expect()` describes what should have happened instead of what did happen, but in context of the error message I think it is too easy to misread the result as the expectation is that the word list should have been empty and wasn't. I this copy edit still keeps the spirit of documenting what *should* be happening (the expectation) but with less possible confusion in the context of the error message.

It would be even better if these were `.context(...)?` instead of `.expect()`, but the necessary refactoring to bubble up errors is obviously out of scope for this PR.

crates/teamtype/src/wormhole.rs
```diff
@@ -17,9 +18,26 @@ pub async fn put_secret_address_into_wormhole(address: &str, magic_wormhole_rela
...
+            let noun = nouns
+                .iter()
+                .choose(&mut rand::rng())
+                .expect("Word list is not empty");
```
alerque, line 33:
```suggestion
                .expect("Word list cannot be empty");
```

crates/teamtype/src/wormhole.rs
```diff
@@ -17,9 +18,26 @@ pub async fn put_secret_address_into_wormhole(address: &str, magic_wormhole_rela
...
+            let password = format!("{adjective}-{noun}")
+                .parse()
+                .expect("Password is long enough and has enough entropy");
```
alerque, line 36:
```suggestion
                .expect("Password must be long enough and have enough entropy");
```

alerque (changes requested):
Overall this looks fine to me, just a couple panic message copy editing nits.

> Ready for review, I think! Some questions to ponder:
> 
>     * Is `assets/` the best place for the word lists?

Yes, I think that's a good pick.

>     * Magic Wormhole's `Password` newtype is a very cool idea – but are we confident that our passwords will have [enough entropy](https://docs.rs/magic-wormhole/latest/src/magic_wormhole/core.rs.html#795)?

Clearly the entropy check in that function won't be able to cover us. The entropy of even the shortest words in our list if they were character-wise strings instead of word-wise would be far higher than what we actually have.

That being noted, `259 × 333 = 86,274` which is more than `2^16 = 65,536`. So if that's the entropy requirement, we have enough input space.

The next question is if the random number generator is cryptographically secure and if we're using it right for that purpose ... and yes I think `rand::rng()` is setup specifically to be a simple invocation to achieve exactly that. It's widely trusted enough that I don't have concerns about it given where this piece fits in our trust model. If this played a more significant role than establishing one time peer discovery (e.g. it was used to encrypt data over the wire that could be decrypt later/offline) then `rand::ChaCha20Rng` might be a better fit, but think just the basic default is good here.

>     * Feel free to give the word lists themselves another read. We have room for removing more of them, if needed. And we can easily change them later.

I only did a cursory skim of the word lists, but like you said we can iterate on that easily. There are a few that *might* have possible bad pairings (notably body parts/articles of clothing) but nothing glaringly bad.

alerque:
I just rebased this to resolve the merge conflicts: [changing](https://github.com/teamtype/teamtype/compare/3da2577221ef7ae818868e7a2314219b52fdba24..5fab0b8b565deeb9736900cf7fcd8c6aa9dededa) `warn!` to `ui.warn()` and bringing the UI stuff into scope like it is on the default branch now.

alerque:
This will have an effect on #627 as well, which now proposes using the join code as a last resort fallback for the username when a client connects from a join code. Obviously that only works once, but it does make it really easy to identify otherwise un-configured peers who is who if you just passed out join codes to specific people.

PR #576, opened 2026-06-17
Use new Just (1.52) feature so list job doesn't require tooling

alerque:
When we were merging #552 we discussed the inconvenience of having to check for all the devevloper tooling just to run the `list` job that showed what you *could* do. One thing I missed at that point was that the only problem with `list` was that we had it setup as a custom job target to fiddle with the order and allow it to work when invoking `just` with no arguments. It turns out we could always have run `just --list` directly because that doesn't invoke any jobs and hence doesn't evaluate expressions at all, hence not tripping the `require()` functions.

With the recent Just 1.52 release, there is now a way to configure the list action as the default and there have been other cleanups such that I consider our list target to be unnecessary cruft.

I suggest letting this PR sit until @blinry and @zormit's platform of choice has Just 1.52 packaged and installed by default, then we can put this through. There is no urgency, but it's just a tiny bit tidier.

**Self-review checklist**

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -22,6 +22,8 @@ export TEAMTYPE_BINARY := justfile_directory() + "/target/debug/teamtype"
 set script-interpreter := ['bash', '-eu']
 set shell := ['bash', '-eu', '-c']
 
+set default-list
+set default-script
 set positional-arguments
 set unstable
 
@@ -50,11 +52,6 @@ check-cargo *ARGS:
 [group('check')]
 check-typos:
     {{ typos }}
-
-[default]
-[private]
-@list:
-    {{ just }} --list --unsorted
 
 [group('build')]
 build *ARGS:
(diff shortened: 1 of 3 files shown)


zormit:
I'm choosing nix on MacOS, which seems to have 1.55 now, and on 1.53 since [three weeks ago](https://github.com/NixOS/nixpkgs/blame/691bc285bc31189c3186c00b88b10ca829a884ef/pkgs/by-name/ju/just/package.nix).

```
$ just --version
just 1.55.1
```

blinry also should be on the same version via NixOS, if they're updating accordingly.
From my perspective we can merge this. WDYT @blinry?

zormit (approved):

alerque:
And with this merge I confirmed that GitHub is rebasing commits so they all get new commit IDs even when a fast forward is possible. It stripped my signed commits and made it's own in my name.

Good to know for the release automation considerations — and not so good that it does that.

blinry:
I'm also fine with this being merged. :)

PR #580, opened 2026-06-19
Touchup issues with alternate build environments

alerque:
In working on release tooling I discovered that we broke `cargo install --git <url_to_repo>` because of an assumption about Cargo setting workspace related env vars, and local execution of `cargo publish` has been broken for even longer (discovered with a --dry-run) because the readme needs to be part of the same crate (previously papered over because CI was handling this copy).

(Stacked on #579).

**Self-review checklist**


crates/teamtype/build.rs
```diff
@@ -57,21 +60,19 @@ fn instantiate_initial_automerge_doc() -> Result<()> {
 
 fn output_completions() -> Result<()> {
```
blinry, line 62: Just for my understanding: Why was this broken? This now seems to fetch the environment variable at runtime instead of compile time?

blinry (approved):

crates/teamtype/build.rs
alerque: Aaah, but runtime of what?

In this case it is the `build.rs`, so after the compile time for the build script is actually compile time for the actual binary & library crates — still during the overall "build process" if you will and notably not the runtime of teamtype itself.

This was broken because we're relying on a Cargo variable set in our workspace via the project's cargo config. This works fine for compiling from Git and even compiling from a `git archive` generated source tarball (e.g. what a distro might fetch from a tagged release). However it is actually broken for compiling from the crate sources as published to crates.io since that does not get built in a workspace, it gets it's own lockfile with all the workspace dependencies resolved.

This means that our builds have actually been broken since the workspace re-arranging, but in a fairly subtle way. It will actually *work* if you've even previously built the workspace from the Git sources, but *fail* on a system where the Rust build cache has not seen the workspace before.

```console
$ cargo install --locked --force --git https://github.com/teamtype/teamtype
    Updating git repository `https://github.com/teamtype/teamtype`
  Installing teamtype v0.9.2-dev (https://github.com/teamtype/teamtype#9a713aa7)
    Updating crates.io index
   Compiling noise-rust-crypto v0.6.2
   Compiling rcgen v0.13.2
   Compiling sha2 v0.11.0
   Compiling iroh-relay v0.35.0
   Compiling inotify v0.11.2
   Compiling tar v0.4.46
   Compiling derive_more-impl v2.1.1
   Compiling spake2 v0.4.0
   Compiling iroh v0.35.0
   Compiling rmp-serde v1.3.1
   Compiling serde_json v1.0.150
   Compiling flate2 v1.1.9
   Compiling stun_codec v0.3.5
   Compiling teamtype v0.9.2-dev (/home/caleb/.cargo/git/checkouts/teamtype-c3e511e5806a7bcc/9a713aa/crates/teamtype)
error: environment variable `CARGO_WORKSPACE_DIR` not defined at compile time
  --> crates/teamtype/build.rs:61:40
   |
61 |     let workspace_root = PathBuf::from(env!("CARGO_WORKSPACE_DIR"));
   |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: Cargo sets build script variables at run time. Use `std::env::var("CARGO_WORKSPACE_DIR")` instead

error: environment variable `CARGO_WORKSPACE_DIR` not defined at compile time
  --> crates/teamtype/build.rs:74:40
   |
74 |     let workspace_root = PathBuf::from(env!("CARGO_WORKSPACE_DIR"));
   |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: Cargo sets build script variables at run time. Use `std::env::var("CARGO_WORKSPACE_DIR")` instead

   Compiling libgit2-sys v0.18.5+1.9.4
   Compiling x509-parser v0.16.0
error: could not compile `teamtype` (build script) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
error: failed to compile `teamtype v0.9.2-dev (https://github.com/teamtype/teamtype#9a713aa7)`, intermediate artifacts can be found at `/home/caleb/.local/share/cargo`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_BUILD_BUILD_DIR` to that path.
```

Note the killer here is the assumption on the part of the build.rs script that `env!` will succeed. It's basically an `unwrap()` on the env var being available at build time. Most of the time that will be fine, but not *all* the time. Since this is a compile time check for the build script, we can't even avoid the panic by having logic that checks for the env var and doesn't call the code at all, using `env!` here will introduce a panic even if we had it in a dead code block.

By the way I'm pretty sure there is actually a Cargo bug lurking in here because the build.rs script should not be rerun from a cached build given the source environment changed and there were `env!` macros in it.

By using runtime `std::env` and `std::var_os` we're able to check whether the env var is set in our build environment and code around the two possible cases. If we don't have a workspace we con't use it as an output location for extra assets, so we just skip building them.

Since the "runtime" in question is still prior to the app being built and is the runtime of the build environment, we still get the right variable.

alerque: Here are some docs from the Rust lang reference suggesting this exact work around for a related problem scenario:

https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts

PR #584, opened 2026-06-23
feat(vscode-plugin): follow cursor

m10090:
**Self-review checklist**

### DIFF
--- a/vscode-plugin/src/cursor.ts
+++ b/vscode-plugin/src/cursor.ts
@@ -1,5 +1,6 @@
 // SPDX-FileCopyrightText: 2024 blinry <mail@blinry.org>
 // SPDX-FileCopyrightText: 2024 zormit <nt4u@kpvn.de>
+// SPDX-FileCopyrightText: 2026 Mohamed El tamawey <tammwy22@gmail.com>
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
@@ -24,6 +25,7 @@ interface RemoteCursor {
 }
 
 let cursors: Map<string, RemoteCursor[]> = new Map()
+let followedUser: string | undefined = undefined
 
 export function setCursor(userid: string, name: string, uri: vscode.Uri, selections: vscode.DecorationOptions[]) {
     let usersCursors = cursors.get(userid)
@@ -84,4 +86,49 @@ export function drawCursors(editor: vscode.TextEditor | undefined) {
             .map((cursor) => cursor.selection)
         editor.setDecorations(selectionDecorationType, allSelections)
     }
+}
+
+/// this function does nothing if no cursor to follow
+export async function syncFollowCursor() {
+    try {
+        if (followedUser === undefined) {
+            return
+        }
+        const cursorsToFollow = cursors.get(followedUser)
+        if (!cursorsToFollow || cursorsToFollow.length === 0) {
+            return
+        }
+        const cursorToFollow = cursorsToFollow.at(-1) as (typeof cursorsToFollow)[number] // follow last cursor like the neovim plugin
+        const textDocument = await vscode.workspace.openTextDocument(cursorToFollow.uri)
+        const editor = await vscode.window.showTextDocument(textDocument, {preview: false})
+        editor.selection = new vscode.Selection(
+            cursorToFollow.selection.range.start,
+            cursorToFollow.selection.range.end,
+        )
+        editor.revealRange(cursorToFollow.selection.range, vscode.TextEditorRevealType.InCenterIfOutsideViewport)
+    } catch {}
+}
+export async function followCursor() {
+    const items: vscode.QuickPickItem[] = [
+        {label: "$(circle-slash) Stop following", description: "__stop__", alwaysShow: true},
+    ]
+
+    cursors.forEach((remoteCursors, uid) => {
+        if (remoteCursors.length > 0) {
+            const label = followedUser === uid ? `$(link) ${remoteCursors[0].name}` : remoteCursors[0].name
+            items.push({label, description: uid})
+        }
+    })
+
+    const picked = await vscode.window.showQuickPick(items, {
+        placeHolder: "Follow which user?",
+    })
+
+    if (!picked) {
+        return
+    }
+
+    const useridToFollow = picked.description !== "__stop__" ? picked.description : undefined
+    followedUser = useridToFollow
+    await syncFollowCursor()
 }
--- a/vscode-plugin/src/extension.ts
+++ b/vscode-plugin/src/extension.ts
@@ -1,5 +1,6 @@
 // SPDX-FileCopyrightText: 2024 blinry <mail@blinry.org>
 // SPDX-FileCopyrightText: 2024 zormit <nt4u@kpvn.de>
+// SPDX-FileCopyrightText: 2026 Mohamed El tamawey <tammwy22@gmail.com>
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
@@ -10,7 +11,7 @@ import * as path from "path"
 import * as fs from "fs"
 var Mutex = require("async-mutex").Mutex
 
-import {setCursor, getCursorInfo, drawCursors} from "./cursor"
+import {setCursor, getCursorInfo, drawCursors, followCursor, syncFollowCursor} from "./cursor"
 
 function findMarkerDirectory(dir: string, marker: string) {
     if (fs.existsSync(path.join(dir, marker))) {
@@ -204,8 +205,8 @@ function openCurrentTextDocuments() {
     vscode.workspace.textDocuments.map(processUserOpen)
 }
 
-function documentForUri(uri: string): vscode.TextDocument | undefined {
-    return vscode.workspace.textDocuments.find((doc) => doc.uri.toString() === uri)
+function documentForUri(uri: string): Thenable<vscode.TextDocument | undefined> {
+    return vscode.workspace.openTextDocument(vscode.Uri.parse(uri))
 }
 
 // Apply given TextEdits to content of the TextDocument, and return the resulting content.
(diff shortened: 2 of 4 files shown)


m10090:
So you could say that this PR is currently a draft, until I know weather this method is the right way forward.

vscode-plugin/src/cursor.ts
```diff
@@ -3,7 +3,7 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-import * as vscode from "vscode"
+import * as vscode from "vscode";
```
alerque: I'm not saying it's wrong because I know very little about TypeScript, but this expression ending semicolon seems like an extensive style change mixed in throughout this PR. Can you comment on why? Is this a required change? What coding style guide are you following? Is there a code formatter we could use for consistency?

m10090: this isn't required.
Sorry is done by the formater.

blinry: We've been using [Prettier](https://prettier.io/), but haven't enforced it yet in the CI :)

vscode-plugin/src/cursor.ts
```diff
@@ -78,8 +81,52 @@ export function drawCursors(editor: vscode.TextEditor | undefined) {
...
+            .filter((cursor) => cursor.uri.toString() === uri.toString())
+            .map((cursor) => cursor.selection);
+        editor.setDecorations(selectionDecorationType, allSelections);
     }
-}
\ No newline at end of file
+}
+
+/// this function does nothing if no cursor to follow
+export async function syncFollowCursor() {
```
alerque, line 92: I'm not very familiar with TS, but the naming of these functions is pretty confusing. `followCursor()` is okay, but how that differs from `syncFollowCursor` is not clear from the function name or code comments, and both are exported symbols. Adding to the confusion is that both have *async* traits, so you have an *async* function with `sync` in the name.

m10090: okay, I try to think of a better name but couldn't find a great one.
can you help me :)

vscode-plugin/src/cursor.ts
```diff
@@ -78,8 +81,52 @@ export function drawCursors(editor: vscode.TextEditor | undefined) {
         let uri = editor.document.uri;
         let allSelections = Array.from(cursors.values())
             .flat()
-            .filter(cursor => cursor.uri.toString() === uri.toString())
-            .map(cursor => cursor.selection);
-        editor.setDecorations(selectionDecorationType, allSelections)
+            .filter((cursor) => cursor.uri.toString() === uri.toString())
+            .map((cursor) => cursor.selection);
```
alerque: Why did these singleton variable names need isolating? Was there a type change here or is this just a coding style change or what?

m10090: just a coding style

vscode-plugin/src/extension.ts
```diff
@@ -204,8 +204,8 @@ function openCurrentTextDocuments() {
     vscode.workspace.textDocuments.map(processUserOpen)
 }
 
-function documentForUri(uri: string): vscode.TextDocument | undefined {
-    return vscode.workspace.textDocuments.find((doc) => doc.uri.toString() === uri)
+function documentForUri(uri: string): Thenable<vscode.TextDocument | undefined> {
+    return vscode.workspace.openTextDocument(vscode.Uri.parse(uri))
 }
```
m10090, line 211: So what do you think about it?
here the editor will need to load all the files in memory - I think -, at least the ones edited by other users.

alerque: This seems reasonable, but only when a follow mode is enabled. If not only the files the local users should open and *if* they have remote cursors display those. But yes if a follow mode is enabled I would expect all files that have remote cursors to be loaded into memory and/or visibly shown as open.

vscode-plugin/src/cursor.ts
alerque: Okay. I have no opinion on TS coding styles, but lets separate the functional changes from the style ones in two commits, and maybe even two PRs. The functional changes should be a single commit following the coding style of the file at the commit point. If you want to restyle, do that in an earlier commit and get the style all sorted out and consistent, then layer the functional changes in another commit.

For style changes in general I think it would be good if you/we followed some consistent style guide, and even better if that could be formatted deterministically by some tool. Is there a standard tool for doing that with TS?

vscode-plugin/src/cursor.ts
alerque: At the very least spell out `syncronize` if it is a reference to an action and not a trait. But that doesn't really clarify what is going on. Can you describe in prose what the difference between `followCursor()` and `syncFollowCursor()` is?

vscode-plugin/src/cursor.ts
blinry: As mentioned elsewhere, please just reformat with [Prettier](https://prettier.io/). :)

alerque: I see [this comment](https://github.com/teamtype/teamtype/pull/584#discussion_r3458830835) has info about the formatter previously used, so closing this thread to discuss it in one place.

vscode-plugin/src/cursor.ts
alerque: If the output of `prettier` is deterministic lets keep following what it does. If the current code doesn't match it's output, lets add a `style(vscode): Reformat TypeScript with prettier` commit that cleans it up first, then commit the functional changes from this PR as a second commit on top of that and keep following the same style to isolate just the relevant changes.

I have no dog in the fight for what TS style guide actually gets used so if changes to that are desired we could consider than separately, but for the sake of a feature PR like this one style shouldn't be mixed in except to make sure the starting point is consistent with the existing style expectations.

alerque: I just posted #585 which will add a Just task for formatting & linting typescript. It also reformats the existing code which is a little out of sync. If you 

1) Run `prettier --write **.ts` yourself now and amend this commit to have the deterministic formatting applied here, then
2) Rebase this to be stacked on that PR (or on main after it merges) the changes in this PR should be a lot cleaner with only the relevant bits showing in diffs.

alerque: I just did both those steps for you. You can `git pull --rebase` from your branch to get the changes locally. [The diff for this PR](https://github.com/teamtype/teamtype/pull/584/changes) now looks like an actual feature addition :smile.

vscode-plugin/src/extension.ts
m10090: > open and if they have remote cursors display those.

No we need to include it in all cases. 
the issue is that this extension takes the RPC and convert it to typescript types.
when converting the cursor Pos it converts it from the remote type of the daemon to the `vscode.Selection` type.
the issue is if the file isn't loaded in memory you can't construct `vscode.Selection` as it needs the files to be loading in vscode's memory.
```
    const document = await documentForUri(uri)

    try {
        let selections: vscode.DecorationOptions[] = []
        if (document) {
            selections = cursor.ranges
                .map((r) => teamtypeRangeToVSCodeRange(document, r))
                .map(vsCodeRangeToSelection)
                .map((s) => {
                    return {
                        range: s,
                        hoverMessage: cursor.name,
                    }
                })
        }
        setCursor(cursor.userid, cursor.name || "anonymous", vscode.Uri.parse(uri), selections)
```
so the users that are editing files other than the ones you are opening can't be see.
like if you call `Teamtype: List other cursors positions` you will not see the cursors outside your view.
also you can't follow what you can't see.

m10090: it could be fixed by lazily construct this `selection` only when it is needed.
but this will demand rewrite to the cursor logic in vscode's extention

vscode-plugin/src/cursor.ts
m10090: `followCursor` you choice which cursor to follow. 
`syncFollowCursor` moves your cursor to the cursor you followed.

alerque: Maybe `activateFollowCursor()` and `moveCursorToFollowed()` or something like that. Just brainstorming, but something that doesn't conflict with the same of the overall feature and clarifies what exactly the functions do seems in order.

vscode-plugin/src/extension.ts
alerque: > the issue is if the file isn't loaded in memory you can't construct `vscode.Selection` as it needs the files to be loading in vscode's memory.

I'll deffer on the internals of how VSCode works as that's not my wheelhouse, but just to clarify: is there a reason we have to construct the `vscode.Selection` at all in that case? Couldn't the construction process be aborted and the RPC message skipped with an early break of some kind if both:

1. The file is not one the user has open, and
1. The follow cursor feature is not enabled?

m10090: So if I am editing a file that is not open in your vscode editor, I am basically invisible for you.
Therefore, you can't follow me because in vscode point of view i am not her.
You can try this open 1 vscode and another editor, move the cursor in some file that isn't opened by vscode.
Now in vscode show cursor info you can't see this cursor.
I don't know if that is a bug or a feature

alerque: Of course if follow cursor is enabled we need to open the relevant files. My initial comment was about what to do when follow mode is *not* enabled, and for that case opening files the user doesn't have open doesn't seem to make sense. If we're talking about when follow mode is enabled I would fully expect it to just open all files that have remote cursors. Does that make sense?

m10090: I am sorry if you didn't understand me 
but this photo should make it clear.

Screenshot 2026-06-26 at 2 00 24 PM

this is after using `Teamtype: List other cursor positions`

alerque:
Something is off on the settings for this repository. Somehow I believe GitHub now supports stacked PRs in such a way that when the target branch merges, PRs based on it should switch their target branch to where the other branch merged. This closure is incorrect and a side effect of GitHub *not* doing that bit. We'll look into it and get this open again.

blinry:
Let's move the discussion to https://teamtype.zulipchat.com/#narrow/channel/557642-general/topic/Stacked.20PRs

PR #586, opened 2026-06-24
docs: Draft CHANGELOG for 0.9.2 release

blinry:
This is based on what we had collected in the CHANGELOG, plus some things I hand-plucked from the git-cliff output.


CHANGELOG.md
```diff
@@ -5,38 +5,98 @@ SPDX-FileCopyrightText: 2024 zormit <nt4u@kpvn.de>
...
+        {
+            function()
+                return require("teamtype").status()
+            end,
+        },
+    },
+}
+```
+
+## When Neovim plugin loses connection to daemon, make buffer read-only
```
alerque:
```suggestion
## When Neovim plugin loses its connection to the daemon, make buffers read-only
```

CHANGELOG.md
```diff
@@ -5,38 +5,98 @@ SPDX-FileCopyrightText: 2024 zormit <nt4u@kpvn.de>
...
-Add a Nix Flake to the project so the repository can be run directly, used as a dependency, or just to supply a developer environment.
+*Teamtype enables real-time peer-to-peer collaborative editing of local files using your own text editor or other tools.
+For example, you can use it for pair programming sessions, brainstorming, or note-taking.
+It's the missing real-time complement to Git and other version control systems!*
 
-Continue trying to contact Magic Wormhole on a 5 minute cycle in the event join codes fail to register.
+This release adds a couple of features and fixes some bugs.
 
-Add status widget to the Neovim plugin with basic information to be used in statuslines.
+Special thanks to @caleb for putting in a lot of work behind the scenes to clean up various aspects of our code base.
```
alerque:
```suggestion
Special thanks to @alerque for putting in a lot of work behind the scenes to clean up various aspects of our code base.
```

Assuming this will mostly be seen on GitHub (or crates.io / docs.rs / etc.). If it is is some context without support for linking usernames in this namespace my first name is fine, but then it probably shouldn't be `@` mentioned.

alerque (approved):
Nice work! We can merge this just before running the other release tasks in case there are more adjustments, but this already sounds pretty good to me.

I may be an advocate for auto-generating changelogs using structured & filtered commits, but that's only to get the ball rolling in the right direction and reduce friction around the whole release process. This kind of curated notes takes lots of effort, but when possible the extra hand-touched copy editing is great.

CHANGELOG.md
blinry: Good catch!

alerque (approved):

PR #591, opened 2026-07-01
Add option `--always-remove-socket`

axelmartensson:
Add option `--always-remove-socket` to always remove `.teamtype/socket` if it already exists on startup without asking for confirmation.

**Self-review checklist**

- [X] I described the changes of the PR in the [CHANGELOG](https://github.com/teamtype/teamtype/blob/main/CHANGELOG.md).
- [X] I have manually tested and self-reviewed my changes.
- [X] I have added myself to the REUSE headers in the files where I made significant changes, see [CONTRIBUTING.md](https://github.com/teamtype/teamtype/blob/main/CONTRIBUTING.md).
- [X] This PR does not contain content generated by LLMs, see our [generative AI policy](https://github.com/teamtype/teamtype/blob/main/CONTRIBUTING.md).

## Motivation
Always removing `.teamtype/socket` if it already exists on startup without asking for confirmation is useful when running the teamtype daemon non-interactively, e.g. when the teamtype daemon is run as a service.

The use-case is enabling the teamtype daemon service to start automatically without user input after a reboot.

The scenario is:

1. Daemon running -> `.teamtype/socket` exists
2. Hard reboot -> `.teamtype/socket` does not get cleaned 
   up, and still exists when the computer is started
3. Daemon starting ->
  - If `--always-remove-socket` is set: the user is not prompted, and the old socket is removed.
  - If `--always-remove-socket` is not set: the user is prompted, like before.

## Example that reproduces the issue

Running this (where we simulate the socket already existing 
by using `touch` to create an empty file):
```sh
set -e
TEAMTYPE_COMMAND=$(pwd)/target/debug/teamtype
TEMP_DIR=$(mktemp -d)
trap "rm -rf $TEMP_DIR" EXIT
cd $TEMP_DIR
mkdir --mode=700 .teamtype
touch .teamtype/socket
teamtype share
```

Results in the prompt:
```
Detected an existing socket '/tmp/tmp.ppw4rAdUZF/.teamtype/socket'. There might be a daemon running already for this directory, or the previous one crashed. Do you want to continue? (y/N): 
```

# Examples that show how this PR fixes the issue

Adding the `--always-remove-socket` option on the command line:
```sh
set -e
TEAMTYPE_COMMAND=$(pwd)/target/debug/teamtype
TEMP_DIR=$(mktemp -d)
trap "rm -rf $TEMP_DIR" EXIT
cd $TEMP_DIR
mkdir --mode=700 .teamtype
touch .teamtype/socket
$TEAMTYPE_COMMAND share --always-remove-socket
```

Results in a successful startup without a prompt

Same thing for the case where we add the option in the 
config file

```sh
set -e
TEAMTYPE_COMMAND=$(pwd)/target/debug/teamtype
TEMP_DIR=$(mktemp -d)
trap "rm -rf $TEMP_DIR" EXIT
cd $TEMP_DIR
mkdir --mode=700 .teamtype
touch .teamtype/socket
echo always_remove_socket=true > .teamtype/config
$TEAMTYPE_COMMAND share
```

Results in a successful startup without a prompt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,14 @@
 <!--
 SPDX-FileCopyrightText: 2024 blinry <mail@blinry.org>
 SPDX-FileCopyrightText: 2024 zormit <nt4u@kpvn.de>
+SPDX-FileCopyrightText: 2026 axelmartensson <axel.martensson@hotmail.com>
 
 SPDX-License-Identifier: CC-BY-SA-4.0
 -->
 
 # 0.9.2 (unreleased)
+
+Add option `--always-remove-socket` to always remove `.teamtype/socket` if it already exists on startup without asking for confirmation.
 
 Add a Nix Flake to the project so the repository can be run directly, used as a dependency, or just to supply a developer environment.
 
--- a/crates/teamtype/src/cli.rs
+++ b/crates/teamtype/src/cli.rs
@@ -49,6 +49,9 @@ pub struct ShareJoinFlags {
     /// or if `HOME` is not set in `/home/$USER/.cache/teamtype/`.
     #[arg(short, long)]
     pub temporary_directory: bool,
+    /// Always remove `.teamtype/socket` if it already exists on startup without asking for confirmation.
+    #[arg(long)]
+    pub always_remove_socket: bool,
 }
 
 #[derive(Subcommand)]
--- a/crates/teamtype/src/config.rs
+++ b/crates/teamtype/src/config.rs
@@ -30,6 +30,7 @@ const EMIT_JOIN_CODE_DEFAULT: bool = true;
 const EMIT_SECRET_ADDRESS_DEFAULT: bool = false;
 // TODO: Generate a random funny name.
 const USERNAME_FALLBACK: &str = "Anonymous";
+const ALWAYS_REMOVE_SOCKET_DEFAULT: bool = false;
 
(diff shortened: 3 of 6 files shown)


zormit:
hey @axelmartensson thanks for your PR, could you please add the [self-review checklist](https://github.com/teamtype/teamtype/blob/main/.github/pull_request_template.md) :)

alerque:
> The use-case is enabling the teamtype daemon service to start automatically without user input after a reboot.

A reboot shouldn't be leaving around a dead socket anyway. Can we get an issue report about how that is happening? That's something to fix in the signal trapping / shutdown procedures.

blinry:
Thanks for trying to solve this, @axelmartensson!! It's been bugging me too.

I created #592 to track this issue and to discuss possible other solutions (it has been on our "dogfooding annoyances" list for quite a while).

My feeling is that adding a flag to ignore this might not result in the best way to go – users might create aliases to always use it, and then they don't have the (desired) protection against situations where there's already a daemon launched for the directory anymore. Instead, we should aim deeper at the core of the issue.

axelmartensson:
Ok, I see what you mean, closing this PR and moving the discussion to #592

blinry:
We gave fixing this a shot in #598, @axelmartensson! :)

axelmartensson:
Awesome, thank you so much :-)

PR #598, opened 2026-07-15
feat(daemon): Auto-detect whether existing socket is connected to daemon

### LINKED ISSUE #592: Solve the "Detected an existsing socket" situation
**Is your feature request related to a problem? Please describe.**

When the daemon crashes, it often leaves behind the `.teamtype/socket`. When the user tries to start the daemon the next time, it will ask:

```
Detected an existing socket '/home/blinry/tmp/playground/.teamtype/socket'. There might be a daemon running already for this directory, or the previous one crashed. Do you want to continue? (y/N):
```

This is meant to be our protection against the user running the daemon on the same directory twice by accident. But it's pretty annoying, and happens to me often.

**Describe the solution you'd like**

Ideally, users would never need to see this prompt. If we would *know* there was another daemon, we could refuse to start. Otherwise, the command should succeed.

**Ideas for possible solutions**

- One way we could solve this was by checking whether the socket is actually connected to another daemon process – if not, we can go ahead and overwrite it? Is there a stable way to tell if there's a process connected to a socket?
- Another direction would be to do our best that the daemon cleans up the socket as best as it can, even when it `panic`s.
- A "hack": The daemon could try connecting to the socket as a client, this should fail if there's no daemon yet, correct?

blinry:
If there's an existing socket, try to connect to it as a client. If that fails, we assume there's no other daemon runing and we can delete the socket.

Fixes #592.

**Self-review checklist**

- [ ] I described the changes of the PR in the [CHANGELOG](https://github.com/teamtype/teamtype/blob/main/CHANGELOG.md). -> skipping for now, to be discussed

### DIFF
--- a/crates/teamtype/src/editor.rs
+++ b/crates/teamtype/src/editor.rs
@@ -6,7 +6,7 @@
 
 //! This module is all about daemon to editor communication.
 
-use std::os::unix::fs::PermissionsExt;
+use std::os::unix::{fs::PermissionsExt, net::UnixStream as StdUnixStream};
 use std::path::{Path, PathBuf};
 use std::{env, fs};
 
@@ -99,7 +99,6 @@ pub(crate) fn strip_current_dir(path: &Path) -> PathBuf {
 pub fn spawn_socket_listener(
     socket_path: &Path,
     document_handle: DocumentActorHandle,
-    bool_prompter: &dyn Fn(&str) -> Result<bool>,
 ) -> Result<()> {
     // Make sure the parent directory of the socket is only accessible by the current user.
     if let Err(description) = is_user_readable_only(socket_path) {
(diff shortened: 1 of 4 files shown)


alerque (approved):
This change simultaneously makes lots of sense and no sense at all. As far as user-facing behaviour goes –and even as far as coding on top of *main* goes– it makes much more sense than the current behavior. It should be both more robust and easier to use.

The caveat is from an architecture perspective this is a regression in the general shape of the library interface because it takes away the ability for the daemon to interact with the user (including returning error messages). The trouble here is that this code was always a half-way measure towards abstracting the user interface to a trait and is correctly identifying the places the UI functionality will be passed. Further the other half of the work was already done and the UI API was already addressed.

As such this would have been much easier to code up and allow better error handling if it had been stacked on #541. It's okay that this is based on *main* and I can rebase the work in other PRs since we already discussed splitting that one up a bit, but I think the general shape of where the UI trait will go and how to pass it seems pretty settled, just the logging vs. messaging split needs work.

crates/teamtype/src/editor.rs
```diff
@@ -112,14 +111,17 @@ pub fn spawn_socket_listener(
...
-        let remove_socket = bool_prompter(&format!(
-            "Detected an existing socket '{socket_path_display}'. There might be a daemon running already for this directory, or the previous one crashed. Do you want to continue?"
-        ));
-        if remove_socket? {
-            sandbox::remove_file(Path::new("/"), socket_path).expect("Could not remove socket");
-        } else {
-            bail!("Not continuing, make sure to stop all other daemons on this directory");
+        // If there's an existing socket, try to connect to it as a client. If that fails, we assume
+        // there's no other daemon running and we can delete the socket.
+        match StdUnixStream::connect(strip_current_dir(socket_path)) {
```
alerque, line 116: This will need some abstraction to work with #534 which allows Windows to use named pipes while Unix continues with sockets.

alerque: And because I just did the necessary refactoring I'll note here something that will come up in review. I'm sure it seemed like reaching for the sync function in `std::net::unix::UnixStream` was the clean way to do this since this is currently a sync function context, but there is a catch. The only available named pipe implementation we can use is the tokio one which of course is async. Hence actually doing this check using the same async `UnixStream` function would actually have made sense even though it is a bit awkward. It's awkwardness we need to overcome or redesign for anyway.

alerque: https://github.com/teamtype/teamtype/pull/534/changes#diff-47e626997c2a1acc29086fcd3e5aa03efff9e418b3b474f4abc5aed6a9381bbdR111-R122

PR #577, opened 2026-06-17
Automate release workflow

alerque:
**Self-review checklist**

- [ ] I described the changes of the PR in the [CHANGELOG](https://github.com/teamtype/teamtype/blob/main/CHANGELOG.md).
- [ ] I have manually tested and self-reviewed my changes.
- [ ] I have added myself to the REUSE headers in the files where I made significant changes, see [CONTRIBUTING.md](https://github.com/teamtype/teamtype/blob/main/CONTRIBUTING.md).
- [ ] This PR does not contain content generated by LLMs, see our [generative AI policy](https://github.com/teamtype/teamtype/blob/main/CONTRIBUTING.md).

PR #599, opened 2026-07-15
Tooling updates for this release cycle's changelog handling 

alerque:
So #577 is collecting some bits that are hard to test and I'm still fiddling with how to handle the release process.

This is an "early" part of the same branch, and I'm sorting the commit order by readiness for review, so this not chronologically (or even logically) the order, but it is the parts that are ready for review. #577 will continue to work on the bits that need further adjustment and testing.

**Self-review checklist**

~~- [ ] I described the changes of the PR in the [CHANGELOG](https://github.com/teamtype/teamtype/blob/main/CHANGELOG.md).~~


CONTRIBUTING.md
```diff
@@ -37,3 +38,15 @@ We will do this until the number of contributors grow too large and then switch
...
+
+When committing changes to Git, commit messages should follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) pattern.
+Commits of type `feat:`, `fix:`, and `perf:` will become part of the changelog draft during the release process.
+Commits of type `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, and `test:` are filtered out unless they are marked as breaking changes.
+If in doubt about what type best applies, the most likely culprit is trying to mix multiple types of work and should be split into multiple commits.
+
+Note that the commit type `fix:` should only be applied to commits that fix a bug *in an existing tagged release*.
+If the bug being fixed is only on the default branch but not yet released, `chore:` is suitable to not generate irrelevant changelog entries.
+If the bug being fixed is only in an un-merged branch, just amend the commit that introduced the bug to the development cycle.
+Likewise follow up work on something already called out since the last release as a new feature via `feat:` should just be `chore:` or similar.
```
blinry, line 52: Thanks for documenting this!!

Justfile
```diff
@@ -186,3 +188,27 @@ preview-branch-changelog:
...
+
+    	Changelog: {{ read-release-url(semver) }}
+
+    	Installation: https://github.com/teamtype/teamtype#-installation
+
+    	Release Highlights:
+    	- 
+    	- 
+
+    	#teamtype #collaboration #p2p #localfirst #plaintext #editing
```
blinry: I think I wouldn't use these tags "by default" when posting a toot, so I'd say we leave them out of the template.

Justfile
```diff
@@ -186,3 +188,27 @@ preview-branch-changelog:
...
+prepare-release-toot semver:
+    cat <<- EOF
+    	━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+    	Happy to announce the v{{ semver }} release of Teamtype! 🎉
+
+    	Teamtype enables real-time peer-to-peer collaborative editing of local files using your own text editor.
+
+    	Changelog: {{ read-release-url(semver) }}
+
+    	Installation: https://github.com/teamtype/teamtype#-installation
```
blinry: I can see why you went for the `#-installation` URL, but it doesn't seem very stable… Maybe rather "Project: https://github.com/teamtype/teamtype"?

Justfile
```diff
@@ -5,6 +5,8 @@
 cargo := require('cargo')
 cargo-deny := require('cargo-deny')
 git-cliff := require('git-cliff')
+gh := require('gh')
+jq := require('jq')
```
blinry, line 11: I also needed to add `git := require('git')`, as well as

```
default-remote := "origin"
default-branch := "main"
```

Justfile
```diff
@@ -169,3 +172,43 @@ nvim *ARGS: build-test
...
 teamtype *ARGS: build-test
     $TEAMTYPE_BINARY {{ maybe-pass(ARGS) }}
+
+# Get an early look at what the changelog draft would look like for a release.
+[group('release')]
+preview-release:
+    {{ git-cliff }} --unreleased --bump
+
+read-last-tag() := shell(git + ' tag --list v[0-9]*')
+read-git-pager() := shell(git + ' config pager.diff')
```
blinry: I don't have a pager set explicitly, it seems, so this command returns nothing, and makes the command fail. Can we add a sensible default?

Justfile
```diff
@@ -169,3 +172,43 @@ nvim *ARGS: build-test
 [no-cd]
 teamtype *ARGS: build-test
     $TEAMTYPE_BINARY {{ maybe-pass(ARGS) }}
+
+# Get an early look at what the changelog draft would look like for a release.
+[group('release')]
+preview-release:
+    {{ git-cliff }} --unreleased --bump
```
blinry, line 182: This currently outputs:

```
WARN  git_cliff_core::changelog > process_commits: 1 commit(s) were skipped due to parse error(s) (run with `-vv` for details)
WARN  git_cliff                 > The next version is the same as the current version, there is nothing to bump
## [0.9.2] - 2026-07-16
```

I'm not sure whether these warnings are okay, or how I could test this otherwise.

Running `git-cliff --unreleased --bump -vv` shows a lot of

```
TRACE git_cliff_core::process   > process_commits: 378283a - Grouping error: `Skipping commit` (chore(tooling): Process commit messages with spellcheck before they land in changelogs)
```

as well as

```
DEBUG git_cliff > Failed to get remote from repository: RepoError("no remotes configured or HEAD is detached")
```

(that's because of `jj`)

Cargo.toml
```diff
@@ -69,6 +69,20 @@ too_many_lines = "allow"
...
+filter_commits = true
+commit_parsers = [
+  { message = "(tooling)", skip = true },
+  { message = "^feat", group = "<!-- 0 -->New Features" },
+  { message = "^fix", group = "<!-- 1 -->Bug Fixes" },
+  { message = "^perf", group = "<!-- 2 -->Performance Improvements" },
+  { message = ".*", skip = true },
+]
+commit_preprocessors = [
+  { pattern = '.*', replace_command = 'typos --quiet --write-changes -' },
```
blinry, line 84: This seems to be run for *every* commit, it seems, even for ones that are later filtered. Calling `git-cliff` on its own takes ~50 seconds on my laptop that way; it used to be pretty fast before. Could we constrict this so that `typos` is only ran for commits that are included in the changelog?

Also, running `git-cliff` gives me an error:

```
ERROR git_cliff_core::command > process_commits: process: preprocess: error: `calle` should be `called`, `caller`, `calls`
  ╭▸ -:4:19
  │
4 │ the other way and calle the "more local" one a "copy".
```

So here are multiple suggested corrections, I think, which cannot be auto-fixed? That probably shouldn't completely halt the changelog generation.

.github/pull_request_template.md
```diff
@@ -13,7 +13,7 @@ See https://github.com/teamtype/teamtype/blob/main/CONTRIBUTING.md for more guid
 
 **Self-review checklist**
 
-- [ ] I described the changes of the PR in the [CHANGELOG](https://github.com/teamtype/teamtype/blob/main/CHANGELOG.md).
+- [ ] Commit messages follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) pattern.
```
blinry: It would be good to reference the new section in CONTRIBUTING.md here. Two other of the points already link to it, but I think that's fine. Good to make it as easy for people to find the linked information as possible.

So in this point: ", see [CONTRIBUTING.md](...) for details how we use it in this project." ?

Justfile
blinry: Ah, as soon as I actually have a commit in the commit log that's one of the included ones, it seems to work fine.

Justfile
```diff
@@ -169,3 +172,43 @@ nvim *ARGS: build-test
 [no-cd]
 teamtype *ARGS: build-test
     $TEAMTYPE_BINARY {{ maybe-pass(ARGS) }}
+
+# Get an early look at what the changelog draft would look like for a release.
+[group('release')]
+preview-release:
```
blinry: Wondering whether this should be `preview-changelog`, to match the `preview-branch-changelog`.

blinry (changes requested):

Justfile
alerque: Yup, missed that in re-sequencing commits.

Justfile
alerque: Fair enough. At worst the anchor would go away, but since it also saves characters that could be used for describing things, I'll just nuke it.

Justfile
alerque: Try this version. Less explicit but possibly more robust. This is in here because I use a special pager (`delta`) to get colored side-by-side diffs and I can't just hard code that in here without messing others up but it's nice to get the default tooling I get with `git diff`.

Justfile
alerque: Yes. I've been testing this with some dummy `feat: ...` commits strewn around so I actually get something to look at. I'll look at testing the case of completely empty change sets too.

Cargo.toml
alerque: That's surprising to me as the speed difference for me is less than one second (~700ms). However there does seem to be a way to do this filtering on the whole final generated blob instead of each commit so that should address your concern.

Justfile
alerque: One of your warnings is because 961e678 was merged without proper formatting (it is missing the `:` separator).

alerque: I just fixed the *other* error going on here (incorrectly resolving the last released tag). There are still warnings involved if you try to preview a changelog and nothing goes in it, but I think that's probably okay.

Justfile
blinry: Works!

Cargo.toml
blinry: Much better! Maybe Nix wraps `typos` in a way so that it executes a bit slower, and it just compounds.

blinry:
Re-tested everything and seems to work great now. Thanks, @alerque!

blinry (approved):

blinry:
Oh hang on. `just preview-branch-changelog` now fails:

```
WARN  git_cliff_core::changelog > process_commits: 1 commit(s) were skipped due to parse error(s) (run with `-vv` for details)
error: Difftastic does not support being called with 8 arguments.

You can pass 2 arguments, or arguments in the form used by GIT_EXTERNAL_DIFF (7 or 9 arguments). See --help for more details. 

For reference, difftastic was invoked as `/home/blinry/.nix-profile/bin/difft /dev/fd/63 /tmp/nix-shell.jQrOlL/nix-shell.7GbtDc/git-blob-vtXcjv/63 0000000000000000000000000000000000000000 120000 /tmp/nix-shell.jQrOlL/nix-shell.7GbtDc/git-blob-SU54vd/62 0000000000000000000000000000000000000000 120000 /dev/fd/62`.

fatal: external diff died, stopping at /dev/fd/63
```

Investigating.

blinry:
Interestingly, `bash -c 'diff <(git-cliff v0.9.2..origin/main) <(git-cliff --unreleased)'` works just fine. How can I echo the invidual commands that `just` runs again?

For reference, my .gitconfig contains 

```
[diff]
	algorithm = "histogram"
	external = "difft"
```

alerque:
Hmm. I changed my Git config to use difftastic like yours and can replicate that error. Let me take another stab at something that will work across configs.

alerque:
Well that was a rabbit trail and a half!

Justfile
```diff
@@ -169,3 +175,39 @@ nvim *ARGS: build-test
...
+[group('release')]
+preview-changelog:
+    {{ git-cliff }} --unreleased --bump
+
+read-last-tag() := shell(git + ' describe --tags --abbrev=0 --match="v[0-9]*" HEAD')
+
+# Review what changes the current branch will bring to the next release's changelog draft.
+[group('release')]
+preview-branch-changelog:
+    {{ git }} diff --no-ext-diff --no-index -- \
```
alerque, line 189: I went round and round with this, but the short version is that `difftastic` just does not support some possible use cases. It can diff arbitrary files itself and handles process substitution okay, but in this case Git is resolving the files and then adding more args with meta data. Difftastic can handle some invocations but not others (which is part of the reason the recommend using it as `git difftool` instead of universally).

Since the whole point of using `git diff` instead of just `diff` here is to play nice with any GIT_PAGER that the user may have setup, the GIT_EXTERNAL_DIFF is not particularly relevant, so I'm just side-stepping it entirely.

PR #601, opened 2026-07-24
Add CLI flag to adjust verbosity, default to being less verbose

alerque:
**Self-review checklist**

### DIFF
--- a/crates/e2e-tests/tests/fuzzer.rs
+++ b/crates/e2e-tests/tests/fuzzer.rs
@@ -75,9 +75,9 @@ async fn main() -> Result<()> {
         std::process::exit(1);
     }));
 
-    logging::initialize()?;
+    logging::initialize(true)?;
 
-    let ui = &UserInterface::new(FuzzerInteractions {});
+    let ui = &UserInterface::new(FuzzerInteractions {}, true);
 
     // Set up files in shared directories. The directories will get cleaned up automatically when
     // the handle goes out of scope. We don't *use* the handle but we do need to keep it in scope.
--- a/crates/e2e-tests/tests/fuzzer.rs
+++ b/crates/e2e-tests/tests/fuzzer.rs
@@ -75,7 +75,7 @@ async fn main() -> Result<()> {
         std::process::exit(1);
     }));
 
-    logging::initialize(true)?;
+    let () = logging::initialize(true)?;
 
     let ui = &UserInterface::new(FuzzerInteractions {}, true);
 
--- a/crates/teamtype/src/cli.rs
+++ b/crates/teamtype/src/cli.rs
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: 2025 blinry <mail@blinry.org>
 // SPDX-FileCopyrightText: 2025 zormit <nt4u@kpvn.de>
 // SPDX-FileCopyrightText: 2026 axelmartensson <axel.martensson@hotmail.com>
+// SPDX-FileCopyrightText: 2026 Caleb Maclennan <caleb@alerque.com>
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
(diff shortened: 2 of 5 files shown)


alerque:
Was this (6de6b6e) what you guys were trying to get at yesterday @blinry & @zormit? Looking at the code (instead of working from memory while also getting three little kids around a city on public transit where I don't even speak the language!) I think downgrading the quiet mode handling to only touch the CLI interface implementation makes some sense. I can still see a potential future for generic implementations on that trait, but I also don't see any harm in only implementing this locally for the CLI since that's the only practical use case currently in view and saving the generic implementations for when we have a clearer idea what –if any– abstractions would be useful across multiple UIs.

I still hold that the concrete handle type for holding the Arc where we control things like the Debug implementation are more ergonomic and will be a better setup for use as a library.

crates/teamtype/src/logging.rs
```diff
@@ -5,93 +5,38 @@
...
-            .with_default_directive(LevelFilter::DEBUG.into())
-            .from_env()?;
-        let subscriber = FmtSubscriber::builder()
-            .with_env_filter(filter)
-            .with_thread_ids(true)
-            .with_timer(timer)
+pub fn initialize(verbose: bool) -> Result<()> {
+    if let Ok(level) = env::var("RUST_LOG") {
+        let builder = FmtSubscriber::builder()
+            .with_env_filter(EnvFilter::new(format!("teamtype={level},fuzzer={level}")))
```
blinry: This removes a flexibility we had before: We could set things like `RUST_LOG=iroh=debug` to debug what's going on in the Iroh library, for example. So I'd prefer not to set a custom env filter here, but to keep using the default `from_env()` version.

crates/teamtype/src/cli.rs
```diff
@@ -13,6 +14,9 @@ use clap::{Args, Parser, Subcommand};
 #[command(about, long_about = None)]
 #[command(propagate_version = true)]
 pub struct Cli {
+    /// Suppress non-essential informational messages.
```
blinry: For clarity:

```suggestion
    /// Suppress non-essential informational messages in share/join mode.
```

crates/e2e-tests/tests/fuzzer.rs
```diff
@@ -75,7 +75,7 @@ async fn main() -> Result<()> {
         std::process::exit(1);
     }));
 
-    logging::initialize()?;
+    let () = logging::initialize(true)?;
```
blinry: Is this addressing a new nightly clippy lint? I tried removing it and mine didn't complain. We "handle" the Result using `?` – is this to make explicit that this doesn't return anything?

blinry (changes requested):
Yeah, I like this better to have the `quiet` flag in `ConsoleInteractions` right now!

One regression and two nits:

crates/teamtype/src/logging.rs
alerque: Hmmm, definitely didn't mean to remove that functionality—something I too find useful.

crates/e2e-tests/tests/fuzzer.rs
alerque: I don't remember. I personally don't like having to use this construct, but sometimes the linter forces it so I assume it was a lint because I wouldn't put it there otherwise. But it doesn't reproduce now, so it might have gotten thrown during some intermediate phase where the logging instantiator had a different shape.

crates/teamtype/src/cli.rs
alerque: This wouldn't be strictly true, only currently true by coincidence. Lets be more specific and mention that this is specific to the console output of the CLI.

crates/teamtype/src/logging.rs
alerque: Fixed, although perhaps with a too complex mechanism. I still made it so you could set `RUST_LOG=debug` and it would setup the `teamtype=debug,fuzzer=debug` env filter, but perhaps it is best to just assume if you set RUST_LOG at all you should set it to the standard format with a module=level qualifier.

blinry: Yeah, I think I'd prefer less complex code and default behavior here. :) (I don't see an updated version in the diff, currently, I think.)

alerque: Oops, I pushed the updated branch to the wrong remote.

alerque: Okay I just made the simplification I was thinking of and pushed it (to the right remote this time). It should be reviewable again @blinry.

blinry (approved):
Looks good, thanks for the follow-ups! :)

PR #603, opened 2026-07-24
Deconflate logging vs. user interactions, remove all ui from the lib crate

alerque:
This is the substantive replacement for #541 which got overloaded and stripped down to just a small step in this direction.

I'm not sure how much sense this "stack" makes either. I was playing around with GitHub stacks and it made sense when I started the split and less and less as I went on. This should be evaluated in quick succession with the next bits of the stack: #611 and #601. Some of the refactoring here will not make sense until the final commit of #601, and testing intermediate commits will (in some cases) produce duplicate output. The follow up #616 is unrelated and should just be chronologically later.

Since previous discussion is spread out all over the place I'm going to attempt to pull it together here. The following replies are to notes in our dev wiki, plus comments on previous PRs, plus some Zulip threads.

----

> - Idea for the logging PR: Maybe maybe -v the default, and add -q (--quiet)?

I implemented this, but I do not prefer it. I would still prefer to make the default output as direct and to the point as possible so people don't float around in a sea of irrelevant messages when just getting started. That being said I understand tastes differ here and the Unix world is pretty split between apps that default to being verbose and have `--quiet` flags and ones that default to being terse and have `--verbose` flags. If I'm out voted on this point that's fine, but I'd also be quite happy to flip it back around any time before the next tagged release.

> - option 1: keep log levels (info! warn! error!) as they were for now, but make them into ui methods; use debug! for actual logging
>    - calebs main concern: we'd have a ui.warn that we might not need in the future

This is roughly the route I went, but even a step farther. I completely reserved the `tracing` macros for actual logging, and added UI methods `ui.log()`, `ui.inform()`, and `ui.warn()` for everything user facing. The actual output of `ui.log()` output is gated behind the quiet/verbose mode toggle.

> - option 2: iterate on which messages need to go into which levels

I considered this an unsatisfactory option because no matter what level stuff got put in the tracing macros were still getting overloaded for use in a UI, and no amount of shuffling them would enable us to shoehorn them into alternative UIs to the CLI where logging could mix with the console output.

>    - Reason for immediately accepting the "y": deletions might not be handled well in terminals?
>        -> dialoguer seems to allow pressing enter

I switched from `dialoguer` to `inquire` for a couple reasons. I had initially rejected it because we already had most of the dependency tree for `dialoguer` in our SBOM and it was a minimalistic alternative and I thought `inquire` would be heavier. Also I'm less used to using it. It turns out there is some overlap and `tracing-subscriber` and others are already pulling in most of what `inquire` needs anyway.

The interactions for `Confirm` prompts seem more in line with our expectations, and when we need them the other widgets seem pretty nice to use as well.

Lastly `inquire` seems to have better support for Windows terminal nuances. That's going to matter soon.

>    - some warn! ings we had might be "errors" for the user?
>        -> caleb called it ui.error

I have now called it `ui.warn()`, but again only one level of UI function for showing something is wrong. I could not find a good use case for another level. What we do have is user facing errors that are the result of panics or other bubbled up rust errors. I think the wording of `ui.warn()` for things that seem like they might be a concern but we are still running and functioning, and the use of Rust errors for when things go catastrophically wrong is clearer than overloading "error" for both outcomes.

>  I'm surprised by the new (y/n) confirmation behavior where you don't need to press enter anymore. What's your reasoning? Users not realizing that the need to press enter? It will be a bit jarring to old users at first, I think.

The use of `inquire::Confirm` as well as the options I selected both require enter and show typed output instead of toggling a widget state.

> The new (y/n) dialog doesn't look very pretty on my terminal, it seems to "repeat" the question after confirmation on narrow terminals:

Also better handled by `inquire` than `dialoguer`.

> The new -v flag can only be provided directly after teamtype, not after the subcommands, which makes it less discoverable (in the --help entries of the subcommands, for example). Can we promote it to go after share/join, like --directory?

Addressed, the verbosity choice is a global flag.

> With --verbose, the output contains mixed "regular" output and lines with "INFO:" prefix. Seems @zormit also was a bit surprised a bit. From a user perspective, it'd be nice if the messages were formatted the same, probably?

This has been addressed, no log output at all will go to the terminal unless using `RUST_LOG` which is a non standard use case.

> When using --verbose, some messages are duplicated, like the Git username, or the join code.

This is addressed by only using UI not logging functions, but the effect doesn't "take" until #601.

> What is logged as "share code" in one place, we call "join code" instead.

Now only one place.

> I'd miss some messages in the non --verbose mode: Mosty the messages about which files are initialized/changed/deleted. The "Editor #n connected" was also often useful as a confirmation. If all plugins would have a good status display, maybe we wouldn't need it, but I think we're not there yet.

c.f. notes about quiet vs. verbose defaults above.

> Some general thoughts on the design of logging/showing messages to the user:

>   While the "debug" messages probably don't need to be displayed in a user interface in "regular usage", I feel like maybe most messages we previously showed with "info!" should be displayed even in a TUI?

Yes, and most are ported to `ui.log()` where they could be implemented by a TUI.

>    To avoid having to duplicate messages going to the user and going to the logs in the code, might it make sense to have some kind of wrapper that does both?

Yes, done.

----

**Self-review checklist**

### DIFF
--- a/crates/teamtype/src/cli_ui.rs
+++ b/crates/teamtype/src/cli_ui.rs
@@ -4,6 +4,7 @@
 
 use anyhow::{Context, Result};
 use inquire::Confirm;
+use nu_ansi_term::Color;
 use teamtype::traits::Interactions;
 use tracing::debug;
 
@@ -17,5 +18,16 @@ impl Interactions for ConsoleInteractions {
             .with_default(false)
             .prompt()
             .context("Failed to read answer to y/n prompt")
+    }
+
+    fn inform(&self, message: &str) {
+        debug!("UI inform event: {message}");
+        println!("{message}");
+    }
+
+    fn warn(&self, message: &str) {
+        let yellow = Color::Yellow;
+        debug!("UI warn event: {message}");
+        eprintln!("{}", yellow.paint(message));
     }
 }
--- a/crates/teamtype/src/cli_ui.rs
+++ b/crates/teamtype/src/cli_ui.rs
@@ -2,9 +2,20 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
+use anyhow::{Context, Result};
+use inquire::Confirm;
 use teamtype::traits::Interactions;
+use tracing::debug;
 
 #[derive(Clone)]
 pub struct ConsoleInteractions {}
 
-impl Interactions for ConsoleInteractions {}
+impl Interactions for ConsoleInteractions {
+    fn confirm(&self, question: &str) -> Result<bool> {
+        debug!("UI confirm event: {question}");
+        Confirm::new(question)
+            .with_default(false)
+            .prompt()
+            .context("Failed to read answer to y/n prompt")
+    }
+}
--- a/crates/teamtype/src/cli_ui.rs
+++ b/crates/teamtype/src/cli_ui.rs
@@ -4,7 +4,7 @@
 
 use anyhow::{Context, Result};
 use inquire::Confirm;
-use nu_ansi_term::Color;
+use nu_ansi_term::{Color, Style};
 use teamtype::traits::Interactions;
 use tracing::debug;
 
@@ -18,6 +18,12 @@ impl Interactions for ConsoleInteractions {
             .with_default(false)
             .prompt()
             .context("Failed to read answer to y/n prompt")
+    }
+
+    fn log(&self, message: &str) {
+        let dimmed = Style::new().dimmed();
+        debug!("UI log event: {message}");
+        println!("{}", dimmed.paint(message));
     }
 
     fn inform(&self, message: &str) {
--- a/crates/teamtype/src/cli_ui.rs
+++ b/crates/teamtype/src/cli_ui.rs
@@ -1,0 +1,10 @@
+// SPDX-FileCopyrightText: 2026 Caleb Maclennan <caleb@alerque.com>
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+use teamtype::traits::Interactions;
+
+#[derive(Clone)]
+pub struct ConsoleInteractions {}
+
+impl Interactions for ConsoleInteractions {}
--- a/crates/teamtype/src/main.rs
+++ b/crates/teamtype/src/main.rs
@@ -5,8 +5,6 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-use std::io::Write;
-use std::io::{stdin, stdout};
 use std::path::{Path, PathBuf};
 use std::process::exit;
 use std::{env, panic};
@@ -312,7 +310,7 @@ fn get_directory(temporary_directory: Option<&TempDir>, cli: &Cli) -> Result<Pat
 fn setup_teamtype_directory(
     directory: &Path,
     temporary_directory: Option<&TempDir>,
-    _ui: &UserInterface,
+    ui: &UserInterface,
 ) -> Result<()> {
     if has_ethersync_directory(directory) {
         let old_directory = directory.join(config::LEGACY_CONFIG_DIR);
@@ -322,7 +320,7 @@ fn setup_teamtype_directory(
             &old_directory.display()
         );
 
-        if prompt_bool(&format!(
+        if ui.confirm(&format!(
             "Do you want to rename {}/ to {}/?",
             config::LEGACY_CONFIG_DIR,
             config::CONFIG_DIR,
@@ -351,7 +349,7 @@ fn setup_teamtype_directory(
                 &directory.display()
             );
 
-            if prompt_bool(&format!(
+            if ui.confirm(&format!(
                 "Do you want to enable live collaboration here? (This will create an {}/ directory.)",
                 config::CONFIG_DIR
             ))? {
@@ -367,18 +365,4 @@ fn setup_teamtype_directory(
 
 fn get_current_directory() -> Result<PathBuf> {
     env::current_dir().context("Could not access current directory")
-}
-
-fn prompt_bool(question: &str) -> Result<bool> {
-    print!("{question} (y/N): ");
-    stdout().flush()?;
-    let mut lines = stdin().lines();
-    if let Some(Ok(line)) = lines.next() {
-        match line.to_lowercase().as_str() {
-            "y" | "yes" => Ok(true),
-            _ => Ok(false),
-        }
-    } else {
-        bail!("Failed to read answer");
-    }
 }
--- a/crates/teamtype/src/main.rs
+++ b/crates/teamtype/src/main.rs
@@ -23,7 +23,7 @@ use teamtype::{
 };
 use tempfile::{TempDir, tempdir_in};
 use tokio::signal;
-use tracing::{debug, info, warn};
+use tracing::{debug, info};
 
 use self::cli::{Cli, Commands, ShareJoinFlags};
 
@@ -104,7 +104,7 @@ async fn run_daemon(app_config: AppConfig, init_doc: bool, ui: &UserInterface) -
     config::ensure_teamtype_is_ignored(&app_config.base_dir)?;
 
     if app_config.sync_vcs && config::has_local_user_config(&app_config.base_dir).is_ok_and(|v| v) {
-        warn!(docstr!(
+        ui.warn(docstr!(
             /// You have a local user configuration in your .git/config.
             /// In --sync-vcs mode, this file will also be synchronized between peers.
             /// If your version "wins", all peers will have the same Git identity.
@@ -211,7 +211,7 @@ async fn run_client(directory: PathBuf) -> Result<()> {
         .context("JSON-RPC forwarder failed")
 }
 
-async fn trap_shutdown(_ui: &UserInterface) {
+async fn trap_shutdown(ui: &UserInterface) {
     let interruption = async {
         signal::ctrl_c()
             .await
@@ -231,10 +231,10 @@ async fn trap_shutdown(_ui: &UserInterface) {
 
     tokio::select! {
         () = interruption => {
-            info!("Got SIGINT (Ctrl+C), shutting down");
+            ui.inform("Got SIGINT (Ctrl+C), shutting down");
         }
         () = termination => {
-            info!("Got SIGTERM, shutting down");
+            ui.inform("Got SIGTERM, shutting down");
         }
     }
 }
@@ -315,13 +315,11 @@ fn setup_teamtype_directory(
     if has_ethersync_directory(directory) {
         let old_directory = directory.join(config::LEGACY_CONFIG_DIR);
 
-        warn!(
-            "You have an '{}/' directory, back from when the project was called \"Ethersync\" until October 2025.",
-            &old_directory.display()
-        );
-
-        if ui.confirm(&format!(
-            "Do you want to rename {}/ to {}/?",
+        if ui.confirm(&docstr!(format!
+            /// You have an '{}/' directory, back from when the project was called \"Ethersync\" until October 2025.
+            ///
+            /// Do you want to rename {}/ to {}/?
+            config::LEGACY_CONFIG_DIR,
             config::LEGACY_CONFIG_DIR,
             config::CONFIG_DIR,
         ))? {
@@ -343,21 +341,17 @@ fn setup_teamtype_directory(
                 &directory.display()
             );
             sandbox::create_dir(directory, &teamtype_dir)?;
+        } else if ui.confirm(&docstr!(format!
+            /// '{}' hasn't been used as a Teamtype directory before.
+            ///
+            /// Do you want to enable live collaboration here? (This will create an {}/ directory.)
+            directory.display(),
+            config::CONFIG_DIR
+        ))? {
+            sandbox::create_dir(directory, &teamtype_dir)?;
+            info!("Created! Resuming launch.");
         } else {
-            warn!(
-                "'{}' hasn't been used as a Teamtype directory before.",
-                &directory.display()
-            );
-
-            if ui.confirm(&format!(
-                "Do you want to enable live collaboration here? (This will create an {}/ directory.)",
-                config::CONFIG_DIR
-            ))? {
-                sandbox::create_dir(directory, &teamtype_dir)?;
-                info!("Created! Resuming launch.");
-            } else {
-                bail!("Aborting launch. Teamtype needs a .teamtype/ directory to function");
-            }
+            bail!("Aborting launch. Teamtype needs a .teamtype/ directory to function");
         }
     }
     Ok(())
--- a/crates/teamtype/src/main.rs
+++ b/crates/teamtype/src/main.rs
@@ -65,11 +65,11 @@ async fn main() -> Result<()> {
 
     logging::initialize().context("Failed to initialize logging")?;
 
-    let _ui = &UserInterface::new(ConsoleInteractions {});
+    let ui = &UserInterface::new(ConsoleInteractions {});
 
     let temporary_directory = get_temporary_directory(&cli)?;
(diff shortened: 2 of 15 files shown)


crates/teamtype/src/main.rs
```diff
@@ -326,23 +338,22 @@ fn setup_teamtype_directory(directory: &Path, temporary_directory: Option<&TempD
...
-            warn!(
+            ui.warn(&format!(
                 "'{}' hasn't been used as a Teamtype directory before.",
-                &directory.display()
-            );
-
-            if prompt_bool(&format!(
+                directory.display(),
+            ));
+            if ui.confirm(&format!(
```
blinry: I feel like these two lines should go in one `ui.confirm` call, as the current `warn` message adds context to help people check if they're launching in the correct directory.

crates/teamtype/src/main.rs
```diff
@@ -299,16 +307,20 @@ fn get_directory(temporary_directory: Option<&TempDir>, cli: &Cli) -> Result<Pat
...
-        warn!(
+        ui.warn(&format!(
             "You have an '{}/' directory, back from when the project was called \"Ethersync\" until October 2025.",
-            &old_directory.display()
-        );
+            old_directory.display()
+        ));
 
-        if prompt_bool(&format!(
+        if ui.confirm(&format!(
```
blinry: Same here.

crates/teamtype/src/cli_ui.rs
```diff
@@ -0,0 +1,39 @@
...
+use teamtype::traits::Interactions;
+use tracing::debug;
+
+#[derive(Clone)]
+pub struct ConsoleInteractions {}
+
+impl Interactions for ConsoleInteractions {
+    fn confirm(&self, question: &str) -> Result<bool> {
+        debug!("UI confirm event: {question}");
+        Confirm::new(question)
```
blinry, line 17: By default, `inquire` seems to print yellow question marks in front of questions, and `>` characters after answering questions. I think that adds a bit of noise that carries no benefit. Let's see whether we can disable it?

Image

blinry (changes requested):
This all reads very clean and reasonable.

I don't *love* passing the `ui` object around like that; was thinking briefly whether it could become part of our magic global `AppConfig`, and then we only need to pass in one thing?

What's your best argument again for why we should use an `Interactions` trait vs relying on `tracing` and possibly implementing custom tracing `Subscriber`s. The latter seems to allow more flexible format adaptions, like printing where in the source code the log was caused. It would avoid the need for passing the `ui` in, but instead keep relying on the global `info!` etc. macros. And I liked how you could use `RUST_LOG` to very purposefully set what logging you want to see. (One argument I see that this wouldn't allow questions, but if I see that correctly we don't have any questions inside of the daemon right now.)

(I haven't looked at the stacked PRs yet, I'm assuming some of this will clear up.)

Some minor remarks:

blinry:
> What's your best argument again for why we should use an `Interactions` trait vs relying on `tracing`

(I'm not really questioning that choice, and know that you've based lots of other code on it. Just wanting to make an informed decision.)

alerque:
> This all reads very clean and reasonable.
> 
> I don't _love_ passing the `ui` object around like that; was thinking briefly whether it could become part of our magic global `AppConfig`, and then we only need to pass in one thing?

Yes, passing it around looks like a bit of pain at first. But this is Rust, and this is what you are supposed to do.

The alternatives are magic global state or abuse of internals.

Yes, eventually melding the `AppConfig` struct with the `Interactions` trait and having some kind of instantiated `App` struct might make sense, but it was too much refactoring to do in this series of PRs. Already it's been a huge pain to deal with the scope creep in these branches.

> What's your best argument again for why we should use an `Interactions` trait vs relying on `tracing` and possibly implementing custom tracing `Subscriber`s.

Currently the Teamtype CLI is **abusing** tracing's log mechanism for use as one direction of the command line interface (user facing output). It is doing so in a way that blocks both things we want to accomplish:

* We can't make a headless system service out of the daemon or otherwise make proper use of actual logging, because the log levels are inverted (e.g. some INFO level message being more important than WARN) and many of the messages going to logging would be inappropriate in an actual log file.

* We can't make alternative UIs (most urgently plugins that want to control the UI via RPC, but also a possible TUI, and any use of the shared library directly without an external process) because messages that are supposed to be user interface are mixed up with actual logging, and trying to Jerry-rig a wrapper that bridges a logging subscriber to some other API for use in UI's is a huge mess. It wasn't made for that and it doesn't adapt easily to it. It was *made* for logging, not interactive use, and it deliberately uses some macros with their own global state to manage the backend to get *out of the way* of the flow of code. We need our user interactions to mix into our workflow.

> The latter seems to allow more flexible format adaptions, like printing where in the source code the log was caused. It would avoid the need for passing the `ui` in, but instead keep relying on the global `info!` etc. macros. And I liked how you could use `RUST_LOG` to very purposefully set what logging you want to see.

Logging is still a thing and you can still do that, it has just been decoupled from *interactions*.

A logical next step would be to add options to the CLI to log to a file and/or via syslog, etc. Logging will still be relevant, especially for people writing and debugging plugins.

> (One argument I see that this wouldn't allow questions, but if I see that correctly we don't have any questions inside of the daemon right now.)

That's partly because we ask all the questions up front in the CLI before starting the daemon. If the user interface is over RPC or shared library or WASM bindings or whatever we need a way to do handle user interaction outside of the console. Whether the `Daemon::run` section needs interactivity inside that function is not materially relevant, we need the UI channels anyway and once we have them it makes sense to use them properly everywhere.

alerque:
> > What's your best argument again for why we should use an `Interactions` trait vs relying on `tracing`
> 
> (I'm not really questioning that choice, and know that you've based lots of other code on it. Just wanting to make an informed decision.)

I edited a partial response to this into my previous comment, but let me try this for a TL;DR:

The `Interactions` trait is basically a spec for what any/all new interfaces for Teamtype will need to implement. This includes plugins that want to handle the UI themselves instead of relying on the user to start it up in a terminal out of band. We can't expect people to implement a custom `TracingSubscriber` and filter out what they want and figure out how to sync the messages they get from the data they need to send back *to* the service. That is the wrong tool for the job and all attempts at it will necessarily be brittle and clunky and limit wider adaption.

crates/teamtype/src/main.rs
alerque: I think you've got a good point here, I'll look into that. It probably means sending multiline UI messages, but we already expect to be able to do that so I don't think it's a problem.

crates/teamtype/src/main.rs
alerque: This entire legacy message already gets dropped in #616, but I'll apply refactoring matches the other code in this PR.

crates/teamtype/src/cli_ui.rs
alerque: Somewhat surprisingly there is not. I just [opened a PR for one](https://github.com/mikaelmello/inquire/pull/347) but as of this writing that would require either a Git only dependency or using `cargo patch-crate`.

That being said I'm not convinced that it is just noise with no benefit. The idea [as you can see in the demo](https://github.com/mikaelmello/inquire#demo) and examples is that interactive bits of the UI that expect the user to give inputs are visually separated from bits that are just output and not interactive. This has become a very common paradigm for CLI tools lately and I think people will find it familiar. Even this PR was created using `gh pr create` which uses the exact same widget design (in Go not Rust, but evidently a clone of the widget handling).

Besides just Y/N confirmation prompts I think we'll have use cases for text inputs and maybe other widgets. For example we should be able to prompt for and read a text value for share codes. After that allowing the CLI to "kick" a peer by showing a list of currently connected peers and taking actions on them.

Robust terminal handling (including Windows) was one reason to move towards an input widget library, but allowing the CLI to be more interactive would be another reason.

blinry: Hmm interesting. Open to trying it and possibly iterating on it. It just looked like a typo to me at first.

blinry (approved):

blinry:
Actually – can you give me some more time to think about this, @alerque? :thinking: Just out of time for today.

blinry:
(I noticed that this PR doesn't convert *all* tracing calls inside the daemon into `ui` calls, and all that aren't will be gone from the UI two stack-steps up. So reviewing which ones will be dropped needs to happen here.)

alerque:
> (I noticed that this PR doesn't convert _all_ tracing calls inside the daemon into `ui` calls, and all that aren't will be gone from the UI two stack-steps up. So reviewing which ones will be dropped needs to happen here.)

That is correct.

We can of course iterate on which messages are UI vs. which are debug tracing, but this PR makes an initial attempt to split them up. The logic in roughly to answer questions like these:

* "Would this message be potentially relevant to an end using using Teamtype through a wrapped interface (a plugin that handles the daemon in the background, etc.)?" vs.
* "Is this message directed at developers writing a plugin or debugging a problem case?"

If the former, UI it is. If the latter, they can still use tracing.

That means for the *most part* I made `error!`, `warn!`, and `info!` into UI, and `debug!` stayed as logging, but there were some `info!` invocations that stayed because they seemed really far removed from end user activities (they would be noise to surface inside an editor) and/or the place in the code didn't make sense to be contacting the user. There were a couple places I could have gone either way, but having a UI object embedded in some of the automerge wrappers for example (document actors, etc.) did not feel right. In fact I did id and got it working and then ripped it back out because without most of those functions even returning a `Result` it seemed like a bad design to have side effects of user interface messaging happening. If there is an error in there we should probably rewrite those traits to bubble up errors to a place that makes sense to contact the user, not embed the UI deeply into those handling functions.

blinry (changes requested):
So in my own usage, I very much like to see logs for file events (the `info!`-level logs like "Removing file", "Creating file", "Initializing binary" or "Detected change of"). I rely on them to better "see" what's going on, for example whether we accidentally shared the `target/` folder, or which files are shared inside of `.git` when using `--sync-vcs`.

So I'd like it if all `info!` logs would be converted to `ui.log` invocations, otherwise this PR stack is a real regression for me. I started giving it a go myself, but this indeed led to situations where we need to store a `ui` field in structs like `Document` or `DocumentActorHandle`. (Left an incomplete attempt at that on the https://github.com/teamtype/teamtype/tree/deep-ui branch.) And that would mean that in all tests involving those, we'd need a dummy UI again. I see three options:

1. Accept that "Rust wants us to do that" and store references to the `UserInterface` deeply in the code.
2. Refactor the code so that logs don't happen so deeply. For example, probably the log could be moved from `Document.set_file` to the places where that function is called in `Daemon`? (Leading so some redundancy/less "encapsulation" of that log, potentially.)
3. Investigate whether we could write our own "UI output" macros that run on a "singleton UI object" or something? Have other applications found solutions for this that we could copy, maybe?

I haven't looked yet at the current `warn!` and `error!` logs yet, but my feeling is that I'd *want* them to be surfaced to all users always, so they can tell when something goes wrong, right? This doesn't seem to be relevant only for developers who could use `RUST_LOG`. (This might involve potentially rewriting more methods to return `Result`s, like you described?)

alerque:
> So in my own usage, I very much like to see logs for file events (the `info!`-level logs like "Removing file", "Creating file", "Initializing binary" or "Detected change of"). I rely on them to better "see" what's going on, for example whether we accidentally shared the `target/` folder, or which files are shared inside of `.git` when using `--sync-vcs`.

Okay. I'm far more interested in getting the separation of concerns under control than over-thinking what level of verbosity is useful where. The later is easy to iterate on, the former is holding everything else up. And I'm not saying you are over-thinking it, just that my trying to pick and choose at this point would be.


> So I'd like it if all `info!` logs would be converted to `ui.log` invocations, otherwise this PR stack is a real regression for me.

Stand by for a quick refactor with all messages ported.

> I started giving it a go myself, but this indeed led to situations where we need to store a `ui` field in structs like `Document` or `DocumentActorHandle`.

I've done this at least three times already, no need for a forth ;-)

> (Left an incomplete attempt at that on the [`deep-ui`](https://github.com/teamtype/teamtype/tree/deep-ui) branch.) And that would mean that in all tests involving those, we'd need a dummy UI again. I see three options:

I don't think it's *that* complicated to inject the dummy UI. Here is the whole thing.


```rust
use crate::traits::Interactions;

struct TestInteractions {}

impl Interactions for TestInteractions {
    fn confirm(&self, _question: &str) -> Result<bool> {
        Ok(false)
    }

    fn log(&self, _message: &str) {}

    fn inform(&self, _message: &str) {}

    fn warn(&self, _message: &str) {}
}
```

Looking at your branch I think you were on the right track. It doesn't have to be too complicated because the UI can be made part of other structs so all the functions in the impl blocks can reach it.

>     1. Accept that "Rust wants us to do that" and store references to the `UserInterface` deeply in the code.
> 
>     2. Refactor the code so that logs don't happen so deeply. For example, probably the log could be moved from `Document.set_file` to the places where that function is called in `Daemon`? (Leading so some redundancy/less "encapsulation" of that log, potentially.)
> 
>     3. Investigate whether we could write our own "UI output" macros that run on a "singleton UI object" or something? Have other applications found solutions for this that we could copy, maybe?

Yes of course there are ways, but the main way they accomplish that is ripping all the UI hooks out of all the "deep" places in the API and handling application logic as close to the top layer as possible. This keeps the UI out of places that don't make sense to pass in the struct.

Also when you're working with just an application with only one possible UI you have some freedom to be more abbusive with things like global singletons.

I think we'll do fine fixing and matching 1 and 2 here: starting with just passing it where it is used, then gradually refactoring it to move the UI calls "up" and encapsulate other functionality "deeper".

> I haven't looked yet at the current `warn!` and `error!` logs yet, but my feeling is that I'd _want_ them to be surfaced to all users always, so they can tell when something goes wrong, right? This doesn't seem to be relevant only for developers who could use `RUST_LOG`. (This might involve potentially rewriting more methods to return `Result`s, like you described?)

Long term more function signatures need to return `Result`s, but I think it will be better to refactor one thing at a time and first introducing the UI, then moving on to how Rust errors are handled will be more productive than trying to refactor both logging and Rust error handling.

alerque:
I [took another swing at this](https://github.com/teamtype/teamtype/compare/aa0ede3f337a9c7b30ead61816bd2a0ec3a8b065..a75468fffb9fa083b07a67c5e80936a7661c0267) and we're down to 3 invocations of `warn!`. I'm out of time for the moment, but my first instinct is to propose that those really are in places where we should only be logging, not reaching into the UI. Two of them are in a `TryFrom` implementations for other people's types and the other is truly a debug statement targeting our fuzzer and we really should resolve what is going on there not expect it to become a UI interaction.

About half the other places we're now passing the UI make perfect sense, and the other half can be refactored. I think that should be part of the gradual API cleanup I'm trying to do ... making the published API surface make sense and modularizing things etc. I don't think this is worse than what we have which is a magic global singleton doing UI from deem inside those same places. In fact I think it is better and much easier to refactor now that the functions are not relying on side effects. It should be much easier to reason about how to reshape them with the UI visible than invisible.

alerque:
I took another look and it turns out two `warn!`s were in the implementation for a custom Enum type `PatchEffect`, not `Automerge::Patch` like I read it the first time.

I just added a commit to change the implementation of a standard trait to a custom one and injected the UI. I still assert this is too deep to be throwing warnings to the users, but that whole type conversion needs a major overhaul to address that. I could take it or leave it personally, but for now if you insist that all previous log messages get surfaced, this does that. Bar one.

The last one is is in `OTServer` and already a weird heuristic. I'm going to suggest that one stays how it is ... or make it `debug!`.

By the way once this split is complete we should be free to go back and start doing much more logging with normal level usage available to us.

alerque:
For the record I have tried merging #617 into this and the fuzzer does pass, so at the end of the day all the changes here still pass that testing. There is a small merge conflict to resolve but I think it's easier to manage these PRs and review them separately *without* stacking them. Either one can go first.

blinry (approved):

PR #608, opened 2026-07-24
Notify user when attempting to remove defunct socket

alerque:
I probably should have thought of this when reviewing #598, but this is a perfect use case for logging. ~~The user doesn't normally need to see this, but when you are trying to debug where things are getting to it's helpful to have debug messages as each operation phase is launched.~~ The user may elect not to see this information, but the fact that a defunct socket was found at all suggests some previous failure mode they may not have understood or wish to investigate further.

**Self-review checklist**

### DIFF
--- a/crates/teamtype/src/editor.rs
+++ b/crates/teamtype/src/editor.rs
@@ -21,7 +21,7 @@ use tokio_util::{
     bytes::BytesMut,
     codec::{Decoder, Encoder, FramedRead, FramedWrite, LinesCodec},
 };
-use tracing::{debug, error, info};
+use tracing::{debug, error, info, warn};
 
 use crate::daemon::{DocMessage, DocumentActorHandle};
 use crate::editor_protocol::{
@@ -120,6 +120,9 @@ pub fn spawn_socket_listener(
                 );
             }
             Err(_) => {
+                warn!(
+                    "An existing socket was found for this directory, but since the daemon seems to be defunct it is being removed."
+                );
                 sandbox::remove_file(Path::new("/"), socket_path).expect("Could not remove socket");
             }
         }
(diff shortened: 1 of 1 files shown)


zormit:
interesting that this makes it fail to compile... Otherwise I like the idea and would approve :)

alerque:
It doesn't fail to compile, it fails `clippy` lint checks. What's funny is the logic it is complaining about is from #598, but it does not trigger the lint there or on the default branch. It does trigger as soon as the match arm has 2 statements in it. The same logic applies either way and I think the lint probably *should* catch both situations. I'd have to dig into the lint code itself to figure out why it makes an exception for the single statement arm.

In any case the lint is worth fixing, the suggested refactor is quite a bit easier to follow than the original.

crates/teamtype/src/editor.rs
```diff
@@ -113,16 +113,13 @@ pub fn spawn_socket_listener(
...
-            }
-            Err(_) => {
-                sandbox::remove_file(Path::new("/"), socket_path).expect("Could not remove socket");
-            }
+        if StdUnixStream::connect(strip_current_dir(socket_path)).is_ok() {
+            bail!(
+                "Detected an existing daemon running for this directory. Rejecting to start another one."
+            );
         }
+        warn!("Found defunct socket, attempting to remove.");
```
zormit: So now you chose a `warn` (whereas the previous [commit](https://github.com/teamtype/teamtype/commit/59415af03b7516cb340e6297b94515b837908bf7) had `debug`)? That probably assumes that we will merge the other ui / messaging PRs soon - without them I would expect it to become a user-visible message after only this is merged? (didn't check any of this - I might be misunderstanding something)

alerque: Yes I changed the level from my first suggestion, but this is not dependent on the logging or UI changes at all. Having thought about what was going on here I am suggesting that as an end user I kind of expect some heads up in the event of an improper shutdown. Even in the case where no harm is done, the fact that a defunct socket was found at all is information I would typically like to be notified of even though I'm allowed to move on.

alerque: I just made this message a bit more verbose and 'user friendly' as it is less of a debug diagnostic and more a heads up about something slightly anomalous the user should understand.

zormit:
Thanks for working & iterating on this!

Issue #455, opened 2025-11-24
Update the video in the README to say "teamtype"


blinry:
(It still uses the old name, "ethersync".)

blinry:
When we do this, we can also remove the note at the top with the old name.

PR #616, opened 2026-08-01
Drop legacy Ethersync compatibility shims

alerque:
These have served through a couple releases since the name change and we've all agreed they have probably served their time. Removing them has come up several times but since I had already refactored them and didn't want to rebase to extract all the changes I just saved it for a convenient time.

https://teamtype.zulipchat.com/#narrow/channel/557642-general/topic/Longevity.20of.20Ethersync.20compatibility.20hooks/with/598424113

Also having this in until now has served as an easy way to test #603: just `mkdir .ethersync` then try `just teamtype share` to get both a warning an a prompt ;-)

**Self-review checklist**

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


blinry:
Some commits of the last stacked PR seem to be in here, as well, I wonder whether that's caused by a bug in GitHub's stack feature, or in the way we use it.

Content looks good, though! After this, the only occurrences of /[Ee]thersync/ seem to be in old changelogs, in links, and in EdJoPaTo's email address.

There is one other place: The note at the top of the README. I'd say we remove that once we re-record the demo GIF (#455). Right now, it can help ease the confusion.

I think this is a "feature" removal (or rather, deprecation warning removal) that would be nice to show in the changelog. Are there common ways to do that using Conventional Commits?

blinry (approved):

alerque:
> Some commits of the last stacked PR seem to be in here, as well, I wonder whether that's caused by a bug in GitHub's stack feature, or in the way we use it.

The stacks feature currently doesn't handle cases where stacked branches don't rebase cleanly when base branches introduce merge conflicts. I forced pushed something to the CLI flag handling that touched a line affected here, so there is a merge conflict and GH stacks throws in the towel and makes you handle it manually.

> I think this is a "feature" removal (or rather, deprecation warning removal) that would be nice to show in the changelog. Are there common ways to do that using Conventional Commits?

Yes, at least one commit involved in the shim removal should flag (`!`) in the type and call out `BREAKING CHANGE: ...` in the description noting the handling change. There is no special handling for deprecations / removals except to call them out as breaking changes. Normally `chore` commits wouldn't make it past our changelog filtering, but anything marked as breaking regardless of commit type/scope should end up in the changelog draft.

blinry:
> Yes, at least one commit involved in the shim removal should flag (`!`) in the type and call out `BREAKING CHANGE: ...` in the description noting the handling change.

Ohh, that's a good solution!

PR #617, opened 2026-08-01
Dust off the fuzzer

alerque:
So the Iroh update broke the fuzzer (specifically 32e00db419114ad66f6e7f0f61d69ebbf4a9d267).

And we didn't notice because we'd silently stopped running the fuzzer in CI. We limit runs of the fuzzer to one run per OS platform using the latest Neovim in the matrix, but then we jinked ourselves by bumping the patch version of Neovim without updating which version the fuzzer was gated on.

I'm still looking into the actual breakage, but first to get CI going again...

**Self-review checklist**

### DIFF
--- a/.github/workflows/general.yml
+++ b/.github/workflows/general.yml
@@ -53,7 +53,7 @@ jobs:
         run: cargo test --locked --no-default-features -p teamtype
       - name: Run default e2e tests
         run: cargo test --locked --no-default-features -p e2e-tests
-      - if: ${{ matrix.nvim-version == 'v0.11.2' }}
+      - if: ${{ matrix.nvim-version == 'v0.12.2' }}
         name: Run e2e fuzzer
         run: cargo test --locked --no-default-features -p e2e-tests --test fuzzer
 
--- a/.github/workflows/general.yml
+++ b/.github/workflows/general.yml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        nvim-version: [v0.12.2]
+        nvim-version: [v0.12.4]
         include:
           - os: ubuntu-latest
             nvim-version: v0.9.5
@@ -53,7 +53,7 @@ jobs:
         run: cargo test --locked --no-default-features -p teamtype
       - name: Run default e2e tests
         run: cargo test --locked --no-default-features -p e2e-tests
-      - if: ${{ matrix.nvim-version == 'v0.12.2' }}
+      - if: ${{ matrix.nvim-version == 'v0.12.4' }}
         name: Run e2e fuzzer
         run: cargo test --locked --no-default-features -p e2e-tests --test fuzzer
 


blinry:
Oops. Feel free to document somewhere what you learn about the Fuzzer breakage, maybe we'd have more thoughts.

Waiting for a fix there probably makes sense before merging this, right?

alerque:
Yes, it makes sense to wait until we have a *fix* for the fuzzer to re-enable CI for it. We (now) know it is broken, but the development builds of the daemon are already usable for testing Iroh 1.0 so clogging up CI and blocking other progress doesn't make sense. I bisected and am pretty certain the breakage is some Iroh expectation, not anything else we have done or are doing.

alerque:
@blinry Your suggestion of timing changes got me poking in a different direction and I may have gotten closer to the root cause. The panic we're hitting in the fuzzer seems to be this one:

https://github.com/alerque/teamtype/blob/dust-off-fuzz/crates/teamtype/src/document.rs#L236-L238

It only happens because we're creating two directories with blank files, then starting two daemons, one in share and one in client mode, and we then link them and feed one side changes and see how they resolve on the other side. The trouble seems to be with the initial document document states not being part of the same document tree. For the client we were waiting until the blank file exists before starting the editor, but it is possible (apparently now probable) for the daemon to still be working through the initial negotiation and for a FileRemoval event to trigger after we've proceeded with other setup because the client is still sorting out whose empty document is going to win.

We can avoid this in the fuzzer by not expecting the sync to dig out of an out-of-sync initial state at all while we're still setting it up (and these commits do that) but it doesn't resolve the possible panic situation. I added a unit test that (I think) crashes the daemon the same way the fuzzer was, then took a stab at resolving it.

I'm fairly sure the changes to the fuzzer make sense, but I'm not at all confident that papering over it in the document handler the way I do here is correct.

crates/e2e-tests/tests/fuzzer.rs
```diff
@@ -60,24 +60,24 @@ async fn main() -> Result<()> {
...
-    app_config2.peer = Some(config::Peer::SecretAddress(daemon.address.clone()));
-    let peer = Daemon::new(app_config2, false, false).await?;
+    app_config2.peer = Some(config::Peer::SecretAddress(daemon1.address.clone()));
+    let daemon2 = Daemon::new(app_config2, false, false).await?;
 
     // Wait until file2 appears.
     while !file2.exists() {
         dbg!("{&file2} doesnt");
-        sleep(Duration::from_millis(100)).await;
+        sleep(Duration::from_millis(500)).await;
```
blinry: Did this change turn out to be important?

blinry: Note: Might have been to get around CI failures…

crates/teamtype/src/document.rs
```diff
@@ -465,6 +478,25 @@ mod tests {
...
+    // event from another peer. Expect the edit to be applied, but to an otherwise empty document
+    // not the content still cached from before the deletion.
+    #[test]
+    fn update_after_removal_recovers_from_cache() {
+        let mut document = Document::default();
+        let file = RelativePath::new("text");
+
+        document.set_file(Content::String("foo".to_string()), &file);
+
+        document.files.remove(&file);
```
blinry, line 501: While this test indeed fails without the included fix, I'm wondering how the situation we set up here can happen in the real world: `document.files` is our Document-internal "file content cache", and it's not `pub`/not used from outside the Document.

When a peer deletes a file, it arrives via an `AutomergeSyncMessage`, we detect a `PatchEffect::FileRemoval`, and update `self.files` there.

If creating the same file in both peers was part of the problem, an Automerge conflict seems to have been involved. Needs some more investigation; I think this points at another cache inconsistency bug.

blinry: 1. `AutomergeSyncMessage` arrives with `PutMap`, we detect a conflict in `PatchEffect::try_from()`, and emit a `FileRemoval`.
2. In `Document.receive_sync_message_log_patches()`, because it's a `FileRemoval`, we delete from `self.files`.
3. But then, when processing the `DocMessage::ReceiveSyncMessage`, we detect that the file is open in an editor (we don't `own()` it). So we "only" send a delta to make it empty.

Result: The file exists in `Document`'s `doc`, but not in `files`, reproducing the situation we test here. There's probably a better way to fix this at the root of the problem. But out of time for now.

alerque:
Recent push just rebases and resolves conflicts, it doesn't address PR review.

PR #619, opened 2026-09-11
Create dedicated types to properly handle data structures

alerque:
This is the second half of what used to be #571.

In Rust, having dedicated types for each bit of data you pass around internally makes everything better.

----

**Self-review checklist**

- [x] This PR does not contain content generated by LLMs¹, see our [generative AI policy](https://github.com/teamtype/teamtype/blob/main/CONTRIBUTING.md).

¹: <sub>A caveat on this front. Most of this was implemented directly by me, but after writing the `impl From<bool> for ConsoleVerbosity` myself I got a bit hung up on how to get Clap to us it when handling a boolean flag. I always get stuck on the incantation. I ended up 'searching' for the answer and getting the `value_parser` bit using an LLM agent with access to the source code in plan mode.</sub>

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


crates/teamtype/src/config.rs
```diff
@@ -34,9 +38,75 @@ pub enum Peer {
...
+            format!(
+                "Could not compute the absolute, canonical form of the path of directory {}",
+                path.display(),
+            )
+        })?;
+        Ok(Self::Permanent(base_dir))
+    }
+}
+
+impl Deref for BaseDir {
```
blinry, line 72: Making this type `Deref` is very clever, I love how cleanly this can be used now!

crates/teamtype/src/config.rs
```diff
@@ -34,9 +38,75 @@ pub enum Peer {
...
+        match self {
+            Self::Permanent(path) => path,
+            Self::Temporary(temp_dir) => temp_dir.path(),
+        }
+    }
+}
+
+impl Default for BaseDir {
+    fn default() -> Self {
+        Self::Permanent(PathBuf::from("./"))
```
blinry, line 85: This `Default` behavior seems a bit strange to me, like it could unintentionally create a BaseDir pointing to a directory where `teamtype` is run.

We need it because `Config` wants to derive `Default`, because it's used in `parse_client_config`. My favorite solution would be to remove the Default derives, and to make `run_client` take a `BaseDir` (or `Path`?) directly. The client subcommand is very different from the share/join subcommands anyway, and won't react to anything else right now.

crates/teamtype/src/daemon.rs
```diff
@@ -91,6 +88,8 @@ pub async fn run_daemon(config: Config, init_doc: bool, ui: &UserInterface) -> R
         ));
     }
 
+    debug!("Starting Teamtype on {}.", config.base_dir);
```
blinry, line 91: Possibly a rebasing mistake? There's already a line like this in Daemon::new(), which seems better (to also log in e2e tests).

crates/teamtype/src/daemon.rs
```diff
@@ -163,27 +162,29 @@ struct DocumentActor {
...
-    ui: UserInterface,
+    base_dir: BaseDir,
+    username: Option<String>,
     save_fully: bool,
+    sync_vcs: bool,
+    ui: UserInterface,
 }
 
 impl DocumentActor {
+    // TODO: un-typed boolean args are a no-no!
```
blinry, line 173: Thanks for taking care of that! :D

crates/teamtype/src/sandbox.rs
```diff
@@ -108,17 +108,13 @@ pub fn exists(absolute_base_dir: &Path, absolute_file_path: &Path) -> Result<boo
...
+    if !sync_vcs {
         ignored_things.extend([".teamtype", ".git", ".bzr", ".hg", ".jj", ".pijul", ".svn"]);
     }
 
-    let base_dir = &config
-        .base_dir
-        .clone()
-        .expect("Temporary directory not initialized");
-    let walk = WalkBuilder::new(base_dir)
+    let walk = WalkBuilder::new(&**base_dir)
```
blinry, line 117: Whoa! Could we create this reference once, and reuse it below, and add a comment here on what's going on?

It first dereferences the BaseDir to a &Path, I guess, but why do we need another &*?

crates/teamtype/src/sandbox.rs
```diff
@@ -180,16 +176,15 @@ pub(crate) fn enumerate_non_ignored_files(config: &Config) -> Vec<PathBuf> {
...
-    let canonical_file_path = check_inside_base_dir_and_canonicalize(
-        &config
-            .base_dir
-            .clone()
-            .context("Temporary directory not initialized")?,
-        absolute_file_path,
-    )?;
-
-    Ok(!enumerate_non_ignored_files(config)
+// TODO: loose the un-type args!
```
blinry, line 179: We can drop this comment now, I think.

crates/teamtype/src/setup.rs
```diff
@@ -9,50 +9,42 @@ use anyhow::{Context, Result};
...
-            ///
-            /// Do you want to enable live collaboration here? (This will create a {CONFIG_DIR}/ directory.)
-            base_dir.display(),
-        ))? {
-            sandbox::create_dir(&base_dir, &teamtype_dir)?;
-            ui.log("Created! Resuming launch.");
-        } else {
-            bail!("Aborting launch. Teamtype needs a {CONFIG_DIR}/ directory to function");
+        match base_dir {
+            BaseDir::Temporary(_) => {
```
blinry, line 25: Pretty cool that we now don't need to confirm() anymore in this branch!

crates/teamtype/src/config.rs
```diff
@@ -104,6 +104,19 @@ impl Clone for BaseDir {
     }
 }
 
+#[derive(Default, Clone, Debug)]
+pub enum VcsMode {
```
blinry, line 108: Maybe we could add a little comment here? "Whether or not to sync version-control directories like .git/ or .jj/."

crates/teamtype/src/config.rs
```diff
@@ -104,6 +104,19 @@ impl Clone for BaseDir {
...
+    Sync,
+    #[default]
+    Ignore,
+}
+
+impl From<bool> for VcsMode {
```
blinry, line 114: For clarity, how about a VcsMode::from_sync_vcs(sync_vcs: bool), to make the meaning of the bool clearer?

crates/teamtype/src/daemon.rs
```diff
@@ -79,7 +79,9 @@ pub async fn run_daemon(config: Config, init_doc: bool, ui: &UserInterface) -> R
 
     ensure_teamtype_is_ignored(&config.base_dir)?;
 
-    if config.sync_vcs && has_local_user_config(&config.base_dir).is_ok_and(|v| v) {
+    if matches!(config.vcs_mode, VcsMode::Sync)
```
blinry, line 82: Slight preference for deriving PartialEq for the newtypes and using `==` instead of using `matches!`, if you see no downside to that.

crates/teamtype/src/cli.rs
```diff
@@ -7,15 +7,16 @@
...
+use clap::{ArgAction, Args, Parser, Subcommand};
 
 #[derive(Parser)]
 #[command(about, long_about = None)]
 #[command(propagate_version = true)]
 pub struct Cli {
     /// Suppress non-essential informational messages in the console output.
-    #[arg(short, long, global = true)]
-    pub quiet: bool,
+    #[arg(short, long, global = true, action = ArgAction::SetTrue, value_parser = BoolishValueParser::new().map(ConsoleVerbosity::from))]
```
blinry, line 18: Don't love this verbosity. Would it hurt a lot to explicitly convert when we go from the quiet flag to the verbosity variable instead?

crates/teamtype/src/cli.rs
```diff
@@ -81,3 +82,16 @@ pub enum Commands {
...
+
+#[derive(Default, Clone, Copy, Debug, PartialEq, Eq)]
+pub enum ConsoleVerbosity {
+    #[default]
+    Verbose,
+    Quiet,
+}
+
+impl From<bool> for ConsoleVerbosity {
+    fn from(value: bool) -> Self {
```
blinry, line 94: Similarly here, I'd prefer a from_quiet method for example.

blinry (changes requested):
Excited about these refactorings! :)

Noticed that instead of passing the entire Config, you now sometimes explicitly pass the parts of it that the subcomponents need. What's your rule of thumb there? It used to be like this, but when we introduced the Config struct, we thought we might as well pass the entire thing to avoid longer parameter lists, and to make it extensible in the future.

crates/teamtype/src/daemon.rs
alerque: Yes, this was *moved* in some earlier refactoring; it only belongs in one location.

crates/teamtype/src/daemon.rs
alerque: I'm down with that. It's not a derive I remember to reach for but it do like using it that way.

crates/teamtype/src/config.rs
alerque: The downside of that is we end up with a `bool` crossing a module boundary ... if fact it would actually cross crate boundaries. I don't like that at all, hence the two types: one in the CLI arg parsing module (which is part of both the build-time crate *and* the bin crate) and one in the config with explicit code for mapping between types for the different use cases. Your suggestion would end up holding and passing a boolean at some point.

crates/teamtype/src/cli.rs
alerque: Verbosity where? This attribute? None of this is user facing and the user just gets to use the flag as if it was a boolean, but we get it directly mapped to a type in the parsed args.

crates/teamtype/src/config.rs
alerque: Yes. My experience being new to Rust was that it felt awkward to use types and it seemed like a hassle ... "I just need a `String`, why should I wrap it in a type?". But the more I use Rust the more ergonomic it seems to be as explicit as possible with types — and eventually when you hit the right abstraction it seems like they just sort of melt into the background and do the right thing while making every place that uses them simpler than it would be to code using more primitive types directly.

crates/teamtype/src/daemon.rs
alerque: A few more to go, but I can only keep track of so much refactoring at a time ;-)

crates/teamtype/src/sandbox.rs
alerque: The first deref does not end up as `&Path`, it is `Path`. The second deref makes the owned type a reference instead.

This is a fairly common pattern (and Clippy even pushes it on you in some cases) but if you don't like the ergonomics that's fine. I can make it look prettier by just adding an `impl AsRef<Path> for BaseDir`, then you can just use `base_dir` directly in these situtations.

crates/teamtype/src/sandbox.rs
alerque: I'm not convinced we're done there, the `&Path` type in one of the args may still be too low level an abstraction. I wanted to revisit this and evaluate whether a custom type that enforced the absolute path requirement and/or was linked to a specific sandbox instance might be more appropriate. The boolean situation was more pressing but not the whole story.

crates/teamtype/src/setup.rs
alerque: Sure, the silly prompt to approve use of a temporary directory we just created is one bug I had in mind to eliminate before even starting down this road. Of course it could have been fixed in the previous PR with `if temp_dir.is_some() || ui.prompt(...) {`, but we already lost track of the temp dir pointer several times in that spaghetti code. This makes it much easier to not loose the handle but only care about the difference between share types when it matters.

It's possible we might want to benefit from embedding even more info in this structure such as a "pre-approved" status. For example I could see a daemon mode or library entry point that might want to specify both a base directory and automatically skip the confirmation prompt step, so it could generate a type that was a base dir variant that didn't need the extra confirmation.

crates/teamtype/src/sandbox.rs
alerque: I updated the comment to clarify the current intention there.

crates/teamtype/src/config.rs
alerque: It was commented in the Config stuct, but a doc comment here does make more sense. Done.

crates/teamtype/src/config.rs
alerque: It looks like this should be implemented more similarly to how I handled the verbosity flag where the concept of a bool only lasts as far as Clap's arg parsing and we get a meaningful type out of it as early as possible.

crates/teamtype/src/cli.rs
alerque: The downside of that is we end up with a bool crossing a module boundary ... if fact it would actually cross crate boundaries. I don't like that at all, hence the two types: one in the CLI arg parsing module (which is part of both the build-time crate and the bin crate) and one in the config with explicit code for mapping between types for the different use cases. Your suggestion would end up holding and passing a boolean at some point.

crates/teamtype/src/sandbox.rs
alerque: Pushed that `impl`... feel free to re-open discussion if hiding that extra coercion entirely doesn't sit right. I'm wasn't very bothered by the explicit `&**` to both Deref and AsRef it, but am just as happy with the trait being implemented directly and just hiding that step away. Rust won't let you use it wrong either way.

alerque:
> Noticed that instead of passing the entire Config, you now sometimes explicitly pass the parts of it that the subcomponents need. What's your rule of thumb there? It used to be like this, but when we introduced the Config struct, we thought we might as well pass the entire thing to avoid longer parameter lists, and to make it extensible in the future.

I was roughly aiming for the *public* interface being as extensible as possible by just assuming the whole `Config` would be available whenever we need it, but keeping the *internal* interfaces as lean as possible and only giving them what they actually use. If it needs more than a couple of items from the config then sure maybe giving it the whole thing makes sense, but if the function is a clearly defined atomic unit with no public API surface anyway then I tend to think less is more.

I don't know that there is a hard rule we can follow for this. For example I'm fine with passing the whole `UserInterface` into a function that is only going to use `ui.log()`. Technically we could make the function signature just accept a `Fn<&str>` and only have access to the one function, but that doesn't feel right to me. On the other hand if the only thing that a function needs is the base directory, passing it `BaseDir` instead of `Config` feels right.

So the rule of thumb is ... just my gut feeling.

crates/teamtype/src/cli.rs
alerque: > Would it hurt a lot to explicitly convert when we go from the quiet flag to the verbosity variable instead?

Yes I think it would hurt, [per this comment](https://github.com/teamtype/teamtype/pull/619#discussion_r4093845466). The boolean only makes any sense when *very* closely tied to the exact wording of the flag. If the flag was --verbose instead of --quiet the meaning of the boolean would reverse. Passing the boolean value across a crate boundary instead of immediately converting the flag (which is worded a specific way) into something that defines its own meaning no matter where it originates seems more robust against handling mistakes to me.

alerque: Put another way, with this arrangement you could change the CLI  from having a `--quiet` flag into having a `--verbose` flag and reverse the default just by changing `SetTrue` to `SetFalse`, and none of the other code (variable names, method names, code comments, etc.) need updating to still carry their intended meaning.

crates/teamtype/src/config.rs
alerque: I had a look at this and I agree that there is room for improvement, but I don't agree that this is the way to do it. This would end up putting CLI argument handling (`from_sync_vcs()`) into the library crate. The boolean is created in the bin crate in the CLI module. Sadly the CLI arguments themselves need to be handled in both the bin crate and the *build.rs* (which is it's own crate). That makes it really tricky to do anything sensible with types here.

The way the verbosity flag is handled is better because that stays completely inside the bin crate and doesn't need to be accessible to the library, hence we can also compile it into *build.rs* as well (for man pages and completions).

I'm going to suggest that if you don't like the `impl From<bool>`, that just having a helper function in the bin crate's cli module (where we already fiddle with munging CLI values into Config values) is a better place to put an explicit function like you're looking for.

That change is pushed now, does that resolve this for you?

crates/teamtype/src/cli.rs
alerque: Also note where this ends up in #635 (in the bin crate's cli args module). This `impl` is very clearly scoped to the Clap argument to type casting, not available to anywhere it could be confused for anything else.

crates/teamtype/src/config.rs
zormit: As an observer/cursory reader of the PR discussion I was curious if this one was answered elsewhere or whether it was resolved accidentally?

alerque: Good catch, definitely an accident. I think a badly delayed GH interface is probably to blame for a click on the wrong block.

alerque: > This `Default` behavior seems a bit strange to me, like it could unintentionally create a BaseDir pointing to a directory where `teamtype` is run.

Of course it can, but I'd argue it is not unintentional. It's exactly what's indented by every use case we have, and anybody using this from an API should also expect this.

It's not going to create a directory, only a value of type `BaseDir`. The only time a directory is created is using the `create_new()` method, not `::default()`. If you 'accidentally' create a config without specifying something other than the default directory *and* you pass that to `Daemon::new()` (or `run_daemon()` or whatever the public API ends up being) then you should fully expect it to run in $PWD — and I would consider it intentional.

This is also the current/existing behavior. All of `teamtype share`, `teamtype join`, and `teamtype client` default to the current working directory if you don't explicitly add a `--directory` argument. And it isn't even just the CLI, the existing `Config` already implements `Default` and already defaults to the current working directory. This is how the CLI's defaults are already powered. This is just moving where that default is generated from `PathBuf`'s `Default` trait to our own so that we can specify that it is a permanent not a temporary variant.

> We need it because `Config` wants to derive `Default`, because it's used in `parse_client_config`.

Actually we use it in *all* the parse modes, and that isn't a change with this PR, PathBuf's default trait was being used before too.

> My favorite solution would be to remove the Default derives, and to make `run_client` take a `BaseDir` (or `Path`?) directly. The client subcommand is very different from the share/join subcommands anyway, and won't react to anything else right now.

It's entirely possible the `run_client()` function will be better off receiving less config data, but I don't want to mix up that refactoring  with this yet. You're right it needs a lot less information –and possibly structurally different information– to run. That's going to be a major point to settle on when announcing what the public API is. For now I'd like to not refactor that in this pass because it's going to branch out into some other topics too.

Whether we pass `run_client()` a `Config` or a `BaseDir` I think I could go either way on, except that every time we've had to touch the assorted `run_*()` functions I've been trying to normalize them to take similar data and return similar types (all async, all return a handle to the tokio pool doing the work, all being launched with a config, etc.). I don't know yet if we'll possible have other config keys that are relevant to the client.

In summary, I don't agree at all that trying to remove the `Default` trait would be an improvement, and I'd like to put off decision about what to exclude from the API. For now for our own use internally trying to shape them to be more similar to each-other is feeling much nicer than being totally disparate.

alerque: Addendum: I was incorrect in some of that working from memory. We aren't actually making use of the fact that `Config.base_dir` is a `PathBuf` or that it has a `Default` trait, we're doing our own in the CLI module (for all modes) and manually filling in empty values with `env::current_dir()`. Besides checking *prior* to run time whether the directory exists, I'm not sure that is a clear win. Frankly all of that `parse_x_config()` stuff looks like it could benefit from better types and `Into` traits, but again I can only pull on so many threads at once.

I switched to `PathBuf::default()` instead of `PathBuf::from("./")` since they are equivalent and we're not absolutizing or checking writability or anything at the time we generate a config, only when the daemon tries to use it.

PR #620, opened 2026-09-11
Correct directory initialization order

alerque:
This is the first part of what was #571.

----

This resolves all sorts of scope problems and leaves a few more for later. This kicked off with knowing that needing an async function to parse a local config was just wrong. That turned out to be because it was attempting to resolve network peers inside the function that was supposed to parse the CLI arguments. That should definitely not need to be async, and that would be a problem for API consumers.

Attempting to move that around, it became apparent that the entire system of pre-initializing some parts of the config, then interacting with the UI (confirming directory creation, etc.) then generating a new config, then finally passing that to actual functions was just off. It put a lot of things on the wrong side of bin/lib splits and nested in functions that had other jobs.

Unraveling this meant making the config struct a static one-time initialization thing, then checking it and initialization the directory from that, then moving on to other steps.

There are some awkward bits in here (it adds a bunch of repetitive expects) but those are all for cases we can a) prove internally won't happen because they will have errored out previously, and b) fixing those will involve adding correct error handling to a whole bunch of functions that have other unwraps in them too for no good reason.

----

**Self-review checklist**


crates/teamtype/src/config.rs
```diff
@@ -35,9 +36,9 @@ pub enum Peer {
 }
 
 #[derive(Clone, Default, Debug)]
-#[must_use]
+// #[must_use]
```
zormit: drop it?

crates/teamtype/src/config.rs
```diff
@@ -245,17 +252,19 @@ fn get_username_from_config_file(
...
-            /// Using the Git username '{username}' as username, to display next to the cursors other people see.
-            /// Teamtype uses the Git username as username by default.
-            /// You can set the configuration value `username` in your `.teamtype/config` to override this username.
-            /// You can also use the flag `--username` when using the `share`/`join` subcommands.
-        ));
-    }
-    username
+fn get_username_from_git(base_dir: Option<&Path>, ui: &UserInterface) -> Option<String> {
+    base_dir.and_then(|base_dir| {
+        let username = get_git_username(base_dir);
```
zormit, line 260: Result of this change (because we don't resolve the base_dir as early as we can?): If there's no `base_dir` (e.g. when you join it temporarily), then we don't query for the git username.

Discussion around how to determine the username: In a case where we don't have a git directory, it could still be nice for users to have the option to get the username from the global git config. Maybe provided that as a question as some point of startup?

crates/teamtype/src/daemon.rs
```diff
@@ -54,17 +57,32 @@ use crate::wormhole::put_secret_address_into_wormhole;
...
+
+    let base_dir = config
+        .base_dir
+        .clone()
+        .context("Temporary directory not initialized")?;
+
+    // We *may* be in join mode and *may* be doing so from a join code. If so make sure that's
+    // resolved to a regular peer before continuing.
+    config = config
+        .resolve_peer(ui)
```
zormit, line 71: This panics in the case of `teamtype share`. To fix, `resolve_peer` should simply be a noop in the case when no peer is configured.

zormit (commented):
Didn't look at the next PR yet, but probably makes sense to look into it before merging this one. Stopped reading after `daemon.rs`, leaving some comments for now.

crates/teamtype/src/daemon.rs
alerque: Now action only happens `if Some(peer) ...`.

alerque: The whole construct gets a small ergonomics bump in the following PR because with the new dedicated types we have the ability to implement `Copy` instead of just `Clone` on the `Config` struct which means we don't have to clone as a mutable and can create a new struct with just one modified value, but that isn't very important to the logic fix here.

crates/teamtype/src/config.rs
alerque: The code in *this PR* is indeed a logic change that happened accidentally because the `&Path` → `Option<&Path>` change forced the conditional. Incidentally I still hold that this is what we should be doing by default.

The *next PR* changes `Option<&Path>` → `BaseDir` which is an enum that can always be derefed to `&Path` whether it is an existing or temporary directory. This [effectively reverts the logic change to what it was](https://github.com/teamtype/teamtype/pull/619/changes#diff-804efa190926392468c940b322c32d3acedb62b4e55c81d3ebd456a15e79dc29R339).

I think we should 

1. Just let the change go through in this PR because adding the necessary logic to keep it the same would make this commit even more cluttered than it is, and
2. promote this to a new issue / PR to bring back the accidental win from this PR, this time gated behind a user prompt for permission.

alerque: I've added a prompt and some related logic in #627. Lets discuss the merits of the change there.

The only open question here is should this commit be rehashed to not introduce and then backtrack the logic, or can we just ignore it and review both PRs together?

zormit: We've discussed and it's ok to have a slight regression.

crates/teamtype/src/editor_connection.rs
```diff
@@ -54,13 +54,18 @@ impl EditorConnection {
...
     ) -> Vec<EditorProtocolMessageToEditor> {
+        let base_dir = &self
+            .config
+            .base_dir
+            .clone()
+            .expect("Temporary directory not initialized");
         match message {
             ComponentMessage::Edit { file_path, delta } => {
-                if let Some(ot_server) = self.ot_servers.get_mut(file_path) {
+                self.ot_servers.get_mut(file_path).map_or_else(Vec::new, |ot_server| {
```
zormit, line 65:
```suggestion
                // Fall back to empty Vec (=do nothing), if we don't have the file open.
                self.ot_servers.get_mut(file_path).map_or_else(Vec::new, |ot_server| {
```

Re-introducing the comment that explains the larger context of why we return an empty Vec.

crates/teamtype/src/sandbox.rs
```diff
@@ -85,17 +85,20 @@ pub fn create_dir(absolute_base_dir: &Path, absolute_dir_path: &Path) -> Result<
...
     Ok(())
 }
 
-fn create_dir_all(absolute_base_dir: &Path, absolute_dir_path: &Path) -> Result<()> {
+pub(crate) fn create_dir_all(absolute_base_dir: &Path, absolute_dir_path: &Path) -> Result<()> {
     let canonical_dir_path =
         check_inside_base_dir_and_canonicalize(absolute_base_dir, absolute_dir_path)?;
-    fs::create_dir_all(canonical_dir_path)?;
+    fs::create_dir_all(canonical_dir_path)
+        .context("UNable to create directory(s) with FS function")?;
```
zormit:
```suggestion
        .context("Unable to create directory(s) with FS function")?;
```

zormit (approved):
We don't know why, but noticed a small regression: `--temporary-directory` seems to crash in this PR. In `main` it works and #619 it works again. So maybe we should only merge this, when #619 is also approved, so we don't publish a broken build.

```
% tt join 6-orlando-erase --temporary-directory
Using the fallback value for username 'Anonymous' as username, to display next to the cursors other people see.
You can set the configuration value `username` in your `.teamtype/config` to override this username.
You can also use the flag `--username` when using the `share`/`join` subcommands.
> '/Users/mn/.cache/teamtype/.tmpCnUkxB' hasn't been used as a Teamtype directory before.
Do you want to enable live collaboration here? (This will create a .teamtype/ directory.) Yes
Created! Resuming launch.
Derived peer from join code. Storing in config (overwriting previous config).
Storing peer's address in /Users/mn/.cache/teamtype/.tmpCnUkxB/.teamtype/config
Connected to peer: ae437d6fc0b9f48...0fb0358

thread 'tokio-rt-worker' (4204427) panicked at crates/teamtype/src/daemon.rs:284:29:
Failed to persist to '/Users/mn/.cache/teamtype/.tmpCnUkxB/.teamtype/doc'
```

The rest of the PR looks fine, added some minor suggestions.

alerque:
> So maybe we should only merge this, when #619 is also approved, so we don't publish a broken build.

I agree with this.

I'll fix up the spelling and comments and whatever here, but it really only makes sense to merge this when #619 is approved for merge too because this would be a bad point to run or branch from for in between work.

crates/teamtype/src/editor_connection.rs
alerque: Good call, that's a useful code comment. I'm not 100% sold on the coding style Clippy forced on us here — at least to my brain some of these fluent chains are not always as clear as more explicit logic structures, and even with the original logic structure the code comment did help understand *why* something was being done, not just being able to deduce the *what* from the code.

alerque:
I took a look at the regression in temp dir handling. It's definitely a bug in this patch. In moving things around to get the temp dir creation into the daemon instead of the CLI I inadvertently also affected the scope where the temp dir handle is held and it now gets dropped after the daemon starts up but before the daemon itself goes out of scope and gets dropped. That's causing a race where the daemon is initializing fine, but then the temp dir is getting deleted out from under it.

The next PR fixes this because the new type for holding the base dir also holds the handle for as long as *anything* uses the base dir. This is the obvious **correct** fix. This patch was developed while working on that change, and all this  stuff is still needed to get there.

I tried coding up what it would take to get this precursor refactoring to not have that bug and I got it working but the code is *hideous*. It basically backtracks on some of what is accomplished here and passes around tuple types to not let the variable go out of scope. It would need to be reverted in the next PR.

The only solution to make all commits pass all tests is to squash this with the BaseDir type addition in the next PR. I don't like that because it adds a whole bunch of scope refactoring on top of the change in types and makes it harder to follow the changes. I could add on a patch here and revert it in the next PR, but I don't really like the hideas intermediary form. The BaseDir type that takes care of this problem is where we're trying to arrive, and this broken partial refactor only came about because I was writing that code at the same time as realizing this refactoring needed to happen to get there.

Let me know how you'd like to proceed. I'd prefer to just wait for approval on the next PR and merge these together.

blinry:
Yes, I think way to go about it is totally reasonable :)

PR #631, opened 2026-09-17
Gate Unix specific bits and add Windows specific bits 

### LINKED ISSUE #258: Windows support

### LINKED ISSUE #485: tracking issue: Windows Support

alerque:
This is a replacement for #534. Since our PR review queue is perpetually backlogged and this one is tricky to rebase as a separate branch, I'm hoping putting it in a specific line will make it easier to review and keep rebased until we can get it merged.

Closes #258
Closes #485 

See previous PR for some testing history.


alerque:
This is now tested on Windows and some kinks with sanboxing (e.g. projects on drives other than `C:>`, network paths, etc.) and the watcher (Windows emits less specific events on file changes). It now works pretty comparably to its Linux cousins. Shortcuts do not sync properly, but symlinks from Linux don't sync right either so it's on par there. Binary files seem to get copied fine.

alerque:
The fuzzer is having trouble dealing with symlinks in the temporary file path (and hence canonicalization issues) on MacOS. I have a fix for that, but it's a little difficult to test what parts of the fix are relevant without also basing on #617, so possible that PR should get merged before this one is finalized and made to be green in CI. The Windows stuff is good to review.

